PR #2, opened 2013-02-20
Features/django sitetree

jefftriplett:
This branch is going to definitely needs some eyeballs on it. I made several assumptions along the way:
- Pages that aren't in urls redirected to `inner`. I debated making a `todo` page which might be better going forward.
- I commented out the `main_navigation` section in the docs, pypi, or shop index templates. The currently default to the default (python) navs.
- `sitetree/top.html` for the `top` block.
- `sitetree/submenu.html` for the `main_navigation` block.
- `sitetree/footer.html` for the footer / `sitemap` block.
- I'm using the `alias` field on Tree Items to store the css class name. 
### CSS Classes used for selection
- The css class `current_item` and `selected` are added to help style navigation for the trunk / top-most link. 
- The css class `current_branch` and `selected` are applied if the link for the trunk / top-most link.

IE: clicking on "About" would in the menu with a hierarchy of Python -> About would give:
- Python: would contain the classes `current_branch` and `selected`
- About: would contain the classes `current_item` and `selected`
- To see the full menus please use `django-admin.py loaddata fixtures/sitetree_menus.json` (you may also use `manage.py` if you prefer). I'm open to a better way of doing this since fixtures are a pita. Sitetree has a `sitetreedump` and `sitetreeload` command but I don't see how they are any better fwiw. :poop:


jhogue:
Finally got this up and running. Sorry about the delay. Loaded the sitetree json data as well, thanks for that. Your assumptions were mostly correct, but I want to clarify one thing before it goes too much further: 
- The footers are always a reflection of the current `.main-navigation`. So, in the PSF section, the footer needs to reflect the same content as the main navigation. Same for any of the _"sister sites"_ (PSF, Docs, PyPi, etc...). 
- When on an inner page (/inner/) I assume the classes for `.current_branch` and `.selected` are not in place yet, but your assumptions around how they should work are correct. 
- The IDs on those `.main-navigation .tier-1` LIs are needed later on for some JS functions. Right now, their value is "None". In the mockups we used the section slug (about, documentation, downloads, etc...) but in the end we may want to use something less likely to change, like the ID of the item... a class like `.itemid-XX`. 
- Since you are using Django Sitetree, I assume at some point we can have a nice Google XML sitemap for SEO inclusion. Let us know what the path to that will be for the robots.txt file, or add it in there at some point. 

Thanks for this. Looks good to me otherwise.

jhogue:
Oh, I also noticed a double declaration of the doc type in base.html. 

```
<!doctype html>{% load sitetree %}
{% load compressed %}<!doctype html>
```

Not sure which one needs to be removed.

fcurella:
According to the HTML5 spec it doesn't really matter if the the document starts with the `\n` (removing the first one) or straight with the doctype (removing the second one).

FWIW, I usually choose to start with the doctype, just because that was the way in the XHTML days.

jhogue:
Yes, but I mean, when I looked at the source of the page, there were two in
there as well.

<!doctype html>

<!doctype html>

 etc...

On Fri, Feb 22, 2013 at 11:04 AM, Flavio Curella
notifications@github.comwrote:

> According to the HTML5 spec it doesn't really matter if the the document
> starts with the \n (removing the first one) or straight with the doctype
> (removing the second one).
> 
> FWIW, I usually choose to start with the doctype, just because that was
> the way in the XHTML days.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/proevo/pythondotorg/pull/2#issuecomment-13950603.

## 

jefftriplett:
@jhogue Sorry about that. I think that happened in merge of master but I pushed a fix. I'm not sure why it showed up with the double header but it's fixed now :)

jhogue:
Should I go ahead and hit the Merge button now?

PR #3, opened 2013-02-26
Features/jobs

fcurella:
detail view and creation view are already stubbed but they will need to be completed when their mockups are ready


jhogue:
Might I add a few suggestions, though it may be too early to add this in... Jobs could also cross reference a Company object, where we can store some boilerplate (bio) info about the company as well as URL and logo image. A single Job could also have a field for an optional URL to a more complete job posting, presumably to the company's website, or simply as a place to send someone to apply. 

I will work on a Job listing Detail template.

fcurella:
@jhogue That's a good idea. I wondered about that, but I couldn't find any Company list or detail page, and so I opted for the simplest thing possible. I'll work on making that field a FK.

jhogue:
Sure thing. I'm not sure that a Company will be a front-end thing (except
when we want to pull a logo) but it will make adding new job posts easier
if they get connected to some repetitive information like the Company can
contain.

On Tue, Feb 26, 2013 at 2:15 PM, Flavio Curella notifications@github.comwrote:

> @jhogue https://github.com/jhogue That's a good idea. I wondered about
> that, but I couldn't find any Company list or detail page, and so I opted
> for the simplest thing possible. I'll work on making that field a FK.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/proevo/pythondotorg/pull/3#issuecomment-14133664
> .

## 

fcurella:
I've also added a 'jobs by company' view, but i'm not sure where should we link to it in the templates.

jacobian:
Yeah, I agree that a "company" object would be good to have; we can link
that up to success stories, sponsorships, etc.

fcurella:
maybe it should be out of the 'jobs' app and be in its own app? (BikesheddingWarning)

cms/forms.py
```diff
@@ -0,0 +1,28 @@
+from django.conf import settings
+from django.utils import timezone
+from django import forms
+
+
+class ContentManageableModelForm(forms.ModelForm):
+    created = forms.DateTimeField(required=False)
```
jefftriplett: I don't think you want to expose these three fields at all in the frontend. By default they show up as editable (albeit non-required fields). I'd hide them and set them on `save_model`.

peterlandry: Just as a matter of curiosity, what's your personal heuristic on save_model(), save() vs signals? I haven't kept up to date on the discussion around that

jefftriplett: I'd update `creator` in the `save_model` because it has access to `request.user`. Otherwise, I tend to do all non-blocking updates in the model's `save` method. I normally don't use signals but I do make a rare exception for profile updates and making sure a default profile exists. For the most part, celery replaces my need for using signals (unless my signal fires off celery but I digress).

fcurella: `save_model` will work for the admin, but not for the generic views. I'm actually ok with showing those fields in the admin, but I'm worried about them showing up with things like `{{ form.as_p }}`

peterlandry: Access to request.user is a good point, thanks!

I'm curious though about the celery replacing signals part -- especially in the case of a third party app wanting to non-invasively take some action. Do you watch for tasks to be executed, or do you just not see that case happen often?

jefftriplett: @peterlandry for third party apps I use signals and most of the time I'll have the signal call a celery task. If there is a performance penalty then my hammer is to use celery to help take the load off the frontend and keep the frontend speedy. Updating a search index, sending emails, making a web api call (like twitter), etc are typically what I'll build tasks for. If it's just a trivial update to a field in a 3rd party app then a signal by itself is fine.

jefftriplett:
looks good :+1:

PR #4, opened 2013-02-27
Design/more inner pages

jhogue:
Added more content and markup to the Single Job template. Content is mocked in and template tags need to be added. 

The Django template has comments in to explain where we intended some of the content to come from. Let me know if there are questions.


jhogue:
I just realized that I did not commit the rendered CSS on purpose (we dont usually, as we run Compass on the server) but you may need those files here in this instance. Let me know if you do.

fcurella:
we shouldn't need to update the rendered css (its only purpose is to make travis happy).

Everything looks good, the models will need some changes but I was expecting that.

I only have one doubt: seems like the 'telecommuting' attribute is on the `Company` model. Shouldn't it be on the `Job` model?

jhogue:
Agreed. Job model makes more sense, as some positions in a company may not
allow telecommuting but others might allow it. Better to not set it
company-wide.

On Wed, Feb 27, 2013 at 2:28 PM, Flavio Curella notifications@github.comwrote:

> we shouldn't need to update the rendered css (its only purpose is to make
> travis happy).
> 
> Everything looks, the models will need some changes but I was expecting
> that.
> 
> I only have one doubt: seems like the 'telecommuting' attribute is on the
> Company model. Shouldn't it be on the Job model?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/proevo/pythondotorg/pull/4#issuecomment-14193748
> .

## 

PR #5, opened 2013-02-27
Features/success stories

jefftriplett:
A work-in-progress of sorts. Needs:
- Integration with @fcurella's company app and to add additional fields.
- Integration with @fcurella's cms changes.
- Migration post Company integration.
- Categories are simple so MPTT might be needed. I just went with something simple for now.
- Templates from Proevo. I used the index page as a basis for the others but templates need styled.

The spec was very lose so please tell me what else is needed if anything?


jefftriplett:
Jacob said it looked good. So I'm going to merge and we can create tickets for anything that needs touched up.

PR #6, opened 2013-02-28
Features/sponsors

jefftriplett:
Added a basic sponsors app with a template that needs styled.
- Needs logo support on the Company app. should be a different pull request.
- Page needs styling.


jefftriplett:
Merging via Jacob. We can fix problems via tickets.

PR #8, opened 2013-03-06
Design/device icons

jhogue:
Added files and support for a new set of custom Pythonicons. Changed markup (names of classes) where new names were needed to match the new icons.


jacobian:
Merged & deployed, thanks!

PR #14, opened 2013-03-11
Features/pages template

jefftriplett:
Misc cleanup page cleanup.


jacobian:
:horse:

jhogue:
The feedback form... can we add a checkbox (separate from the arrays of boxes there now) that says something like "Yes, consider me a Beta Tester! I want to be notified when there are new features to review and send me instructions with other ways to help."

jefftriplett:
I'm going to merge and then add the consider me a beta tester feature to trello. I'll work on that in the morning. I need to add migrations too and have Jacob fake them otherwise, I'd just add them tonight.

PR #17, opened 2013-03-12
Added an is_beta_tester field to feedbacks app.

jefftriplett:
Added an is_beta_tester field to the Feedbacks model along with form and test support.

### DIFF
--- a/feedbacks/admin.py
+++ b/feedbacks/admin.py
@@ -7,8 +7,8 @@ from cms.admin import NameSlugAdmin
 class FeedbackAdmin(admin.ModelAdmin):
     date_hierarchy = 'created'
     filter_horizontal = ['feedback_categories']
-    list_display = ['pk', 'name', 'issue_type', 'created']
-    list_filter = ['issue_type', 'feedback_categories']
+    list_display = ['pk', 'name', 'issue_type', 'is_beta_tester', 'created']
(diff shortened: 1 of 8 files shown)


jacobian:
So we;d like to support sign-up for Thursday (the idea being people who sign up for the site at pycon get a special "pycon2013" badge or something), and we'd planned to ask at that point if people awnted to be beta testers. Should we tie these two things together, or just be OK with two pools of beta testers?

jefftriplett:
What's the signup suppose to be? Just following this card: https://trello.com/c/3NHBUmbE

jacobian:
There's this: https://trello.com/c/1esfqFLo

I don't think we have to nail all of that for Thursday, but at least a
user/email/password signup process would be good.

I'm OK with not integrating these two things if this makes for blocking,
just trying to think out loud.

jhogue:
Jacob: The idea in my mind was to have them be separate, but they feed each other. As in, one person may not leave feedback, but want to start their profile to get the badge and while they are doing that, they sign up to be a Beta Tester. In both places, we ask them if they want to sign up. Can't hurt to get them on either end.

jacobian:
Yup, that's the conclusion Jeff and I came to, too. We had a quick IRC
discussion and I think this is good to go now.

PR #20, opened 2013-03-14
added custom user model. Also declared migration bankruptcy.

fcurella:


fcurella:
the `feedbacks` app will have to be fake-migrated, but I think the other apps/table are safe to destroy.

fcurella:
Ready for review

jefftriplett:
LGTM. I'm going to modify it to work with social auth and django-allauth post merge.

PR #23, opened 2013-03-19
Improvement: Added a CSS3 transition as a hoverintent replacement

jhogue:
Navigation drop downs were too sensitive. Used CSS transitions and a transition delay to recreate a JS hover intent effect. Modern browsers only. 


jefftriplett:
LGTM. :+1:

PR #36, opened 2013-08-29
Updates to base.html from current best practices

jhogue:
More meta tags and {% block %} tags. Added repetitive info to the settings/base.py file into a variable array called SITE_INFO. Added a humans.txt colophon file.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -6,18 +6,24 @@
 <!--[if IE 8]>         <html class="no-js ie8 lt-ie9">             <![endif]-->
 <!--[if gt IE 8]><!--> <html class="no-js" lang="en">              <!--<![endif]-->
 
-<head prefix="og: http://ogp.me/ns# fb: http://ogp.me/ns/fb# website: http://ogp.me/ns/website# article: http://ogp.me/ns/article#">
+<head>
 
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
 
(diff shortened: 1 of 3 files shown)


jhogue:
I realize now that I need a context processor somewhere. Example: 

def site_variables(request):
    return {'SITE_INFO': settings.SITE_VARIABLES}

And then that processor will have to be registered in the settings. 

Since the structure of this repo is what it is and I know just enough to be dangerous, can someone please add that stuff in where it belongs before this PR is merged?

rschamp:
@jhogue Funny you should say that — your example looks like a template tag, and that might be a better way to do it than a context processor.

templates/base.html
```diff
@@ -43,30 +49,58 @@
...
+    <meta property="og:section" content="{% block og-section %}{% endblock %}"> {# A high-level section name. E.g. Technology #}
+    <meta property="og:url" content="{% block og-url %}{{ get_absolute_url }}{% endblock %}">{# permalink to the curent page #}
+    <meta property="og:image" content="{% block og-image %}{% endblock %}">{# A path to an image used on the page. Helpful for telling crawlers what image to use for a page preview. Can be an array, meaning, there can be more than one of these tags (duplicate the WHOLE tag). #}
+    <meta property="og:video" content="{% block og-video %}{% endblock %}">
+    
+    <link rel="author" href="{{ STATIC_URL }}humans.txt">
+    
+    {% comment %} No support for these yet... 
+     
+    <link rel="sitemap" type="application/xml" title="{{ SITE_INFO.site_name }} Sitemap" href="/sitemap.xml">
```
rschamp, line 80: This could be `{% get_site_variable "site_name" %}` or something like that.

PR #38, opened 2014-01-27
Added a default page content template for the PSF section

jhogue:
We needed this, as there was no template with PSF navigation and styles present to handle PSF content pages.


jhogue:
@frankwiles Please review as soon as you can and then publish these changes to the Preview site. We'd like this template to be available for content editors as they build out Sitetree.

PR #40, opened 2014-02-04
Features/job comments

fcurella:


frankwiles:
Other than pre-filling in the email on the form if we know it, LGTM.

PR #94, opened 2014-02-20
Bugfix/86 dupe calendars

fcurella:


fcurella:
after this gets merged, we'll have to remove the dupe calendars from the admin. I'd do it now but the importer runs every hour and it will justt add them back.

PR #111, opened 2014-02-20
Reduce size of homepage example to fix overflow

meshy:
http://imgur.com/a/7JJK7


jhogue:
I was going to do some more CSS investigation but this approach is fine with me.

meshy:
I'm still seeing this issue on the site. Has this been deployed?

Issue #81, opened 2014-02-20
The PSF Paypal donate button is too big


malemburg:
We do want to have people click on it, but right now it looks more like a funny balloon on the page :-)

http://python.org/psf/donations/

PR #121, opened 2014-02-20
Fix Donate Button on PSF Donate page

jhogue:
From http://python.org/psf/donations/, input[type=image] needed width reset to auto. Also fixed some border box declarations to remove duplication.

Issue #81 my malemburg


xen:
This is dupe for #81

PR #134, opened 2014-02-20
Simple database-backed code samples.

peterlandry:
The initial samples as shown on the site now are included in fixtures/codesamples.json

### DIFF
--- a/pydotorg/views.py
+++ b/pydotorg/views.py
@@ -16,7 +16,7 @@ class IndexView(TemplateView):
     template_name = "python/index.html"
 
     def get_context_data(self, **kwargs):
-        context = super(IndexView, self).get_context_data(**kwargs)
+        context = super().get_context_data(**kwargs)
         context['code_samples'] = CodeSample.objects.published()[:5]
         return context
 
--- a/pydotorg/views.py
+++ b/pydotorg/views.py
@@ -3,11 +3,22 @@ import itertools
 from django.http import HttpResponse
 from django.core import serializers
 from django.conf import settings
+from django.views.generic.base import TemplateView
 
+from codesamples.models import CodeSample
 import sitetree.models
 import boxes.models
 import feedbacks.models
 import pages.models
+
+
+class IndexView(TemplateView):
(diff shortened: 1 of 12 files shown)


fcurella:
i think github is marking this as conflicts because the diff contains several `>>>`. Merge with caution.

FWIW, I'm +1, it will at least remove a few PRs

templates/python/index.html
```diff
@@ -8,106 +8,14 @@
 
             <ul class="launch-shell menu" id="launch-shell"><li><span class="message">Launch Interactive Shell</span><a class="button prompt" id="start-shell" data-shell-container="#dive-into-python" href="/shell/">&gt;_</a></li></ul>
 
+            {% get_code_samples_latest limit=5 as code_samples %}
```
fcurella: we could just execute the query from the view. I'm kind of wary of templatetags hitting the DB

peterlandry: The homepage is just a template view, I believe.

fcurella: yes, we can replace that view with a subclass of TemplateView and override `get_context_data`

peterlandry: https://github.com/python/pythondotorg/blob/master/pydotorg/urls.py#L12

That's the main reason I went with a template tag. It could be changed, ofc, I just didn't want to rock the boat too much.

peterlandry: Alright, I'll make that in a bit and push it up. I'd like to leave the template tag for other future uses, though. Sound reasonable?

fcurella: I say go for it

pydotorg/views.py
```diff
@@ -3,13 +3,24 @@
...
 import boxes.models
 import feedbacks.models
 import pages.models
 
 
+class IndexView(TemplateView):
+    template_name = "python/index.html"
+
+    def get_context_data(self, **kwargs):
+        context = super(IndexView, self).get_context_data(**kwargs)
```
fcurella: use py3k's `super()` (no arguments) for consistency with the rest of the ode base

PR #139, opened 2014-02-20
List upcoming events from the database on the homepage.

peterlandry:
There's a little bit of filter gymnastics to get SHORT_DATE_FORMAT to work with the say-no-more span, but it should display date/month order properly localized now.

### DIFF
--- a/events/templatetags/events.py
+++ b/events/templatetags/events.py
@@ -1,0 +1,16 @@
+from django import template
+from django.utils import timezone
+
+from ..models import Event
+
+
+register = template.Library()
+
+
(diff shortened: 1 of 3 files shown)


peterlandry:
@fcurella I'll take the same approach here as with the code samples, to remove the need for the database hit.

fcurella:
@peterlandry since this 'widget' is going to be shared across all pages, I think in this case it makes more sense to have the template tag than to update every view (or maybe use a context processor?)

peterlandry:
Alright. I'll leave it alone. Thanks for the feedback.

PR #141, opened 2014-02-20
Better example of for looping through a list

benhoyt:
I strongly believe examples should show best practices -- for summing numbers, it's Pythonic and much faster/simpler to use the built-in sum(). The example would be just as simple and show the same thing if it was changed to multiplying a list of numbers to get their product.

Also, my example doesn't shadow the builtins `sum` and `list`.

### DIFF
--- a/templates/python/index.html
+++ b/templates/python/index.html
@@ -74,12 +74,12 @@ Hi, Python</span></code></pre>
 
                     <div class="slide-code">
 <pre><code><span class="comment"># For loop on a list</span>
->>> list = [2, 4, 6, 8]
->>> sum = 0
->>> for num in list:
->>>     sum = sum + num
->>> print("The sum is:", sum)
-<span class="output">The sum is: 20</span></code></pre>
+>>> numbers = [2, 4, 6, 8]
+>>> product = 1
+>>> for number in numbers:
+>>>     product = product * number
+>>> print('The product is:', product)
+<span class="output">The product is: 384</span></code></pre>
                     </div>
 
                     <div class="slide-copy">


alex:
This looks good to me, but doesn't merge cleanly anymore, can you merge/rebase.

peterlandry:
Samples are in the database now. Can you push as a change to `fixtures/codesamples.json`?

benhoyt:
@peterlandry I'm not sure I want to update the JSON directly by hand here -- is that done via the Django admin? Are you able to update for me? I don't care about commit credit, I'd just like to see the example updated. :-)

peterlandry:
If you update the fixture, we'll ask the hosting team to load it. I'm just trying to prevent the fixture from getting out of synch with the database. Ultimately I'd like to be able to support community-curated examples, which is why they've moved to the database (you may have noticed the amount of critique directed at their contents!).

Editing it by hand is fine, as long as it will import cleanly with `manage.py loaddata`

johanovic:
I think the list example should be even easier. For complete newbies the idea of dynamically reassigning a new value to a variable might already be too much of a challenge.

+>>> numbers = [2, 4, 6, 8]
+>>> product = 5
+>>> for number in numbers:
+>>>     print(product \* number)

Sure, it is very basic, but I think the carrousel is trying to be basic, right?

benhoyt:
@peterlandry , thanks -- however, I'm afraid I don't have time to action this just now (I just moved half way round the world with my family :-). Anyone else who's interested, please go ahead and do it.

peterlandry:
I can probably take care of that this weekend. Sorry to put up barriers to better examples! I'll push a PR referencing this one with your amended code.

benhoyt:
Sounds great, thanks!

templates/python/index.html
```diff
@@ -74,12 +74,12 @@
...
->>> list = [2, 4, 6, 8]
->>> sum = 0
->>> for num in list:
->>>     sum = sum + num
->>> print("The sum is:", sum)
-<span class="output">The sum is: 20</span></code></pre>
+>>> numbers = [2, 4, 6, 8]
+>>> product = 1
+>>> for number in numbers:
+>>>     product = product * number
```
merwok, line 80: Is it normal that the prompt is not a secondary prompt here?

johanovic, line 80: You should still print product, right?

benhoyt:
@peterlandry -- any update on this?

templates/python/index.html
merwok, line 80: I don’t understand the question.  What I meant was: is it normal that this line is a block but uses the primary prompt `>>>` instead of the usual secondary prompt `...`?

berkerpeksag:
Superseded by #526.

PR #143, opened 2014-02-20
Comment out the python-dev box on /blogs/

akuchling:
At the bottom of www.python.org/blogs/, there's a box titled "Python-Dev Blogs" with some mocked-up links.  Let's comment out this box for now.


frankwiles:
Good idea thanks!

PR #153, opened 2014-02-21
Standardize on ISO representation for latest news and upcoming events.

peterlandry:


cms/templatetags/cms.py
```diff
@@ -0,0 +1,18 @@
...
+def responsive_date(date):
+    """ Returns an ISO date, with the year tagged in a say-no-more span.
+
+    This allows the date representation to shrink to just MM-DD. """
+    date_templ = '<time datetime="{timestamp}">{month_day}<span class="say-no-more">{year}</span></time>'
+    return mark_safe(date_templ.format(
+        timestamp=format(date, 'c'),
+        month_day=format(date, 'm-d'),
+        year=format(date, '-Y')
+    ))
```
fcurella, line 18: I'd rather pass the single 'numbers' to `date_template`, and add the dashes in there rather than in `.format()`, but it's really a bit of a bikeshed. The important thing is that we have a unified way of printing date and that outputs what we want. +1

fcurella:
LGTM :cake: :earth_americas: :earth_africa: :earth_asia:

PR #184, opened 2014-02-24
Fix OS detection for Linux/X11 to show the "source" box.

birkenfeld:
Also change references to the now-defunct "Linux" OS to "Source release".

### DIFF
--- a/templates/downloads/os_list.html
+++ b/templates/downloads/os_list.html
@@ -10,13 +10,19 @@
 {% endblock breadcrumbs %}
 
 
-{% block page_title %}Python Releases for {{ os.name }} | {{ SITE_INFO.site_name }}{% endblock %}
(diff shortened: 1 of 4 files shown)


templates/downloads/os_list.html
```diff
@@ -10,13 +10,23 @@
 {% endblock breadcrumbs %}
 
 
-{% block page_title %}Python Releases for {{ os.name }} | {{ SITE_INFO.site_name }}{% endblock %}
+{% block page_title %}
+{% if os.slug == 'source' %}
+Python Source Releases
```
benjaminp: Presumably we still want {{ SITE_INFO.site_name }} in this title.

birkenfeld: True!

Issue #65, opened 2014-02-20
Remove the feedback bar from the footer.


1st1:

gvanrossum:
Why remove it? Don't you want feedback? Or maybe replace it with a link to the github tracker?

1st1:
A link to this github repo would be great, and probably enough. I myself needed to post on one of python mailing lists to get an idea of where to put my feedback.

The feedback bar also has a lot of references to the beta period (checkbox to make you a "beta tester" etc).

1st1:
There is also a vertical feedback button, glued to the right side of the page. We can retarget that one to point to this repo's issues.

birkenfeld:
If the feedback submitted there is acted upon (where does it go?), it is still much easier for casual visitors to report issues there than here.

After a while, when "trivial" problems like missing content or broken links are sorted, I would leave only a link to this tracker.

benjaminp:
I agree. The feedback table is full of spam, and I doubt anyone is ever going to read through it. Issues should now be submitted here on GH.

malemburg:
I opened #189 to address the github tracker link request.

birkenfeld:
OK, the feedback goes into a table in the DB. There's no spam protection, so we have over 16'200 feedback entries, mostly advertising medical supplies and worse.

Looking at that I recommend disabling the form ASAP to avoid frustrating users with genuine feedback by ignoring them.

malemburg:
Agreed. Related to this, I've opened #190 to address the obvious spam signup issue we will need to solve somehow.

malemburg:
Can we get an export of the feedback table to try to filter out the ham entries ?

benjaminp:
Here's is a an export of the table in [pg text format](http://www.postgresql.org/docs/current/static/sql-copy.html#AEN69009) with names and emails removed: http://python.org/~peterson/feedbacks.tar.xz

malemburg:
On 25.02.2014 16:14, Benjamin Peterson wrote:

> Here's is a an export of the table in [pg text format](http://www.postgresql.org/docs/current/static/sql-copy.html#AEN69009) with names and emails removed: http://python.org/~peterson/feedbacks.tar.xz

Thanks, Benjamin. In order to put the data through an email
junk filter tool it would be useful to have both names and emails
as well, esp. since the table does include some easily recognizable
names from the community.

Could you perhaps send the dump by email with those fields
included ?

## Thanks

Marc-Andre Lemburg

[...]

PR #194, opened 2014-02-25
remove feedback capability (issue #65)

benjaminp:


static/sass/mq.css
```diff
@@ -525,7 +525,7 @@ html[xmlns] .slides { display: block; }
       /* We do this more explictly (the descendant selector) to avoid also styling links in the Supernav content */ }
       .main-navigation .tier-1 > a, .main-navigation .tier-2 > a {
         display: block;
-        padding: .5em 1.5em .4em 1em;
+        padding: 0.5em 1.5em 0.4em 1em;
```
birkenfeld, line 528: these changes appear unrelated?

birkenfeld:
Does this leave the small button on the right with instructions about this tracker?

static/sass/mq.css
benjaminp, line 528: Every css file is generated from a .scss file.

benjaminp:
Nope.

birkenfeld:
Do we _want_ to leave it there? :)

static/sass/mq.css
birkenfeld, line 528: So you have a different sass version that generates slightly different CSS?

benjaminp, line 528: Apparently.

benjaminp:
Note that the current doodad only serves to open the feedback form on the bottom of the page, so we'd have to make some changes to it to provide "instructions about this tracker". IMO, we should just link to the tracker from the bottom of every page of from the /about/help page.

birkenfeld:
Agreed.

Issue #204, opened 2014-02-26
The feedback form should not accept external links


johanovic:
Currently the feedback form is overwhelmed by spam. See https://github.com/python/pythondotorg/issues/65

On the short run it might suffice to make sure the form does not accept any links other than `*python.org/*`.  This will keep quite a lot of spam out.

This can be done by writing a validation method in the form.

PR #208, opened 2014-02-26
Feedbacks redux

fcurella:
This PR reinstates the feedbacks form but adds an honeypot field to the form. We could try this approach and see how much spam is curbed before going with a more aggressive approach as described in #204.

See also #65


frankwiles:
Want to double check with the PSF that they want this back on the site right now.

fcurella:
my understanding is that we don't want the feedback form, in favor of using gh issues or other means of communication. I'm closing this PR. Feel free to reopen it if I'm wrong.

PR #214, opened 2014-02-26
added API endpoints for PEPs

fcurella:


peps/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from .test_views import *
```
benjaminp: Nit: Should have a ending newline.

benjaminp:
I think it looks good, although it looks like the test doesn't cover every entry point.

fcurella:
@gutworth Tests should cover everything now, unless I missed something

PR #216, opened 2014-02-27
Fix a copy-paste error in the "only-published" API authorization class.

birkenfeld:


fcurella:
d'oh, sorry about that. good catch!

PR #217, opened 2014-02-27
Add octocat with link to site source to the footer. Closes #209

### LINKED ISSUE #209: Add a link to site source from the bottom
Site needs a link to source code. Right now there is no way people can see it is open source. It can be made with :octocat: icon and include icons for other channels used for site development.

? how can I add easy-pickling label to the issue, or who does that

techtonik:


templates/base.html
```diff
@@ -270,7 +270,14 @@ <h1 class="site-headline">
             </div> <!-- end .main-footer-links -->
 
             <div class="site-base">
-                <div class="container">
+                <div class="container" style="position: relative">
+                    <div class="collaborate" style="right: 16px; top: 48px; position: absolute;">
```
fcurella, line 274: can't we do positioning with CSS?

techtonik, line 274: I edited and tested this from Chrome development tools. I don't know how CSS is compiled and couldn't setup site on Windows. For somebody more deep knowledge, moving these styles into CSS is a breeze.

templates/base.html
```diff
@@ -270,7 +270,14 @@ <h1 class="site-headline">
...
+                        <a href="https://github.com/python/pythondotorg">
+                            <img src="{{ STATIC_URL }}github-mark-32px.png">
```
merwok, line 276: Is the github logo icon free software?  If not, I don’t think pythondotorg can include it.

(Logos of sponsors may pose the same issue; there is a difference though in that the sponsor wants the logo to be used.)

techtonik, line 276: There is an explicit permission - https://github.com/logos

techtonik:
Ping?

berkerpeksag:
Closing this per https://github.com/python/pythondotorg/issues/209#issuecomment-82591867.

PR #218, opened 2014-02-27
Fix blog time position

luanfonceca:
Defining a `position:relative;` and `top:20px;` to the `<time>` element inside the blog info.

### DIFF
--- a/static/sass/_layout.scss
+++ b/static/sass/_layout.scss
@@ -616,6 +616,11 @@
             width: columns( 4 );
         }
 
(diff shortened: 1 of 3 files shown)


tseaver:
Can you resubmit the request without the touch-everything-and-obscure-the-real-intent 
c873ec4?  "Janitorial" changes should not be mixed with "substantive" ones, because they make the review process tedious and error prone.

berkerpeksag:
Already fixed in #519.

Issue #50, opened 2014-02-20
Sample code box is too tall


munificent:
On the compound data types example, the code pushes the bottom of the box down:

![screenshot](http://i.imgur.com/I117LK2.png)

This is in Chrome on a retina Mac.

jhogue:
I can tackle this one in the next day or two. There is also some RWD funkiness that I want to address (got clobbered maybe awhile ago).

luanfonceca:
@jhogue So, You will do that?!
Cause I want to get in, help in any place so I was searching for easy issues. :)
- Just PSF members can contribute or it's open to all?

fcurella:
@luanfonceca feel free to fork and submit a pull request. Everybody can contribute! :)

krrr:
I see this bug has been fixed( #454 ), should this issue be closed?

PR #223, opened 2014-02-27
Fix slidecode height

luanfonceca:
If anyone can see if this code solves that issue, because in my point of view, is the best approach, when we fixing the height of two "side-by-side" divs, we restrict the layout and then make it more consistent.

### DIFF
--- a/static/sass/_layout.scss
+++ b/static/sass/_layout.scss
@@ -498,7 +498,7 @@
     .slide-code, .slide-copy {
         float: left;
         width: 50%;
-        min-height: 280px;
+        height: 280px;
     }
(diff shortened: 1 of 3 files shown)


berkerpeksag:
#50 is already closed by #454. Closing this. Thanks for the PR!

luanfonceca:
Ok, but in my opinion change the content doesn't fix the bug it self. Is
more like if we are hidden it.

But anyway there's no visual problems anymore.

o/

Em seg, 19 de jan de 2015 1:15 PM, Berker Peksag notifications@github.com
escreveu:

PR #263, opened 2014-03-07
added `contrib.redirects` and show pages using a fallback middleware. Close #229

### LINKED ISSUE #229: Add a way to enter redirects through the admin interface
While porting from the old site or external resources, we often find that some links no longer work because they point to locations which the new site doesn't know about.

It would be great if we could have a way of easily entering redirects for given URLs.

E.g. http://www.python.org/community/jobs/index.html should redirect to http://www.python.org/community/jobs/

At the moment these redirects have to be done via chef recipes which live in a separate repository, AFAIK.

fcurella:
The catch-all urlentry for pages shadows `RedirectFallbackMiddleware`, so I had to remove it in favor of `PageFallbackMiddleware` (which is basically `RedirectFallbackMiddleware` for `Page` models).


malemburg:
@fcurella The PR has conflicts. Could you remove those or rebase your PR ?

fcurella:
conflicts removed

malemburg:
Thanks, Flavio.

Issue #265, opened 2014-03-08
Image on community page sets the wrong tone


nedbat:
I understand that imagery is the most subjective issue to raise, but here goes:  I find the image on the Community page off-putting.  It seems religious in tone.  Python is not heaven-sent, and the point of the community is not to join together to reach for something in the sky.

Could we get a picture of people working together, talking, smiling, something like that?

birkenfeld:
I agree.

nilslindemann:
I agree. Just reverse the sun rays, so that they go bottom up instead of top down. ~~If you have a svg i can do that.~~changed my mind. Dont like this project.

jhogue:
We have a graphic forthcoming. Will be adding it as a Pull Request to the repo as soon as I can test its responsiveness at all breakpoints.

pitrou:
I agree as well.

benjaminp:
Has the responsiveness testing been completed?

PR #274, opened 2014-03-10
Quickfix for #265: remove clouds and heavenly star.

birkenfeld:
This is by no means a real fix; I just removed the "clouds" part to get rid of most of the religious connotation until the graphics people can whip up something nice.


benjaminp:
Okay until the new graphic comes I suppose.

PR #306, opened 2014-03-16
Mod: update code sample to use variable `a_list` instead of `list`

yegle:
`list` is really not a good variable name in Python :-)

### DIFF
--- a/fixtures/codesamples.json
+++ b/fixtures/codesamples.json
@@ -48,13 +48,13 @@
     "model": "codesamples.codesample",
     "fields": {
       "copy_markup_type": "html",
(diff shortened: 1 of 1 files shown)


benjaminp:
Probably this is superseded/complemented by #141?

joncle:
I would argue that `sum` is also not a good name... if `list` is being renamed to `a_list`, then `sum` should be renamed to `total` (or similar). (Depending on outcome of #141).

lepot311:
I agree with @gutworth that #141 addresses this.

berkerpeksag:
Superseded by #526.

PR #313, opened 2014-03-17
Job Board: migrations

joncle:
Migrations to:
- Make company foreign key optional
- Add company to job model
- Add company description to job model


joncle:
Needs work on Markupfields... ignore...

PR #349, opened 2014-03-25
New illustration and SCSS to support the asset

### LINKED ISSUE #265: Image on community page sets the wrong tone

jhogue:
Illustration more reflective of the Python community and the conversations we hope to have happen on the site. Fixes #265


benjaminp:
Thank you!

merwok:
Great image!

PR #366, opened 2014-04-01
Fix tooltip hovering on console launch button.

lepot311:
Tested in FF and Chromium


benjaminp:
Does any CSS need to be regenerated?

lepot311:
Yes this fix only addresses the .scss files.

benjaminp:
It would be good to have the css changes in the diff then.

lepot311:
Updated with compiled CSS

PR #370, opened 2014-04-01
Get content from SVN if no local checkout

### LINKED ISSUE #369: manage.py import_pages_from_svn fails silently if no local checkout
### Description

The local.py variable 'PYTHON_ORG_CONTENT_SVN_PATH' assumes the user has already checked out the beta.python.org data and it exists locally. This is currently hard-coded to a previous contributor's path.
### Suggestion

Since the variable requires a string, this should default to the empty string ''. Also it should be commented out by default and have a reasonable default action since many users will not have already checked the repo out.

lepot311:
Replace PYTHON_ORG_CONTENT_SVN_PATH with commented-out empty string.
Add PYTHON_ORG_CONTENT_SVN_URL variable. Defaults to beta.python.org SVN
If no local path exists, default action is to checkout to /tmp and build

Fixes #369

I'm sure this breaks on Windows. Someone with Windows experience could improve it by adding a similar temporary directory there.

### DIFF
--- a/pages/management/commands/import_pages_from_svn.py
+++ b/pages/management/commands/import_pages_from_svn.py
@@ -72,9 +72,21 @@ class Command(BaseCommand):
         page.save()
 
     def handle(self, *args, **kwargs):
-        self.SVN_REPO_PATH = getattr(settings, 'PYTHON_ORG_CONTENT_SVN_PATH', None)
+        self.SVN_REPO_PATH =      getattr(settings, 'PYTHON_ORG_CONTENT_SVN_PATH', None)
+        self.SVN_REPO_URL =       getattr(settings, 'PYTHON_ORG_CONTENT_SVN_URL',  None)
+        self.SVN_CHECKOUT_PATH =  getattr(settings, 'SVN_CHECKOUT_PATH', '/tmp')
+
         if self.SVN_REPO_PATH is None:
-            raise ImproperlyConfigured("PYTHON_ORG_CONTENT_SVN_PATH not defined in settings")
+            if self.SVN_REPO_URL:
+                import subprocess
+                with subprocess.Popen(['svn', 'co', self.SVN_REPO_URL, self.SVN_CHECKOUT_PATH]):
+                    print('Checking out from:', self.SVN_REPO_URL)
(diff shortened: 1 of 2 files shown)


benjaminp:
Thanks, but I don't think this is an command anyone should need to run again. Were you trying to use it?

lepot311:
Yes to pull the pages into the db. Is there another way to get that data?

benjaminp:
Hmm, I suppose not, but that data is of course stale now that everything is maintained for the new site.

lepot311:
It's helpful to have some data there to aid in styling. Would some lorum ipsum be preferred?

pages/management/commands/import_pages_from_svn.py
```diff
@@ -72,9 +72,21 @@ def save_images(self, content_path, page):
...
     def handle(self, *args, **kwargs):
-        self.SVN_REPO_PATH = getattr(settings, 'PYTHON_ORG_CONTENT_SVN_PATH', None)
+        self.SVN_REPO_PATH =      getattr(settings, 'PYTHON_ORG_CONTENT_SVN_PATH', None)
+        self.SVN_REPO_URL =       getattr(settings, 'PYTHON_ORG_CONTENT_SVN_URL',  None)
+        self.SVN_CHECKOUT_PATH =  getattr(settings, 'SVN_CHECKOUT_PATH', '/tmp')
+
         if self.SVN_REPO_PATH is None:
-            raise ImproperlyConfigured("PYTHON_ORG_CONTENT_SVN_PATH not defined in settings")
+            if self.SVN_REPO_URL:
+                import subprocess
```
benjaminp, line 81: Presumably this could be a toplevel import.

benjaminp:
Okay, I guess I don't mind merging it if it helps you.

lepot311:
It might help someone else in the short term, but it's almost certainly unnecessary to import this data from SVN. As someone working mostly on the visual aspect of the site, I just need some decent placeholder data. I might try adding some fixtures at some point. My only concern is it can throw a new contributor off when this data isn't present and the page layouts appear wrong.

PR #417, opened 2014-05-05
Use console email backend for local development

wbachnik:
I think local development settings should be using console email backend to cut down dependencies on local services.


frankwiles:
Good call. Thanks!

PR #473, opened 2014-08-13
Update README instructions to Python 3.4 from 3.3

venkat:


frankwiles:
While there probably aren't any issues working with the Python.org code base using 3.4, the site is currently deployed using 3.3 so the instructions should reflect that.

PR #486, opened 2014-10-13
Use Sqlite3 in development environment.

berkerpeksag:
This is more helpful for contributors, especially if they don't want to
install Vagrant.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.log
 *.pot
 *.pyc
+*.db
 __pycache__
 .vagrant
 build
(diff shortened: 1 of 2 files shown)


benjaminp:
Can't you just use the `DATABASE_URL` environmental variable?

berkerpeksag:
I can use that, but
- I need to set that variable everytime I start a new terminal session
- If I set that variable in my `.bashrc`, it may clash with another project that uses dj_database_url.
- Setting an env variable for development purposes is not a best practice in Django world. It's okay for deployment, I guess. All development settings are in `pydotorg/settings/local.py` right now, and it would be great if we keep the database settings in that file as well.

benjaminp:
I'm just worried by the fact that this changes the default for everyone.
While sqlite is nice for most things, it's good for development to be
done with the setup as similar to the production deployment as possible
(that is with PostgresSQL).

berkerpeksag:
I can add an sqlite3 entry to https://github.com/python/pythondotorg/blob/master/.travis.yml#L5, so Travis can test both databases(also, the current test coverage is good: %87).

PR #505, opened 2014-11-26
Pull in images when doing a PEP import.

### LINKED ISSUE #503: Images in peps directory don't work
Compare http://legacy.python.org/dev/peps/pep-0480/ (scroll down to "figure1") with https://www.python.org/dev/peps/pep-0480/ . The latter has a broken image.

I think you are not pulling down .png files from the peps repo. Can you fix this?

justinabrahms:
Fixes #503

There's an unfortunate implementation note in that adding `Image`s need to mutate the `Page`s. This is because the path to the image is changing and we don't know all of the images during `Page` creation due to the fact that `Image` instances require the `Page` to already be created.


frankwiles:
Awesome thanks, apparently I didn't notice any PEPs with images in them during my testing.

gvanrossum:
The author of PEPs 458 and 480 complains that these still don't show the images on www.python.org/dev/peps/, while they do show the images on the Legacy site. I pushed whitespace changes to both, and the images still don't show up (despite the last-modified date being changed, so we know the Hg changes made it out). See e.g. https://www.python.org/dev/peps/pep-0458/

frankwiles:
Yep, it's a file permissions issue on the infrastructure side of things that is keeping the images from displaying.  Create a ticket for the infrastructure team (https://github.com/python/psf-chef/issues/146)

frankwiles:
Turns out the infrastructure had moved to entirely different system, but the perms were wrong at some point when this PEP was originally pulled into the system.  Fixed with https://github.com/python/pythondotorg/commit/ba4b00dd820fae57dfc7d3636f2448e652cf6be4 

And production is happy https://www.python.org/dev/peps/pep-0480/ now.

brettcannon:
It's unhappy again (both PEP 480 and [PEP 3147](https://www.python.org/dev/peps/pep-3147/))

frankwiles:
Bah, fixed again with 862d0eb2a3c623b08ac401a4521ed4a959fa6af4

PR #518, opened 2014-12-13
Unbreak links to /getit/

### LINKED ISSUE #427: Bring back /getit workaround
On the legacy site, you can see http://legacy.python.org/getit/ is a mirror of [/download](http://legacy.python.org/download/). However, on the new site this is a 404. While I don't know that the /getit page is still required to work around the Great Chinese Firewall like it was at some point, the /getit page is still linked from some places.

justinabrahms:
Mirror the content so Chinese users can access downloads section.

Fixes #427


berkerpeksag:
Rebased and merged in https://github.com/python/pythondotorg/commit/8a5816d641f04b92820003da2dc6e09978862d45. Thanks!

PR #520, opened 2014-12-14
495 pep rss

### LINKED ISSUE #495: PEP RSS feed broken since website redesign (?)
I used to have http://www.python.org/dev/peps/peps.rss in my RSS reader, but it seems this is gone since the website redesign.

justinabrahms:
#### What's this PR do?

Loads a `peps.rss` feed generated from the peps repo into a Page object.

Additionally, Pages have an optional content type used when serving the file.
#### Where should the reviewer start?

Import script, likely.
#### How should this be manually tested?

In the peps repo: `./pep2rss.py .`
In the pythondotorg repo: `./manage.py generate_pep_pages`
Visit http://localhost:8000/dev/peps/peps.rss
#### Any background context you want to provide?

Had to change the content type thing so we could serve the rss feed with the correct mime type.
#### What are the relevant tickets?

Fixes #495

### DIFF
--- a/peps/converters.py
+++ b/peps/converters.py
@@ -210,7 +210,7 @@ def add_pep_image(pep_number, path):
 def get_peps_rss():
     rss_feed = os.path.join(settings.PEP_REPO_PATH, 'peps.rss')
     if not os.path.exists(rss_feed):
-        print("Could not find generated RSS feed. Skipping.".format(rss_feed))
+        print("Could not find generated RSS feed. Skipping.")
         return
 
     page, _ = Page.objects.get_or_create(
@@ -218,9 +218,8 @@ def get_peps_rss():
         template_name="pages/raw.html",
     )
 
-    content = ""
     with open(rss_feed, "r") as rss_content:
-        content += rss_content.read()
+        content = rss_content.read()
 
     page.content = content
     page.is_published = True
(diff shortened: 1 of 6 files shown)


peps/converters.py
```diff
@@ -206,3 +206,25 @@ def add_pep_image(pep_number, path):
         page.save()
 
     return image
+
+def get_peps_rss():
+    rss_feed = os.path.join(settings.PEP_REPO_PATH, 'peps.rss')
+    if not os.path.exists(rss_feed):
+        print("Could not find generated RSS feed. Skipping.".format(rss_feed))
```
berkerpeksag: Missing a `{}`.

peps/converters.py
```diff
@@ -206,3 +206,25 @@ def add_pep_image(pep_number, path):
...
+        return
+
+    page, _ = Page.objects.get_or_create(
+        path="dev/peps/peps.rss",
+        template_name="pages/raw.html",
+    )
+
+    content = ""
+    with open(rss_feed, "r") as rss_content:
+        content += rss_content.read()
```
berkerpeksag: `content` is already empty, so I'd use `content = ...` instead of `+=`.

justinabrahms:
Addressed your two comments. I thought there was some scoping issues with context managers, hence the `+=`. Thanks for the tip. :)

frankwiles:
LGTM! Thanks!

PR #521, opened 2014-12-14
Fix nav splits if there are only 6 items.

### LINKED ISSUE #491: "PSF Blog" text looks odd
https://www.python.org/psf/membership/

Upper right, down a few lines, no wait.. too far go up a little.. stop!

This is the HTML that needs help:

```
<li id="psf-blog" class="tier-1 element-6 last" aria-haspopup="true">
    <a class="" title="" href="http://pyfound.blogspot.com/">
        PSF Blog
    </a>
</li>
```

oh hey, screen shot....

foo

guessing this is relevant: 

bar

justinabrahms:
#### What's this PR do?

Fixes an issue where if there were only 6 items in the nav, the last cell would only occupy 1/2 of it's space.
#### Where should the reviewer start?

css.
#### How should this be manually tested?

You can see it locally at http://localhost:8000/psf-landing/
#### Any background context you want to provide?

I'm unsure on what browser support looks like for this site. I think not selectors are only back to ie8 or so. I don't think this is possible as a CSS only change.
#### What are the relevant tickets?

Fixes #491 
#### Screenshots (if appropriate)

Links to prod version in #491 
Updated:
screen shot 2014-12-14 at 3 42 19 pm

### DIFF
--- a/static/sass/_mixins.scss
+++ b/static/sass/_mixins.scss
@@ -353,7 +353,7 @@
             float: left; 
             width: 33.333333%;
             
-            &.element-6, &.element-7 { width: 16.6666665%; }
+            &.element-6:not(.unstacked), &.element-7:not(.unstacked) { width: 16.6666665%; }
             
             &.element-1 {
                 @include border-top-left-radius( 8px );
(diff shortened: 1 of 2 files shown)


berkerpeksag:
I think the branch name is wrong :) This looks like #520 to me.

justinabrahms:
Forced pushed an update. This should be the one you want. :)

PR #524, opened 2014-12-17
Update target of "IRC on Freenode" link.

berkerpeksag:
http://irc.freenode.net was not useful, so I added
/community/irc/ page instead.


benjaminp:
Maybe the label could be more helpful to the uninitiated and say "Chat on IRC" instead of "IRC on Freenode".

berkerpeksag:
Good point, updated. Thanks.

PR #525, opened 2014-12-17
Update boxes and downloads fixtures.

berkerpeksag:
boxes fixture lacks a "download-sources" box which is needed for
downloads fixture.


frankwiles:
Thanks!

PR #526, opened 2014-12-17
Add a better example of for looping through a list.

berkerpeksag:
Fixes #141 and #306.

sample

### DIFF
--- a/fixtures/codesamples.json
+++ b/fixtures/codesamples.json
@@ -1,77 +1,87 @@
 [
-  {
+{
     "model": "codesamples.codesample",
     "fields": {
-      "copy_markup_type": "html",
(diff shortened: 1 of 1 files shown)


pitrou:
Looks good to me.

benhoyt:
The commit looks good, but unfortunately on python.org it now shows the `product = product * number` line without any indentation:

image

I see the spaces in the commit [here](https://github.com/berkerpeksag/pythondotorg/commit/1baa72b083a8da2e89a23e0d15bbfc6714390c3a#diff-f55930e78135a093ab0159453b7bceb4R61), but something must be removing them in the server-side rendering, which looks like this:

``` html
                    <div class="slide-code"><pre><code><span class=\"comment\"># For loop on a list</span>
>>> numbers = [2, 4, 6, 8]
>>> product = 1
>>> for number in numbers:
... product = product * number
... 
>>> print('The product is:', product)
<span class=\"output\">The product is: 384</span></code></pre></div>
```

Is Django's templating system doing an lstrip on each line in the output or something?

PR #537, opened 2015-01-05
Fix vagrant

### LINKED ISSUE #533: RuntimeError: dictionary changed size during iteration
Using a freshly-provisioned copy on vagrant, the first time I access the site I see the message

```
RuntimeError: dictionary changed size during iteration
```

There is no indication of where this error is occurring, and the home page does nevertheless seem to appear. However, this is not a satisfactory state of affairs.

henriquebastos:
Also fixes #533

Chef was not working, so I wrote a shell script to provision the box.

`Vagrant.sh` was designed so if you run it multiple times the result will be the same.

You can find detailed explanations on the commit messages.

_NOTE_: This PR does not remove chef because I don't know if it's used elsewhere.

### DIFF
--- a/Vagrant.sh
+++ b/Vagrant.sh
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+set -o pipefail
+set -e
+set -x
+
+########
+# Update
+
+# Add ElasticSearch repository
+wget -qO - https://packages.elasticsearch.org/GPG-KEY-elasticsearch | sudo apt-key add -
+sudo add-apt-repository "deb http://packages.elasticsearch.org/elasticsearch/1.4/debian stable main"
+
+sudo apt-get update
+sudo apt-get --yes dist-upgrade
+sudo apt-get -y upgrade
+
+#######
+# Basic
+sudo apt-get -y -q install build-essential
+sudo apt-get -y -q install git
+
+##################
+#  Python
+sudo apt-get -y -q install python3-dev
+sudo apt-get -y -q install python3-setuptools
+sudo apt-get -y -q install python3-pip
+sudo apt-get -y -q install python-virtualenv
+
+##########
+# Postgres
+sudo apt-get -y -q install postgresql
+sudo apt-get -y -q install postgresql-client
+sudo apt-get -y -q install postgresql-contrib
+sudo apt-get -y -q install libpq-dev
+
+################
+# Setup Database
+
+DBUSER=vagrant
+DBPASS=vagrant
+DBNAME=pythondotorg
+DBTEST=pythondotorg_test  # Django test database
+
+# If dbuser does not exist, create it
+if [ $(sudo -u postgres psql -tc "SELECT count(*) FROM pg_user WHERE usename='${DBUSER}'") == 0 ]; then
+  sudo -u postgres psql -c "CREATE ROLE ${DBUSER} WITH PASSWORD '${DBPASS}' LOGIN SUPERUSER"
+fi
+
+if [ $(sudo -u postgres psql -tc "SELECT count(*) FROM pg_database WHERE datname='${DBNAME}'") == 0 ]; then
+  sudo -u postgres createdb -O ${DBUSER} ${DBNAME}
+fi
+
+if [ $(sudo -u postgres psql -tc "SELECT count(*) FROM pg_database WHERE datname='${DBTEST}'") == 0 ]; then
+  sudo -u postgres createdb -O ${DBUSER} ${DBTEST}
+fi
+
+######################
+# Project dependencies
+sudo apt-get -y -q install libxml2-dev
+sudo apt-get -y -q install libxslt-dev
+sudo apt-get -y -q install ruby1.9.1
+sudo apt-get -y -q install ruby-bundler
+
+sudo apt-get install -y -q openjdk-7-jre-headless
+sudo apt-get install -y -q elasticsearch
+sudo update-rc.d elasticsearch defaults 95 10 # Start ElasticSearch on boot
+sudo /etc/init.d/elasticsearch start
+
+###############
+# Setup Project
+
+# Configure bash to activate virtualenv and cd to project on login/ssh
+if ! grep -q "PYTHONDOTORG" .bashrc
+then
+  cat >> .bashrc <<EOF
+# PYTHONDOTORG: Activate venv and go to project
+source env/bin/activate
+cd pythondotorg
+EOF
+fi
+
+# Create virtualenv if don't exists
+test ! -d env && virtualenv -p python3 env
+
+source env/bin/activate
+cd pythondotorg
+
+# Install pip-accel to speed up installing packages
+# This avoid re-downloads and re-builds
+pip install pip-accel
+
+# Install project requirements
+pip-accel install -r requirements.txt
+bundle install
+
+# Apply schema
+./manage.py syncdb --noinput
+./manage.py migrate --noinput
+
+# Load data to database
+find ./fixtures -name "*.json" -exec ./manage.py loaddata {} \;
+
+# Load calendar events from the web
+./manage.py import_ics_calendars
+
+# Build search index
+./manage.py update_index
+
+# HACK: Create an admin user WITHOUT asking for password.
+python -<<EOF
+from django.core.management import setup_environ
+from pydotorg.settings import local
+setup_environ(local)
+from django.contrib.auth import get_user_model
+User = get_user_model()
+if not User.objects.filter(username='admin').exists():
+  User.objects.create_superuser('admin', 'admin@email.me', 'admin')
+EOF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,33 +1,13 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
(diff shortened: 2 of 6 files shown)


justinabrahms:
Can you provide any more information for the "chef wasn't working"? I successfully provisioned a box with vagrant last month.

henriquebastos:
It seems that the recipes are outdated missing a required attribute `name`. Also after I tried adding the attribute to many recipes it kept complaining about `aws` recipe and `xfs`.

```
(pythondotorg)pythondotorg(master)$ vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Box 'hashicorp/precise64' could not be found. Attempting to find and install...
    default: Box Provider: virtualbox
    default: Box Version: >= 0
==> default: Loading metadata for box 'hashicorp/precise64'
    default: URL: https://atlas.hashicorp.com/hashicorp/precise64
==> default: Adding box 'hashicorp/precise64' (v1.1.0) for provider: virtualbox
    default: Downloading: https://atlas.hashicorp.com/hashicorp/boxes/precise64/versions/1.1.0/providers/virtualbox.box
==> default: Successfully added box 'hashicorp/precise64' (v1.1.0) for 'virtualbox'!
==> default: Importing base box 'hashicorp/precise64'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'hashicorp/precise64' is up to date...
==> default: Setting the name of the VM: pythondotorg_default_1420338346065_62663
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
==> default: Forwarding ports...
    default: 8000 => 8000 (adapter 1)
    default: 22 => 2222 (adapter 1)
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
    default:
    default: Vagrant insecure key detected. Vagrant will automatically replace
    default: this with a newly generated keypair for better security.
    default:
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if its present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
    default: The guest additions on this VM do not match the installed version of
    default: VirtualBox! In most cases this is fine, but in rare cases it can
    default: prevent things such as shared folders from working properly. If you see
    default: shared folder errors, please make sure the guest additions within the
    default: virtual machine match the version of VirtualBox you have installed on
    default: your host and reload your VM.
    default:
    default: Guest Additions Version: 4.2.0
    default: VirtualBox Version: 4.3
==> default: Mounting shared folders...
    default: /home/vagrant/pythondotorg => /Users/henrique/workspace/pythondotorg
    default: /tmp/vagrant-chef-5/chef-solo-2/roles => /Users/henrique/workspace/pythondotorg/chef/roles
    default: /tmp/vagrant-chef-5/chef-solo-1/cookbooks => /Users/henrique/workspace/pythondotorg/chef/cookbooks
==> default: Running provisioner: chef_solo...
    default: Installing Chef (latest)...
Generating chef JSON and uploading...
==> default: Running chef-solo...
==> default: stdin: is not a tty
==> default: [2015-01-04T02:29:19+00:00] INFO: Forking chef instance to converge...
==> default: [2015-01-04T02:29:19+00:00] INFO: *** Chef 12.0.3 ***
==> default: [2015-01-04T02:29:19+00:00] INFO: Chef-client pid: 1587
==> default: [2015-01-04T02:29:20+00:00] INFO: Setting the run_list to ["role[python-dev-box]"] from CLI options
==> default: [2015-01-04T02:29:20+00:00] INFO: Run List is [role[python-dev-box]]
==> default: [2015-01-04T02:29:20+00:00] INFO: Run List expands to [apt, pythondotorg::locale, build-essential, python, ruby::1.9.1, postgresql::client, postgresql::server, git, mercurial, pythondotorg::db, pythondotorg::prereq, pythondotorg::project]
==> default: [2015-01-04T02:29:20+00:00] INFO: Starting Chef Run for precise64
==> default: [2015-01-04T02:29:20+00:00] INFO: Running start handlers
==> default: [2015-01-04T02:29:20+00:00] INFO: Start handlers complete.
==> default: [2015-01-04T02:29:20+00:00] ERROR: Running exception handlers
==> default: [2015-01-04T02:29:20+00:00] ERROR: Exception handlers complete
==> default: [2015-01-04T02:29:20+00:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
==> default: [2015-01-04T02:29:20+00:00] ERROR: Cookbook loaded at path(s) [/tmp/vagrant-chef-5/chef-solo-1/cookbooks/build-essential] has invalid metadata: The `name' attribute is required in cookbook metadata
==> default: [2015-01-04T02:29:20+00:00] FATAL: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
Chef never successfully completed! Any errors should be visible in the
output above. Please fix your recipes so that they properly complete.
```

manage.py
```diff
@@ -3,6 +3,10 @@
 import sys
 
 if __name__ == "__main__":
+    # Add argument needed (and easily forgotten) when using Vagrant.
+    if 'runserver' in sys.argv and len(sys.argv) == 2:
```
berkerpeksag, line 7: Thanks for the patch. I think this is a bad practice and shouldn't be in manage.py.

Vagrantfile
```diff
@@ -1,33 +1,13 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
- # Every Vagrant virtual environment requires a box to build off of.
-  config.vm.box = "hashicorp/precise64"
+  config.vm.box      = 'ubuntu/trusty32'
```
berkerpeksag, line 7: Is there a reason to choose trusty32 instead of trusty64?

Vagrant.sh
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env bash
```
berkerpeksag, line 1: Could you update Chef cookbooks instead? I prefer to use a tool like Chef(or a Python alternative) instead of a handwritten bash script.

manage.py
henriquebastos, line 7: You're welcome. I'm glad to help.

Could you share your reasoning on why you think this is bad practice?

I'll share my perspective so we can dialog.
1. `manage.py` is generated by django, but it's _user code_ and part of the project.
2. this change does not modify how `manage.py` works. It only improves a special case.
3. `runserver` is _only_ used during development which is what the patch tries to facilitate.
4. In earlier Django versions `manage.py` was _the entry point_ for the project, but since version 1.4 it became easier to [load django on any python module](https://docs.djangoproject.com/en/1.7/releases/1.4/#django-core-management-setup-environ).

I see 2 other alternatives to this:
1. Implementing a new management command to either override `runserver` or add a new command. Both seems overkill to me.
2. Add a `runserver.sh` that calls `manage.py runserver 0.0.0.0:8000` and reference it on the README. It's not too bad, but diverges the common `manage.py` usage and assumes people won't forget to reference the README.

Vagrantfile
henriquebastos, line 7: Good question. No reason. Just have been use it on other boxes.

I did a quick search but couldn't find perfomance data pointing to one or another. Found a couple discussions though:
- [Is a 64 or 32-bit VirtualBox guest better on a 64-bit host?](http://serverfault.com/questions/286643/is-a-64-or-32-bit-virtualbox-guest-better-on-a-64-bit-host)
- [32 or 64 guest](https://forums.virtualbox.org/viewtopic.php?f=1&t=60924)

What they _clame_ is that:
1. 32 bits applications has a smaller memory footprint.
2. theres no actual perfomance difference between both if the VM memory < 4gb.
3. 64 bits can be a bit (not significant) faster on new cpus due to some special virtualization instructions.

However theres no data. I can update it to 64 bits if you think it's best.

Vagrant.sh
henriquebastos, line 1: Sorry. Don't know Chef. :(

I've been using bash script to setup development-only boxes for single machine systems. So this patch is simply an adaptation of what I use.

However I don't intend to change your stablished workflow. I just had the problem with chef, bypassed it and I'm sharing what worked for me.

If make sense to you, I could update the patch to keep both provisioners while chef is not fixed. They would be accessible via the `--provision-with` parameter.

malemburg:
The PSF infrastructure has moved on from Chef to SaltStack. For a vagrant setup that's overkill, but perhaps it's time to move to Ansible for the provisioning ?!

It would make maintenance much easier, since we have more Python devs than Ruby devs :-)

berkerpeksag:
@malemburg I already have a WIP branch to switch to Ansible :) See https://github.com/python/pythondotorg/compare/master...berkerpeksag:switch-to-ansible

malemburg:
I think we should discuss this on pydotorg-www first, but your patch certainly look promising.

malemburg:
Kicked off a discussion there.

berkerpeksag:
Thanks for the PR! We are going to switch to use Ansible. See #827.

PR #542, opened 2015-01-11
Make legal statements and privacy policy links more visible.

berkerpeksag:


frankwiles:
This is something the PSF web team needs to make a decision on, but personally I don't think it's really something that should be done.  Those links are pretty low on the priority list and I think just distract really.

PR #573, opened 2015-02-13
remove stray print

ewdurbin:
printing to `stdout` was crashing the `uwsgi` process in production.


frankwiles:
Doh sorry about that! I could have sworn I cleaned that up before committing but apparently I didn't.

Issue #579, opened 2015-02-20
Job listing and details need to have the job title as title, not the company name


malemburg:
Both the listing and the job details page have the company name as title of the entry.

People looking for jobs will want to see the job title as headline, not the company name.

Please move the company name into the details part of the listing/entry and use the job title as headline. The company name should preferably be placed right above the company details on the job entry page.

See e.g. http://www.indeed.com/q-Python-jobs.html for an example of a job listing.

Thanks.

PR #582, opened 2015-02-24
Fix #271: Make region field optional.

### LINKED ISSUE #271: Job Board: "Region" should not be a required field
In some cases (particularly US states), this field has value, but in others a job submission will have just a City and Country. This should be allowed, rather than forcing the submitter to put something in this field.

This might require a database migration; I'm not sure.

SarahPythonista:


jobs/feeds.py
```diff
@@ -18,7 +18,8 @@ def item_title(self, item):
 
     def item_description(self, item):
         """ Description """
-        location = ",".join([item.city, item.region, item.country])
+        location = ",".join(location_part for location_part in\
```
berkerpeksag: no need to add `\` here. I'd write this something like

``` py
location_parts = (item.city, item.region, item.country)
locaiton = ",".join(location_part for location_part in location_parts
                    if location_part is not None)
```

jobs/models.py
```diff
@@ -100,7 +100,11 @@ def __str__(self):
         return 'Job Listing #{}'.format(self.pk)
 
     def save(self, **kwargs):
-        self.location_slug = slugify('%s %s %s' % (self.city, self.region, self.country))
+        location_str = ''
+        for location_part in (self.city, self.region, self.country):
+            if location_part is not None:
+                location_str = ' '.join(location_str, location_part)
```
berkerpeksag: same as above. also, `' '.join(location_str, location_part)` should be `' '.join([location_str, location_part])`

Thanks!

berkerpeksag:
Could you add a test to https://github.com/python/pythondotorg/blob/master/jobs/tests/test_models.py?

SarahPythonista:
Fixed the issues with the original commit. Added a test. 

Thanks @berkerpeksag

SarahPythonista:
I discovered while working on #579 that some of the templates also need to be edited.

I'll get to that tomorrow.

frankwiles:
Merged.  Thanks!

PR #589, opened 2015-02-27
Fix #588: Prepopulate JobForm's email when possible.

### LINKED ISSUE #588: Prepopulate JobForm's email when possible
When a logged-in user is submitting a job, the form should prepopulate their email address for convenience.

ubernostrum:
For non-authenticated users, this simply leaves the form as-is; when
an authenticated user is posting a job, however, it sets JobForm's
initial data to include the user's email address, for convenience.


frankwiles:
Thanks!

PR #592, opened 2015-02-27
Fix #271: Make region optional

### LINKED ISSUE #271: Job Board: "Region" should not be a required field

SarahPythonista:
Adjusts displayed location to accommodate cases where the region field is empty.

### DIFF
--- a/templates/jobs/job_detail.html
+++ b/templates/jobs/job_detail.html
@@ -28,7 +28,7 @@
             {% if user_can_edit %}<a href="{% url 'jobs:job_edit' pk=object.pk %}">Edit your listing</a>{% endif %}
             {% endif %}<span class="company-name">{{ object.display_name }}</span>
         </span>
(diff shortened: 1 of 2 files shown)


templates/jobs/job_detail.html
```diff
@@ -28,7 +28,7 @@ <h1 class="listing-company">
             {% if user_can_edit %}<a href="{% url 'jobs:job_edit' pk=object.pk %}">Edit your listing</a>{% endif %}
             {% endif %}<span class="company-name">{{ object.display_name }}</span>
         </span>
-        <span class="listing-location"><a href="{% url 'jobs:job_list_location' slug=object.location_slug %}" title="More jobs in {{ object.city }}, {{ object.region }}">{{ object.city }}, {{ object.region }}, {{ object.country }}</a></span>
+        <span class="listing-location"><a href="{% url 'jobs:job_list_location' slug=object.location_slug %}" title="More jobs in {{ object.display_location }}">{{object.display_location}}</a></span>
```
berkerpeksag, line 31: Just a nitpick: `{{object.display_location}}` -> `{{ object.display_location }}`

frankwiles:
Thanks!

PR #597, opened 2015-03-04
Update installation guide to Django 1.7 and move it into docs/.

berkerpeksag:
- I didn't touch the current instructions in README for now. I will update README after http://pythondotorg.readthedocs.org/ is up.
- This PR also fixes #541


webknjaz:
In the light of https://github.com/python/cpython/issues/126133, is generating the copyright year this way still the way to go?

firas9941 (approved):

PR #598, opened 2015-03-05
Resolving issue 34 on python-jobs-board.

SarahPythonista:
Adds listener to send email when new job is submitted for approval.

### DIFF
--- a/jobs/tests/test_models.py
+++ b/jobs/tests/test_models.py
@@ -1,5 +1,6 @@
 import datetime
 
+from django.core import mail
 from django.test import TestCase
 from django.utils import timezone
 
@@ -126,7 +127,6 @@ class JobsModelsTests(TestCase):
 
     def test_region_optional(self):
         job = self.create_job(region=None)
-
         self.assertEqual(job.city, "Memphis")
         self.assertEqual(job.country, "USA")
         self.assertIsNone(job.region)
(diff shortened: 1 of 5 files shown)


jobs/tests/test_models.py
```diff
@@ -127,3 +128,13 @@ def test_get_previous_approved(self):
     def test_region_optional(self):
         job = self.create_job(region=None)
 
+    def test_email_on_submit(self):
+        job = self.create_job()
+        mail.outbox = []
```
berkerpeksag, line 143: This line can be removed. Django's TestCase class already does this: https://docs.djangoproject.com/en/dev/topics/testing/tools/#emptying-the-test-outbox

jobs/tests/test_models.py
```diff
@@ -127,3 +128,13 @@ def test_get_previous_approved(self):
...
+        self.assertEqual(0, len(mail.outbox))
```
berkerpeksag, line 144: It would be good to test email subject too.

templates/jobs/email/job_was_submitted.txt
```diff
@@ -0,0 +1,7 @@
+The following job was submitted for approval:
+
+{{ content_object.display_name }}
+
+To view and review:
+
+URL:  https://{{ site.domain }}{{ content_object.get_absolute_url }}
```
berkerpeksag: Two whitespaces after "URL:".

berkerpeksag:
Could you rebase your branch with the latest master? Thanks!

jobs/tests/test_models.py
ubernostrum, line 143: Actually that line is necessary. The listener is looking for a `Job` saved with its status equal to `STATUS_REVIEW`, and that's the _default_ value of the field. So the factory calls caused by `create_job()` will pollute the outbox, and it's necessary to clear it and then do a manual save and test.

berkerpeksag, line 143: Thanks for the explanation. I think adding a comment would be helpful.

PR #612, opened 2015-03-09
Add a "Jobs" link to the header.

berkerpeksag:
There is no link to return back to the jobs list.


malemburg:
Thanks, @berkerpeksag

PR #614, opened 2015-03-10
Delete unused jQuery 1.7.1.

berkerpeksag:
We are using jQuery 1.8.2 via Google's CDN now. There is also
a hosted jquery-1.8.2.min.js in static/js/libs/ as a fallback
to Google's CDN.


frankwiles:
Good catch thanks!

PR #615, opened 2015-03-10
Delete unused OLD_script.js.

berkerpeksag:
It was added in b440b36d3e0eae3e447e6c4171e84e8d50584c87,
unused in any template file and untouched since Feb 2013.


frankwiles:
Thanks for tracking down that it really wasn't in use!

PR #626, opened 2015-03-15
Use DummySASSCompiler in DEBUG mode to fix performance issues.

berkerpeksag:
Currently, django-pipeline runs sass in every requests. Every
request took ~10 on my computer. Until https://github.com/cyberdelia/django-pipeline/issues/313
gets solved, use a dummy compiler class and run "sass --watch"
in a separate terminal.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -83,6 +83,19 @@ Finally, start the development server::
     $ ./manage.py runserver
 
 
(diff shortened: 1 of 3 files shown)


frankwiles:
Awesome thanks.  I'm going to put some comments in on the why and how to undo this in local.py just in case someone gets confused.

PR #641, opened 2015-03-17
Reorder the job entry form

malemburg:
The order now matches that of the old jobs form and more closely
resembles the order used in the job listings.

Updated some of the verbose field names and reformatted
the field definitions.

### DIFF
--- a/jobs/forms.py
+++ b/jobs/forms.py
@@ -16,17 +16,17 @@ class JobForm(ContentManageableModelForm):
     class Meta:
         model = Job
         fields = (
+            'job_title',
+            'company_name',
             'category',
             'job_types',
             'other_job_type',
-            'job_title',
-            'company_name',
-            'company_description',
             'city',
             'region',
             'country',
             'description',
             'requirements',
+            'company_description',
             'contact',
             'email',
             'url',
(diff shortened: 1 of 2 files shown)


berkerpeksag:
I think you also need to create a migration for this.

malemburg:
On 17.03.2015 18:33, Berker Peksag wrote:

> I think you also need to create a migration for this.

Are you sure ? I just reordered the form, not the model class (if I
accidentally did, that's a bug).

## 

Marc-Andre Lemburg

berkerpeksag:
Unfortunately, changing verbose_name attributes in models will create migrations. See https://docs.djangoproject.com/en/1.7/topics/migrations/#the-commands (search for "Django will make migrations for any change to your models or fields")

malemburg:
On 17.03.2015 18:50, Berker Peksag wrote:

> Unfortunately, changing verbose_name attributes in models will create migrations. See https://docs.djangoproject.com/en/1.7/topics/migrations/#the-commands (search for "Django will make migrations for any change to your models or fields")

Ah, ok. Could you add these please ?

I'm currently working on sending out reminders to all job submitters
of the last two months to resubmit using the new site.

## Thanks,

Marc-Andre Lemburg

berkerpeksag:
Sure :) Done in https://github.com/python/pythondotorg/commit/bdafa10312d050b78f23af439262d4c99d07b191.

malemburg:
On 17.03.2015 19:18, Berker Peksag wrote:

> Sure :) Done in https://github.com/python/pythondotorg/commit/bdafa10312d050b78f23af439262d4c99d07b191.

Cool :-)

## Many thanks,

Marc-Andre Lemburg

PR #659, opened 2015-03-17
Fix #657: Only show postings from the last 120 days in review panel.

### LINKED ISSUE #657: Review panel should only show postings from the last 120 days
Anything older than that will no longer be up to date.

At the moment, lots of postings fro 2014 are still shown in the panel, which is a bit confusing.

berkerpeksag:


malemburg:
Thanks, @berkerpeksag

PR #663, opened 2015-03-18
Raise Http404 if user is not authenticated.

berkerpeksag:
I just saw this on Sentry (event ID is 58540588).


malemburg:
Thanks, @berkerpeksag

PR #670, opened 2015-03-18
Only send emails for newly created Job records.

### LINKED ISSUE #656: Reduce the number of notification emails per job posting
The jobs@python.org list is currently getting three notification emails with the same content for each job posting.

The emails have date/times about a second apart, e.g.
- Date: Tue, 17 Mar 2015 19:11:04 -0000
- Date: Tue, 17 Mar 2015 19:11:05 -0000
- Date: Tue, 17 Mar 2015 19:11:06 -0000

but otherwise look pretty identical.

This may be a simple misconfiguration of the mail system, but could also be the result of a bug in the jobs app.

malemburg:
Another attempt to fix #656.

Also cleans up the signal handlers a bit and adds a global
for the jobs board email address.

### DIFF
--- a/jobs/listeners.py
+++ b/jobs/listeners.py
@@ -5,6 +5,12 @@ from django.utils.translation import ugettext_lazy as _
 from django_comments_xtd.conf import settings
 from django_comments_xtd.utils import send_mail
 
+### Globals
+
+# Python job board team email address
+EMAIL_JOBS_BOARD = 'jobs@python.org'
+
+###
 
 def on_comment_was_posted(sender, comment, request, **kwargs):
     """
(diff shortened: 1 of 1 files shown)


malemburg:
Test fails:

# 

FAIL: test_email_on_submit (jobs.tests.test_models.JobsModelsTests)

---

File "/home/travis/build/python/pythondotorg/jobs/tests/test_models.py", line 147, in test_email_on_submit

self.assertEqual(1, len(mail.outbox))

AssertionError: 1 != 0

---

malemburg:
This seems to be a bug in Django 1.7.3. See #656.

PR #671, opened 2015-03-19
Fix #656: Only send email when a job post created.

### LINKED ISSUE #656: Reduce the number of notification emails per job posting

berkerpeksag:
Use AppConfig.ready() to register signals.

### DIFF
--- a/jobs/__init__.py
+++ b/jobs/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'jobs.apps.JobsAppConfig'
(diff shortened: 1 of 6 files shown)


malemburg:
Thanks, @berkerpeksag

PR #672, opened 2015-03-19
Do not show information box if the job is already approved.

berkerpeksag:


malemburg:
Hmm, I think it's useful to be able to see the job entry status as job reviewer, so would rather not like to merge this PR.

malemburg:
Perhaps the wording in the info box could be changed to emphasize more on the status rather than way the entry is visible.

frankwiles:
I agree @malemburg

berkerpeksag:
@malemburg agreed. Do you have a suggestion for wording?

Mariatta:
I'm not a maintainer, but wondering how many status is there for a job entry?
Is it just 'approved' and 'not approved'?
What do you think about displaying the job status as badge/label/tag instead of a box with many words, where all we care about is the status of it.

malemburg:
On 16.09.2016 00:53, Mariatta wrote:

> I'm not a maintainer, but wondering how many status is there for a job entry?
> Is it just 'approved' and 'not approved'?

There are a few more states (see the job model admin page for all
of them).

> What do you think about displaying the job status as badge/label/tag instead of a box with many words, where all we care about is the status of it.

That would be nice to have. I guess the text box was simply easiest
to implement.

Removing this part from the text would likely be enough:

"You're able to see this because you're in the staff or super user group."

## 

Marc-Andre Lemburg
http://www.malemburg.com/

berkerpeksag:
Marc-Andre's comment has been addressed in https://github.com/python/pythondotorg/pull/1012.

PR #678, opened 2015-03-19
Use display_location in job_list.html.

berkerpeksag:
See "Paris, , France" at https://www.python.org/jobs/


malemburg:
Thanks, @berkerpeksag

PR #680, opened 2015-03-19
Do not call .save() twice.

berkerpeksag:


malemburg:
Thanks, @berkerpeksag

PR #686, opened 2015-03-19
Fix #679: Submissions emails should only be sent for new submissions.

### LINKED ISSUE #679: Submissions emails should only be sent for new submissions
While #656 fixed the number of emails we're getting, the problem of getting approval emails for every single change of a record still isn't fixed.

We should really only be getting emails with the subject "Job Submitted for Approval" for new submissions.

It may make sense to also get emails for changes to existing records, but those should then be using a different subject, e.g. "Job Edited", but I'm not sure whether that's really needed.

berkerpeksag:


berkerpeksag:
Great, thanks!

Issue #652, opened 2015-03-17
Job entry give 502 error when using malformed ReST


malemburg:
If a user enters wrong ReST into a description field, the record is not saved in the database and the site returns a 502 error.

The jobs app should return to the edit page and report the error instead.

berkerpeksag:
Do you have any example data to reproduce the issue? I've tried

``` rst
:ref:``

`foo``

:meth:`foo`

.. versionadded:: 3.5
```

etc, but didn't get 502 or traceback.

It only prints error messages like

```
System Message: ERROR/3 (<string>, line 8)

Unknown directive type "versionadded".
```

malemburg:
On 17.03.2015 21:20, Berker Peksag wrote:

> Do you have any example data to reproduce the issue?

This came up when trying to correct some formatting in a
submission.

Here's an example:

```
Bad ReST

* **abc*
* *abc**
* **abc**
```

If you add this to the job description, you can trigger the 502
on the staging system:

https://staging.python.org/jobs/168/edit/

## 

Marc-Andre Lemburg

berkerpeksag:
Wow, that's weird. With prod settings on my machine, I couldn't get 502. I think we really need to fix https://github.com/python/psf-salt/issues/39 to investigate such issues.

malemburg:
On 17.03.2015 21:35, Berker Peksag wrote:

> Wow, that's weird. With prod settings on my machine, I couldn't get 502. I think we really need to fix https://github.com/python/psf-salt/issues/39 to investigate such issues.

Indeed. Having access to the tracebacks would be _really_ useful :-)

## 

Marc-Andre Lemburg

malemburg:
Here's the stacktrace from sentry:

ConnectionRefusedError: [Errno 111] Connection refused

Stacktrace (most recent call last):

```
  File "django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/views/generic/base.py", line 69, in view
    return self.dispatch(request, *args, **kwargs)
  File "django/views/generic/base.py", line 87, in dispatch
    return handler(request, *args, **kwargs)
  File "django/views/generic/edit.py", line 230, in post
    return super(BaseUpdateView, self).post(request, *args, **kwargs)
  File "django/views/generic/edit.py", line 173, in post
    return self.form_valid(form)
  File "jobs/views.py", line 265, in form_valid
    form = super().form_valid(form)
[...]

PR #692, opened 2015-03-20
Do not show docutils error messages for now.

berkerpeksag:
This is a temporary workaround for #652.

Another option would be override the warning_stream option: http://docutils.sourceforge.net/docs/user/config.html#warning-stream

### DIFF
--- a/docutils.conf
+++ b/docutils.conf
@@ -1,0 +1,3 @@
+[general]
+halt_level=5
+report_level=5


malemburg:
Thanks, @berkerpeksag

berkerpeksag:
Looks like it's worked. Just tested on staging.

PR #701, opened 2015-03-21
Two fixes for the events app

berkerpeksag:
Detailed information can be found in the commit messages.


malemburg:
Thanks, @berkerpeksag 

Note that we're not using the event database directly at the moment. All events are read in from the the Google calendars (event calendars, release calendar).

frankwiles:
Just an FYI, but that's slightly incorrect.  The data is read from the Google calendars, but inserted into the events app models for display.  We're currently not using the Events models for anything else, other than those two calendars.  Just wanted to make sure everyone is clear on that.

PR #702, opened 2015-03-21
Fix #488: Use Masonry to sort columns in footer.

### LINKED ISSUE #488: Contribution links not visible in start page
On the new homepage, the only link for contributing, the developer's guide, is buried inside the docs section, with no clear delineation (other doc links are for user docs).

I think it would be nice to have a separate "Contributing" section in the navbar, with the devguide and the bug tracker (am I forgetting something? PEPs perhaps, and python-ideas?).

berkerpeksag:


malemburg:
If understand this patch correctly, it will sort the entries in the doormat footer of the site.

I'm not sure that's a good idea, since the current doormat uses the order defined in the menu structure and that order is intentional :-)

OTOH, I'm not sure whether these doormats are really needed anymore today. Google et al. are smart enough to read menus and humans would probably rather look for a sitemap if they want to get an overview of everything that's available.

Reminds me: we should probably have the "Contributing" footer as menu entry.

berkerpeksag:
Oh, I forgot to share before-after screenshots:

Before:

before

After:

after

malemburg:
Ah, so it's not about the entries, but instead about the formatting of the columns in the footer. That's fine, thanks.

berkerpeksag:
Yeah, I should have given more information about the patch. We probably need to update fixtures for local development :)

PR #730, opened 2015-04-12
Only show sign in box in PSF pages.

### LINKED ISSUE #536: Move sign in dialog to the community page
As has been pointed out, "signing in" is completely unimportant when you want to download Python, browse the docs, or even contribute or participate in the community. It should be moved out of the top page and in a dedicated box in the community page, with a text explaining that it's totally optional even for contributing.

berkerpeksag:
Fixes #536.

### DIFF
--- a/static/sass/style.css
+++ b/static/sass/style.css
@@ -1535,6 +1535,15 @@ input#s,
         top: -1.6em;
         border-color: transparent transparent #666666 transparent; }
 
+.account-signin {
+  display: none; }
+
+.account-signin-authenticated {
+  display: inline-block; }
+
+.psf .account-signin {
+  display: inline-block; }
+
 #site-map-link,
 .adjust-font-size .tier-1 > a,
 .winkwink-nudgenudge .tier-1 > a,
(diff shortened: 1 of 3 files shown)


Mariatta:
I think this works and definitely agree that the sign in box should only appear where it needs to be.

My only concern is that this creates inconsistency with the placement of the search textbox. 
In most pages the search box will be more to the right of the page, whereas in the PSF page, where the sign in page is shown, it will be slightly to the left.

Perhaps not a big deal though, but then again maybe not my call.
Is there anyone who can provide input on this?

Thanks.

berkerpeksag:
With this PR applied, screenshot of the homepage:

screenshot from 2017-08-10 23 01 34

Screenshoot of the PSF landing page:

screenshot from 2017-08-10 23 03 36

MarkMangoba:
@berkerpeksag amazing looks so much cleaner.

berkerpeksag:
I noticed a bug. I think we should still show "Your account" link in every page if user is already logged in:

screenshot from 2017-08-10 23 13 58

I've fixed it in the latest version of the PR.

PR #742, opened 2015-04-30
Delete unused Procfile.

berkerpeksag:


frankwiles:
Yeah not sure why that was there in the first place.  Thanks!

PR #743, opened 2015-04-30
Delete unused tasks.py.

berkerpeksag:
This was probably a leftover from the preview.python.org era.


frankwiles:
Yep left over from early preview days. Thanks!

PR #753, opened 2015-05-14
Clarify setup choice in the 'Installing' guide.

ronjouch:
Formerly there was 'Vagrant', followed by 'Getting Started', which left me puzzled and wondering if/why I should do both. Does this make more sense?


berkerpeksag:
That's better. Thanks!

PR #769, opened 2015-05-31
Remove feedbacks app completely.

berkerpeksag:
It was disabled partially. I think we can remove it now.


berkerpeksag:
@benjaminp @malemburg is the feedbacks app still needed or can I merge this now?

malemburg:
On 22.03.2016 18:37, Berker Peksag wrote:

> @benjaminp @malemburg is the feedbacks app still needed or can I merge this now?

I don't think this is needed anywhere (it was used for a
very short time period when the site was launched, but then
quickly shut down due to too much spam inflow).

## 

Marc-Andre Lemburg
http://www.malemburg.com/

Issue #531, opened 2015-01-02
Cannot view PEP docs on mobile Chrome and Safari


cjerdonek:
For a while I've noticed a serious problem that makes pages like the following unreadable on my mobile device:

https://www.python.org/dev/peps/pep-0440/

It happens for me with both Chrome and Safari on the latest iOS 8.1.2, though I believe it happened for me on older iOS versions, too.  An Android user has also reproduced the issue.

After the page is fully loaded, if I scroll down, say, halfway through the page, the browser will jump back to the very top.  This effectively makes it impossible to read the page.  I have never noticed this issue with any page not hosted on python.org.

smontanaro:
On Fri, Jan 2, 2015 at 10:23 AM, Chris Jerdonek notifications@github.com
wrote:

> After the page is fully loaded, if I scroll down, say, halfway through the
> page, the browser will jump back to the very top. This effectively makes it
> impossible to read the page. I have never noticed this issue with any page
> not hosted on python.org.

I tried with PEP 440 (your example), PEP 1 and PEP 8 on my Android phone. I
didn't encounter this problem with any of them. Does it only happen when
reading PEPs? Can you whittle down one of these pages to a smaller example
which demonstrates the problem?

Thx,

Skip Montanaro

cjerdonek:
Here is another page (non-PEP) that it happens on: https://www.python.org/psf-landing/

cjerdonek:
On the PSF page, for example, if I scroll down, as soon as the scrolling stops, it resets back to the top.

smontanaro:
On Fri, Jan 2, 2015 at 11:29 AM, Chris Jerdonek notifications@github.com
wrote:

> On the PSF page, for example, if I scroll down, as soon as the scrolling
> stops, it resets back to the top.

Thanks. I was able to reproduce this on my Android phone (using Chrome) as
follows. Scroll part of the way down the page. Let everything stop. Scroll
up ever-so-slightly. Zip! It scrolls all the way back to the top. I wonder
if it has anything to do with the "Back to Top" link at the bottom of the
page? I see that the non-mobile version appears not to have such a link.
Actually, it's there, but appears to be hidden:

<a id="back-to-top-1" class="jump-link" href="#python-network
<https://www.python.org/dev/peps/pep-0440/#python-network>"><span
aria-hidden="true" class="icon-arrow-up"><span>&#9650;</span></span> Back
to Top</a>

Skip

RedKrieg:
[...]

PR #788, opened 2015-06-19
fix bug number #531

nirgn975:
# Fix bug #531

**Bug:** the website jump to the nav when initiate scrolling, only on mobile.
**Solution:**
- remove the `scroll_fired = false` that made the jump from the js file.
- edit the css file so that the nav will be hidden when page load.
- edit the js file so clicking on "The Python Network" button will open the nav, and clicking on "Close" button will close the nav.

### DIFF
--- a/static/js/script.js
+++ b/static/js/script.js
@@ -231,15 +231,16 @@ $().ready(function() {
     }
     /* Animate some scrolling for smoother transitions */
 
-    /* ! Not currently working in IE10/Windows 8, Chrome for Android, Firefox (all versions)... something about the animate() function */
     $("#close-python-network").click(function() {
-        $("#close-python-network").toggle();
-        $("ul[role='tree']").toggle();
+        $("#close-python-network").slideToggle();
+        $("ul[role='tree']").slideToggle();
     });
 
     $("#python-network").click(function() {
-        $("#close-python-network").toggle();
-        $("ul[role='tree']").toggle();
+        $("#close-python-network").slideToggle({start: function(){
+        	$("#close-python-network").css("display", "list-item");
+        }});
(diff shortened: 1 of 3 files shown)


berkerpeksag:
Great, thanks! :)

Can you also commit the changes in static/sass/style.scss? It would be nice to avoid making whitespace changes, by the way.

nirgn975:
Of course, I did another commit to the branch.
I avoid whitespace (from the original files), clean up the changes I made (to be minimalist from line numbers perspective), and commit the changes in style.scss too.

Should I do another pull request?

berkerpeksag:
> Should I do another pull request?

No, this looks good. I'll take a look at the patch this weekend. Thanks again!

berkerpeksag:
Sorry for my late response. I've just played with this and looks good to me in general. I will test with different browsers tomorrow.

berkerpeksag:
By the way, can we keep the current animation? Also, the height of the "close" bar looks a bit lower on Firefox.

nirgn975:
I fixed the "close" bar, and create the slide animation.
It's works good, but there is href to "python-network" on the "close" bar that I couldn't delete, so when I press the "close" bar it jumping to "The Python Network" and then do the animation and that causing the page to slide down more then it should. So, I need some help in deleting the "href" on "close-python-network".

berkerpeksag:
To be honest, that JS file is a bit cryptic and contains many dead/unreachable/out-dated code. I used a slightly different approach and played with it in the last ~2 hours. Will merge it soon(tm) :)

berkerpeksag:
Merged in https://github.com/python/pythondotorg/commit/ddf71c63c24714abadcffed188f3a8eae342320d. Thanks @nirgn975!

Issue #756, opened 2015-05-15
OpenGraph tags highly undescriptive and non-standard


aapa:
OpenGraph metatags on the site are highly undescriptive and reduce usability (sharing, search engine indexing, etc...) of site. For example sharing any page on the site looks like this on Facebook:

![image](https://cloud.githubusercontent.com/assets/8679167/7647847/fc308100-fae3-11e4-9753-572ce0fc448c.png)

I propose that `og:title` metatag should either:
a) be removed, or
b) show same content than `<title>` tag, or
c) show same content than `<title>` tag but without  `| {{ SITE_INFO.site_name }}`

Facebook's OG debugger shows [a number of errors](https://developers.facebook.com/tools/debug/og/object?q=https%3A%2F%2Fwww.python.org%2Fdev%2Fpeps%2Fpep-0448%2F) in other opengraph tags too - mostly `Objects of this type do not allow properties named 'og:%s'.`. So the other blocks than `og:title` and maybe `og:description` should either be removed completely or be shown only when explicitly defined.

`og:url` should also be defined and it should contain canonical URL of the page.

`og:image` could be `static/img/python-logo-large.png`, for example.

This might be a duplicate of issue #59 but I'm not sure.

berkerpeksag:
Added og:title to the PEP template in https://github.com/python/pythondotorg/commit/448f2b660b356cf376643d70aa7a1acafe6c0aa6.

PR #823, opened 2015-09-17
Add og:title for each page

lepture:
Currently every page is using the same `og:title="Welcome to Python.org"`, which is not better than nothing.

I am planing to add `og:title` and `og:description` for every page. Here are some pages that I've modified. It seems that the `{% block page_title %}` shares the pattern **og_title + site_name**. I think it is better to write a macro to render them together.

### DIFF
--- a/templates/blogs/index.html
+++ b/templates/blogs/index.html
@@ -2,6 +2,7 @@
 {% load boxes %}
 
 {% block page_title %}Our Blogs | {{ SITE_INFO.site_name }}{% endblock %}
+{% block og_title %}Our Blogs{% endblock %}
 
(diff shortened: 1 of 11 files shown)


berkerpeksag:
Thanks for the PR. I agree that writing a helper would be better. See also https://github.com/python/pythondotorg/issues/756.

berkerpeksag:
Hi @lepture, will you work further on this or do you want me to review current version of the PR?

lepture:
We can make another change later. This PR has improved og titles, better than nothing.

berkerpeksag:
Thanks!

Issue #837, opened 2015-10-23
mail.python.org is dead


lac22:
If you come to #python-infra you can see what we are doing about it, but for the immediate
term we need a banner over every python.org webpage saying that we are aware that all
mail and every mailing list is not working, and we are working on it.

Been broken since about 10.00 UTC (so all day)

I do not have the skills to do this.

linsomniac:
I have a recent backup, and I can get it copied over to Digital Ocean, but I probably can't take it from there to a working system.  Can anyone familiar with the system take it once I have that on the new box?

lac22:
Yes -- the skills that we don't have are the html ones for 'making banner saying we know about this' -- the 'how to run the machine' skills we pretty much got. :)

berkerpeksag:
I'm not sure that's really necessary. Can't we add a https://status.python.org/ link to footer with some CSS?

For example(from a different site):

![screenshot from 2015-10-24 01 19 34](https://cloud.githubusercontent.com/assets/26338/10706186/5b26a77e-79ed-11e5-9605-bdbe7da6c35f.png)

lac22:
That sounds terrific as a general thing for the site to have.

As webmaster, I think that if all the mailing lists don'ẗ work for the weekend, as is possible, I
will have over 2 thousand mails to answer when the system finally comes up.  I'd like to do
something very, very visible so I don't get so many mails of 'my mailing list is broken' to 
the webmaster account which again isn't receiving mails.

ned-deily:
As an interim step, perhaps someone with posting privileges on blog.python.org could post something about it there so that it would appear in the News box on the front page of python.org?  @gutworth ?

lac22:
Great idea ned, but I think berker is about to deploy ...

berkerpeksag:
I've also implemented https://github.com/python/pythondotorg/issues/837#issuecomment-150706032 in https://github.com/python/pythondotorg/pull/840.

berkerpeksag:
I think this can be closed now :)

PR #840, opened 2015-10-28
Add an icon to footer for status.p.o

berkerpeksag:
Refs #837

Screenshot:

screenshot from 2015-10-28 09 05 06


willingc:
@berkerpeksag This change looks good to me.

PR #858, opened 2015-11-27
Disable anonymous postings by returning a 404

malemburg:


willingc:
@malemburg Good fix. :smile:

PR #859, opened 2015-11-30
Job review improvements

pneff:
A number of improvements that make reviewing more efficient. Also adds a preview function for job submissions to cut down on the syntax error postings.

### DIFF
--- a/jobs/views.py
+++ b/jobs/views.py
@@ -306,8 +306,7 @@ class JobPreview(LoginRequiredMixin, JobDetail, UpdateView):
         """
         self.object = self.get_object()
         if self.request.POST.get('action') == 'review':
-            self.object.status = 'review'
-            self.object.save()
+            self.object.review()
             return HttpResponseRedirect(self.get_success_url())
         else:
             return self.get(request)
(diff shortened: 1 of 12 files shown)


pneff:
Added a commit to fix the test run.

jobs/views.py
```diff
@@ -276,6 +376,12 @@ def get_form_kwargs(self):
             kwargs['initial'] = {'email': self.request.user.email}
         return kwargs
 
+    def get_context_data(self, **kwargs):
+        ctx = super().get_context_data()
+        ctx.update(kwargs)
+        ctx['needs_preview'] = not self.request.user.is_staff
```
berkerpeksag, line 382: Two things:
1. We have a "Job Board Admin" group for reviewers and we will switch to use it. See #768.
2. I don't think we should change the usage of the job board for reviewers. It would be good to remove other `needs_` instances, too.

berkerpeksag:
Many thanks @pneff :) I've just did a quick review, but will review and test it this week.

jobs/views.py
pneff, line 382: 1. I will switch to that role.
2. I can change that, in which case I'd introduce specific new views for the review workflow. The main issue is that while editing I want to enforce preview for normal users. For reviewers however that's not needed, as they will be redirected back to the review screen anyway.

pneff:
@berkerpeksag Have you had a chance to test further? Once I have the full feedback, I'm happy to implement the changes.

pneff:
@berkerpeksag #768 still hasn't been merged, so I can't migrate to the job board admin.

I'd really like to get this merged soonish, as the job review process is a arduous at the moment.

/cc @malemburg

malemburg:
Oh, sorry, I thought it had already been merged.

PR #878, opened 2016-02-01
Correct the default database name. Fixes #877

### LINKED ISSUE #877: Default database name is pythondotorg
The installation instructions currently give the default database name as python.org. In fact it is pythondotorg

tjguk:


berkerpeksag:
Good catch, thanks!

Issue #881, opened 2016-02-03
Google logo on PSF page is old


shalabhc:
The Google logo on https://www.python.org/psf/ is quite out of date.

See https://en.wikipedia.org/wiki/Google_logo for the updated one.

PR #485, opened 2014-10-13
Add Google logo to the "Success Stories" page.

berkerpeksag:
I got this log from the devserver:

```
[13/Oct/2014 08:56:42] "GET /static/img/google-trans.png HTTP/1.1" 404 1664
```

And see https://www.python.org/success-stories/ for production.

PR #887, opened 2016-02-14
replaced with the new google logo

creativcoder:
In response to #881.


berkerpeksag:
Thanks for the pull request.

If I remember correctly, in order to merge this, we need to get permissions from companies. See https://github.com/python/pythondotorg/pull/485 for example. @malemburg can give us a better information :)

berkerpeksag:
Thanks!

Issue #911, opened 2016-03-23
What apps require specific PostgreSQL features and why


techtonik:
Reticketed from #907. I think that it is important to explain people things that they can not figure out themselves (see #910). The thing that I can not figure out (and find appropriate code pieces) is where PostreSQL features are required and why.

berkerpeksag:
Like I said before, it's the jobs app that uses PostgreSQL specific features. You can see it by changing the database setting to use SQLite and opening `/jobs/` URL in your browser.

PR #912, opened 2016-03-23
Explain the codebase of python.org website

### LINKED ISSUE #907: Documented specific application requirements
From https://github.com/python/pythondotorg/issues/903#issuecomment-199224793 it is clear that the web site is not a standard Django application - it requires some PostgreSQL specific features, there is a requirement to have Ruby installed, deployment is done through some complicated mechanism.

In ideal world this is documented with a diagram. Maybe there are GSoC students with steAm background that can take the task?..

techtonik:
Closes #907. Addresses #911.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,4 +11,10 @@
 * Mailing list: [pydotorg-www](https://mail.python.org/mailman/listinfo/pydotorg-www)
 * IRC: `#pydotorg` on Freenode
 * Staging site: https://staging.python.org/ (`master` branch)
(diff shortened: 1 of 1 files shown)


berkerpeksag:
This all can be learnt by reading the [Django tutorial](https://docs.djangoproject.com/en/1.9/intro/tutorial01/).

Like I said many times before, dependencies are already documented if you'd like to install them manually and Vagrant box already installs them automatically. Please do not waste our time.

Issue #989, opened 2016-08-13
Issue with sqlparse 0.2


Mariatta:
I followed instructions in https://pythondotorg.readthedocs.io/install.html 

I then started my development server using `manage.py runserver`

When I opened my browser, I saw the following error.

```
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
Internal Server Error: /
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/site-packages/django/core/handlers/base.py", line 204, in get_response
    response = middleware_method(request, response)
  File "/usr/local/lib/python3.4/site-packages/debug_toolbar/middleware.py", line 89, in process_response
    new_response = panel.process_response(request, response)
  File "/usr/local/lib/python3.4/site-packages/debug_toolbar/panels/sql/panel.py", line 187, in process_response
    query['sql'] = reformat_sql(query['sql'])
  File "/usr/local/lib/python3.4/site-packages/debug_toolbar/panels/sql/utils.py", line 28, in reformat_sql
    return swap_fields(''.join(stack.run(sql)))
  File "/usr/local/lib/python3.4/site-packages/sqlparse/engine/filter_stack.py", line 29, in run
    stream = filter_.process(stream)
TypeError: process() missing 1 required positional argument: 'stream'
[12/Aug/2016 14:47:54] "GET / HTTP/1.1" 500 101972
```

I googled the error message, and came across this issue https://github.com/jazzband/django-debug-toolbar/issues/856

One of the suggestion there was to add `sqlparse <0.2` into requirements.txt

I did that and now the page can be loaded.

Perhaps related to https://github.com/python/pythondotorg/issues/987

berkerpeksag:
I'd prefer avoiding adding an indirect dependency to `requirements.txt`. But since we can't upgrade `django-debug-toolbar` to 1.5 (it requires Django 1.8+), I'd be +1 on adding `sqlparse < 0.2` into `requirements.txt` as an interim solution.

PR #990, opened 2016-08-13
add sqlparse<0.2

Mariatta:
added sqlparse<0.2 to requirements.txt to address https://github.com/python/pythondotorg/issues/989

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -43,3 +43,4 @@ uWSGI==2.0.10
 
 raven==5.2.0
 django-waffle==0.11.1
+sqlparse <0.2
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Could you add a TODO comment and explain why we need to add this (something like "delete this when we switch to Django 1.8 and DDT 1.5")?

requirements.txt
```diff
@@ -43,3 +43,4 @@ uWSGI==2.0.10
 
 raven==5.2.0
 django-waffle==0.11.1
+sqlparse <0.2
```
berkerpeksag: Nitpick: sqlparse <0.2 -> sqlparse<0.2

Mariatta:
Thanks for the feedback @berkerpeksag 
Added todo comment and removed the space before < sign

berkerpeksag:
Thanks!

Issue #985, opened 2016-08-11
python.org does not validate


fingolfin:
The W3C validator is rather unhappy with python.org, as one can see here: https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.python.org%2F

Now, I don't believe in "validation for validation's sake", but at least some of these look potentially serious to me --  in particular, the invalid "role" attributes. As these are used for screen readers, I wonder if this breaks screen some screen readers, and thus might impair accessibility?

Fixing those shouldn't be hard ... well, at least to fix the HTML -- I have no idea how this site is implemented, so perhaps actually making the change is much harder.

berkerpeksag:
Thanks! Yes, most of them look easy.

Riyuzakii:
@berkerpeksag I'm a first timer on github and i would like to help fix this bug, could you direct me to the files to be changed and maybe tell a little bit more about the issue.

berkerpeksag:
Thank you for your interest for contributing to python.org. Unfortunately, there is already an open pull request to fix this: #992. Feel free to grab another issue!

vivekvj18:
@berkerpeksag 
What this above task is all about...?
I Would like you to explain the changes to be made....!

karlafalcao:
Hello guys, I would like to contribute to pythondotorg frontend issues. I would like to know if this PR is still valid?

fingolfin:
If you want to contribute, surely a good first step would be to check by yourself whether it is still valid (by running the W3C validator, and/or by checking the source code of the site in the web browser of your choice and comparing it with the changes in the proposed but unmerged pull request that is linked elsewhere on this issue). And then post your findings

uncleSlayer:
I have checked the validator and it still shows bugs and warnings. If nobody is working can this issue be assigned to me? I am a beginner and want to work on this.

PR #991, opened 2016-08-18
Validate html

Mariatta:
Attempt to address issue https://github.com/python/pythondotorg/issues/985
- removed `braile, embossed, tty` value from `media` attribute since they've been deprecated
- replaced inline table attributes: `cellpadding, cellspacing, width, border`, used css 
- adjust values for `role` attribute 

These two errors remained, but I'm not familiar with these. Any suggestion of what to do here?
`Bad value cleartype for attribute http-equiv on element meta.`
`Bad value imagetoolbar for attribute http-equiv on element meta.`

### DIFF
--- a/pydotorg/settings/pipeline.py
+++ b/pydotorg/settings/pipeline.py
@@ -19,7 +19,7 @@ PIPELINE_CSS = {
         ),
         'output_filename': 'stylesheets/mq.css',
         'extra_context': {
-            'media': 'not print, braille, embossed, speech, tty',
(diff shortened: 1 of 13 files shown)


berkerpeksag:
I think your branch is not up-to-date with master: a4de366 and 2c145bf are already merged into master in https://github.com/python/pythondotorg/commit/d5790b07f5a04d709e5b311dbdf2e5d9eccb3890.

I'll take a look at the patch this week, thanks!

Issue #1004, opened 2016-09-11
In python.org/about, Latest News More link should point to blog.python.org


Mariatta:
In https://www.python.org/about/ at the bottom there is a Latest News box section with a link to view More.
When clicked, it keeps me in the same page.
I believe it should point to blog.python.org
<img width="572" alt="screen shot 2016-09-11 at 4 58 44 pm" src="https://cloud.githubusercontent.com/assets/5844587/18421517/46b477d2-7841-11e6-8813-cf4937612c7e.png">

PR #1005, opened 2016-09-12
Link the latest news in about page to blog.python.org

Mariatta:
fix issue #1004

### DIFF
--- a/pydotorg/context_processors.py
+++ b/pydotorg/context_processors.py
@@ -25,9 +25,6 @@ def get_host_with_scheme(request):
 
 
 def blog_url(request):
-    """
-    Return the url for python blog
-    """
     return {
-        'BLOG_URL': settings.PYTHON_BLOG_URL
+        'BLOG_URL': settings.PYTHON_BLOG_URL,
     }
--- a/pydotorg/context_processors.py
+++ b/pydotorg/context_processors.py
@@ -22,3 +22,12 @@ def get_host_with_scheme(request):
     return {
         'GET_HOST_WITH_SCHEME': request.build_absolute_uri('/').rstrip('/'),
     }
+
+
+def blog_url(request):
+    """
+    Return the url for python blog
+    """
+    return {
+        'BLOG_URL': settings.PYTHON_BLOG_URL
+    }
--- a/pydotorg/tests/test_context_processors.py
+++ b/pydotorg/tests/test_context_processors.py
@@ -26,8 +26,11 @@ class TemplateProcessorsTestCase(TestCase):
         mock_request = MockRequest(path='/nothing/here/')
         self.assertEqual({}, context_processors.url_name(mock_request))
 
+    def test_blog_url(self):
+        """
+        test the blog_url context processor
+        :return:
+        """
         mock_request = MockRequest(path='/about/')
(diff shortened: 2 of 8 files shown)


berkerpeksag:
Good catch. Another way to fix this would be adding a context processor to https://github.com/python/pythondotorg/blob/master/pydotorg/context_processors.py.

Mariatta:
@berkerpeksag Thank you, I added blog_url in context_processor

pydotorg/tests/test_context_processors.py
```diff
@@ -24,3 +25,9 @@ def test_url_name(self):
 
         mock_request = MockRequest(path='/nothing/here/')
         self.assertEqual({}, context_processors.url_name(mock_request))
+
+        mock_request = MockRequest(path='/about/')
```
berkerpeksag, line 30: Can you move the test into a new `test_blog_url` method?

pydotorg/tests/test_context_processors.py
```diff
@@ -24,3 +25,9 @@ def test_url_name(self):
...
+        self.assertEqual({'BLOG_URL': settings.PYTHON_BLOG_URL}, context_processors.blog_url(mock_request))
+
+        mock_request = MockRequest(path='/')
```
berkerpeksag: We can remove this one. I know you added it to test `IndexView` case, but since we are unit testing the context processor itself, we can safely remove the `path='/'` test.

pydotorg/tests/test_context_processors.py
Mariatta: Yes that makes sense. I will move this.

Mariatta:
@berkerpeksag I adjusted the test as suggested. Thanks.

pydotorg/tests/test_context_processors.py
```diff
@@ -24,3 +25,12 @@ def test_url_name(self):
 
         mock_request = MockRequest(path='/nothing/here/')
         self.assertEqual({}, context_processors.url_name(mock_request))
+
+    def test_blog_url(self):
+        """
+        test the blog_url context processor
+        :return:
```
berkerpeksag: It looks like this was accidentally added :) Also, we can remove the whole docstring. The test itself is pretty understandable.

Mariatta: Yeah Pycharm does this haha. Ok cool I will remove docstring in this case

pydotorg/context_processors.py
```diff
@@ -22,3 +22,12 @@ def get_host_with_scheme(request):
     return {
         'GET_HOST_WITH_SCHEME': request.build_absolute_uri('/').rstrip('/'),
     }
+
+
+def blog_url(request):
+    """
+    Return the url for python blog
```
berkerpeksag: We can remove this one too.

pydotorg/context_processors.py
```diff
@@ -22,3 +22,12 @@ def get_host_with_scheme(request):
...
+    """
+    return {
+        'BLOG_URL': settings.PYTHON_BLOG_URL
```
berkerpeksag: add trailing comma

Mariatta:
@berkerpeksag docstrings removed, and added the trailing comma as suggested. 
Thanks.

berkerpeksag:
Thanks!

Issue #1002, opened 2016-09-10
Delete https://www.python.org/blog/


brettcannon:
The 3.3.0 release is a bit old. 😄

berkerpeksag:
+1, we should probably redirect `/blog/` to http://blog.python.org/

PR #1006, opened 2016-09-12
Redirect python.org/blog to blogs.python.org

Mariatta:
For #1002

### DIFF
--- a/blogs/tests/test_views.py
+++ b/blogs/tests/test_views.py
@@ -1,6 +1,7 @@
 from django.core.management import call_command
 from django.core.urlresolvers import reverse
 from django.test import TestCase
+from django.conf import settings
 
 from boxes.models import Box
 
@@ -32,3 +33,12 @@ class BlogViewTest(TestCase):
 
         latest = BlogEntry.objects.latest()
         self.assertEqual(resp.context['latest_entry'], latest)
+
+    def test_blog_redirects(self):
+        """
+        Test that when '/blog/' is hit, it redirects to blog.python.org
+        """
+        response = self.client.get('/blog/', follow=True)
+        self.assertRedirects(response,
+                             settings.PYTHON_BLOG_URL,
+                             status_code=301)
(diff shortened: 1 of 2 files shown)


berkerpeksag:
Thank you for working on this. Can you also add a test to `blogs/tests/test_views.py`?

Mariatta:
@berkerpeksag Thanks for the feedback. I added a test case.

pydotorg/urls.py
```diff
@@ -33,7 +33,7 @@
     url(r'^downloads/', include('downloads.urls', namespace='download')),
     url(r'^doc/$', TemplateView.as_view(template_name="python/documentation.html"), name='documentation'),
     #url(r'^community/$', TemplateView.as_view(template_name="python/community.html"), name='community'),
-    url(r'^blog/$', TemplateView.as_view(template_name="python/blog.html"), name='blog'),
+    url(r'^blog/$', RedirectView.as_view(url=settings.PYTHON_BLOG_URL)),
```
berkerpeksag: Can you add `permanent=False` to `RedirectView.as_view()`? I know some people want to move away from Blogger so we may still use `/blog/` in the future.

blogs/tests/test_views.py
```diff
@@ -32,3 +33,12 @@ def test_blog_home(self):
 
         latest = BlogEntry.objects.latest()
         self.assertEqual(resp.context['latest_entry'], latest)
+
+    def test_blog_redirects(self):
+        """
+        Test that when '/blog/' is hit, it redirects to blog.python.org
+        """
+        response = self.client.get('/blog/', follow=True)
```
berkerpeksag: `follow=True` can be removed.

pydotorg/urls.py
berkerpeksag: Wait, on second thought, I think this should be redirected to `/blogs/` instead of blogs.python.org. `/blog/` was probably used as a showcase during development.

berkerpeksag: So we need to pass `permanent=True` (instead of `False`) to be forward compatible with future Django versions.

Mariatta:
@berkerpeksag Thanks. I adjusted this. So `/blog/` now redirects to `/blogs/` and I also added `permanent=True` flag as suggested.

berkerpeksag:
Thanks!

Issue #428, opened 2014-05-08
docutils admonitions no longer stand out


briancurtin:
The change to the new site seemed to have dropped whatever style changes were making admonitions stand out from regular text. For example, the top of https://www.python.org/download/releases/2.7 is currently just plain "Note:" text pointing to the latest release, but changing it to ".. Note::" doesn't seem to matter - makes it worse, actually. Ideally that text would stand out more to make it seen that it's not the latest release.

The note at the bottom of https://www.python.org/psf/codeofconduct/ also used to standout.

Mariatta:
I just made a pull request for adding box around the Note in code of conduct page.
However, as pointed out in the ticket, the Note field in the download page is not using `.. Note::` directive, but rather just a plain note. Therefore it will remain as is.
I'm wondering if the document should be updated to use the `.. Note::` directive?

PR #1020, opened 2016-09-27
Add styling to admonitions

Mariatta:
- added box border and background color

This is to address #428 and python/peps#109

If `.. Note::` directive is used in page, then it will be rendered in a box

example screenshots:
for code of conduct page
screen shot 2016-09-26 at 9 24 48 pm

for pep 339
screen shot 2016-09-26 at 9 23 07 pm

### DIFF
--- a/static/sass/style.scss
+++ b/static/sass/style.scss
@@ -2164,11 +2164,11 @@ $colors: $blue, $psf, $yellow, $green, $purple, $red;
 div.note {
     background-color: #eee;
     border: 1px solid #ccc;
-    padding: 5px 10px;
+    padding: px2em( 5px ) px2em( 10px );
 }
 
 p.admonition-title {
-    margin: 0px 10px 5px 0px;
+    margin-bottom: px2em( 5px );
     font-weight: bold;
 }
 
--- a/static/sass/style.scss
+++ b/static/sass/style.scss
@@ -2161,6 +2161,15 @@ $colors: $blue, $psf, $yellow, $green, $purple, $red;
     background-color: #808080;
 }
 
+div.note {
+    background-color: #eee;
+    border: 1px solid #ccc;
+}
+
+p.admonition-title {
+    margin: 0px 10px 5px 0px;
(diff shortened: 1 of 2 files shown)


gvanrossum:
Nice, but shouldn't there be a little bit of a left margin inside the box?

Mariatta:
Thanks for the feedback @gvanrossum.
I updated the code. These are updated screenshots.

screen shot 2016-09-27 at 9 07 25 am
screen shot 2016-09-27 at 9 07 04 am

gvanrossum:
Perfect! I don't have commit access; maybe Better can merge?

static/sass/style.scss
```diff
@@ -2161,6 +2161,16 @@ $colors: $blue, $psf, $yellow, $green, $purple, $red;
...
 }
 
+div.note {
+    background-color: #eee;
+    border: 1px solid #ccc;
+    padding: 5px 10px;
+}
+
+p.admonition-title {
+    margin: 0px 10px 5px 0px;
```
berkerpeksag: Right margin is not needed here. I'd just use `margin-bottom: 5px`.

berkerpeksag: Also, using `px` would break rendering in other devices. Please use the `px2em()` helper.

berkerpeksag (changes requested):

berkerpeksag:
It looks like `.text p` adds some bottom margin so it would be nice to add something like

``` css
div.note > p {
    margin-bottom: 0;
}
```

to make the white-space more balanced. This way you may also be able to remove the margin attribute in `p.admonition-title`.

Mariatta:
Thanks @berkerpeksag. 
I use px2em() as suggested, and also set the bottom margin to 0

Here are the updated screenshots

screen shot 2016-09-27 at 9 49 22 am

screen shot 2016-09-27 at 9 49 17 am

berkerpeksag (approved):

berkerpeksag:
Looks great, thanks!

PR #1054, opened 2017-02-18
removing the last blank line

abhaystoic:
Just a small readability improvement. Improved support for Python 2 and 3.


berkerpeksag:
Thanks for the PR, but python.org doesn't support Python 2 and LICENSE is taken from http://www.apache.org/licenses/ directly so I rather keep it as-is.

PR #1056, opened 2017-02-25
No left sidebar template

MarkMangoba:
No left sidebar template.


templates/psf/noleft.html
```diff
@@ -0,0 +1,48 @@
+{# ===== PSF Default Content PAGE ===== #}
+
+{% extends "base.html" %}
```
berkerpeksag: I think extending `templates/psf/default.html` would make the template simpler. You'd only need to override the `content_attributes` and `left_sidebar` template blocks:

```jinja
{# Note that I haven't a chance to test this. #}

{% extends "psf/default.html" %}

{% block content_attributes %}{% endblock %}

{% block left_sidebar %}{% endblock %}
```

MarkMangoba: thanks @berkerpeksag i'll test this out.

berkerpeksag:
Is this PR still needed? In any case, I just tried out the idea I left as a review comment at https://github.com/python/pythondotorg/pull/1056/files/64a4e6d450a94e749f7578bfe746aafc31c6b976#diff-658cb47699426b14d6afc0cbaa52c83e and it worked. Also, I renamed it to `noleftsidebar.html`.

PR #1063, opened 2017-03-16
Add support for syntax highlighting in PEPs

### LINKED ISSUE #889: Add some love to the code snippets on the pep pages
I'd like to see syntax highlighting via http://pygments.org/ or a frontend javascript library.

Also don't wrap code snippets on mobile.

```
pre.literal-block {
    white-space: pre;
    overflow-x: scroll;
    word-wrap: normal;
}
```

ilevkivskyi:
Fixes #889 

This PR provides two options for syntax highlighting in PEPs:
* Automatic highlighting of all literal blocks using Python syntax for PEPs in ``PURE_PYTHON_PEPS``. I added few examples there just to illustrate the idea.
* Now there is support for reST ``.. code:: your_favourite_language``. Such literal blocks will be highlighted using the corresponding syntax.

This PR requires [Pygments](https://pypi.python.org/pypi/Pygments) 2.1.3.

I added a CSS based on Pygments default (I made very few modifications, to match the "general spirit" of the site).


Mariatta:
Thanks @ilevkivskyi , this seems to work well :)
I tested this locally, using the change made in https://github.com/python/peps/pull/229

Here's how the colors look like
screen shot 2017-03-22 at 5 37 48 pm

templates/pages/pep-page.html
```diff
@@ -25,6 +25,7 @@
 {% endblock %}
 .
 {% block content %}
+<link rel="stylesheet" type="text/css" href="/static/stylesheets/pygments_default.css" />
```
berkerpeksag: We should use `{{ STATIC_URL }}` instead of hardcoding `/static/`.

requirements.txt
```diff
@@ -44,3 +44,5 @@ uWSGI==2.0.10
 raven==5.2.0
 django-waffle==0.11.1
 sqlparse<0.2  #TODO: delete this when we switch to Django 1.8 and DDT 1.5
+
+pygments==2.1.3
```
berkerpeksag: Could you add a comment so we won't forget to remove the dependency when we move PEPs to RtD?

peps/converters.py
```diff
@@ -160,6 +165,21 @@ def convert_pep_page(pep_number, content):
     # Fix PEP links
     pep_content = BeautifulSoup(data['content'])
     body_links = pep_content.find_all("a")
+    # Fix hihglighting code
```
berkerpeksag: Typo: hihglighting

peps/converters.py
```diff
@@ -7,11 +7,16 @@
...
+from pygments import highlight
+from pygments.lexers import PythonLexer
+from pygments.formatters import HtmlFormatter
+
 from pages.models import Page, Image
 
 PEP_TEMPLATE = 'pages/pep-page.html'
 pep_url = lambda num: 'dev/peps/pep-{}/'.format(num)
 
+PURE_PYTHON_PEPS = [483, 484, 526]
```
berkerpeksag: This is probably not important to have a constant. I'd say either inline it or add a comment to explain why we need it.

peps/converters.py
```diff
@@ -160,6 +165,21 @@ def convert_pep_page(pep_number, content):
     # Fix PEP links
     pep_content = BeautifulSoup(data['content'])
     body_links = pep_content.find_all("a")
+    # Fix hihglighting code
+    code_blocks = pep_content.find_all("pre", class_="code")
```
berkerpeksag: Nitpick: I know we are already pretty inconsistent with this, but please use single quotes where possible.

berkerpeksag (changes requested):
Thanks for the PR! How hard would be to add some tests?

ilevkivskyi:
@berkerpeksag Thank you for review! I addressed your comments (including a simple test) in a new commit.

ilevkivskyi:
This is still marked as "Changes requested", although I think I had implemented the required changes a month ago. Should anything more be changed here?

berkerpeksag:
Yeah, that's because I haven't had a chance to look at the new version of the PR yet.

berkerpeksag:
Thanks for the PR again, @ilevkivskyi. Your changes look good to me, but unfortunately I have to reject this PR. Our ultimate goal is to move PEPs from python.org to Read the Docs so my preference is only fixing bugs in python.org infrastructure and save feature requests to the new infrastructure (also, note that we are going to get this feature for free when we move to RtD)

brettcannon:
Maybe we should open this again since getting the PEPs over to Read the Docs is moving at the speed of mud?

ilevkivskyi:
> Maybe we should open this again since getting the PEPs over to Read the Docs is moving at the speed of mud?

I am fine with both possibilities: if it stays closed -- less works for me, if it is reopened -- we will have syntax highlighting sooner :-)

hugovk:
> Maybe we should open this again since getting the PEPs over to Read the Docs is moving at the speed of mud?

Three years on: I agree, it'd be great get colouring in whilst waiting on RtD (this big recent effort has stalled: https://github.com/python/peps/pull/1385). I find highlighted code much easier to read.

I've rebased and re-opened this as PR https://github.com/python/pythondotorg/pull/1638.

PR #1075, opened 2017-04-28
Correct comment on true (not "classic") division

benhoyt:
On the third code sample on [www.python.org](https://www.python.org/), "Intuitive Interpretation", there is this code:

```python
# Python 3: Simple arithmetic
>>> 1 / 2
0.5
>>> 2 ** 3
8
>>> 17 / 3  # classic division returns a float
5.666666666666667
>>> 17 // 3  # floor division
5
```

However, `17 / 3` is not actually "classic" division -- this is Python 3, so it's "true" division. This is as per [PEP 238](https://www.python.org/dev/peps/pep-0238/), which states that "Classic division will remain the default in the Python 2.x series; true division will be standard in Python 3.0."

So this PR just changes the comment "classic division" to "true division".


berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1096, opened 2017-07-13
Change the Download pages to display links for Python 3 first

CianciuStyles:


berkerpeksag (approved):

berkerpeksag:
Thanks! The content of the "Sources" box at https://www.python.org/downloads/ actually comes from Django admin so I just copied and pasted your changes in `templates/downloads/download-sources-box.html`.

PR #1109, opened 2017-07-28
bpo-30992: Remove Barry's obsolete MD5-signed key.

ned-deily:
Current versions of gpg refuse to import MD5 keys by default, rejecting
the whole file.  The only such key:
pub  1024R/ED9D77D5 1997-12-08 Barry A. Warsaw <bwarsaw@cnri.reston.va.us>
was used to sign releases older than 2.6.  If needed, that key can still
be found via the link to it (http://barry.warsaw.us/barrypub-gpg.asc)
on the downloads page (https://www.python.org/downloads/).
(Will update the website text after the file is deployed.)


berkerpeksag (approved):

berkerpeksag:
We should probably cherry-pick this commit into the release branch instead of doing the usual "merge master into release branch" way since the current state of master may a little bit unstable.

ned-deily:
@berkerpeksag OK, I can do that; I'll wait a bit to give @warsaw a chance to review.

warsaw (approved):
Yeah, that key is super ancient and I haven't used it for years.  I should really update my home page to de-emphasize the old keys, but I won't remove them, so that link should stay good.

PR #1111, opened 2017-07-28
bpo-30992: Remove Barry's obsolete MD5-signed key.

ned-deily:
Current versions of gpg refuse to import MD5 keys by default, rejecting
the whole file.  The only such key:
pub  1024R/ED9D77D5 1997-12-08 Barry A. Warsaw <bwarsaw@cnri.reston.va.us>
was used to sign releases older than 2.6.  If needed, that key can still
be found via the link to it (http://barry.warsaw.us/barrypub-gpg.asc)
on the downloads page (https://www.python.org/downloads/).

(cherry picked from commit ede581dbb8107759c3775a7b481722d2c39598b2)


berkerpeksag (approved):

ned-deily:
Thanks, @berkerpeksag.  What does it take to get this deployed live?

ned-deily:
It looks like it is deployed already.  I'll update the web page and we should be done with this.

Issue #560, opened 2015-02-02
Calendar event pages for full day event show (wrong) times


malemburg:
See e.g. https://www.python.org/events/python-events/246/

This is an event for the days March 28 and 29. The calendar entry uses all day entries for this, without specifying times on each day. Yet the python.org site displays "From 28 March at 12am UTC through 30 March at 12am UTC, 2015", so it's 12 hours off.

The all day event should really only show up as "From 28 March through 29 March".

The iCal entry for the events starts with:

BEGIN:VEVENT
DTSTART;VALUE=DATE:20150328
DTEND;VALUE=DATE:20150330
DTSTAMP:20150202T092425Z
...

without timezone information.

malemburg:
This problem keeps popping up again and again for full day events (all conferences we add to the Google calendars), e.g. most recently here:

https://www.python.org/events/python-events/467/

The event is from April 3-7 and shown correctly in the Google calendar widget:

http://www.pycon.org/#calendar

malemburg:
This bug still exists: https://www.python.org/events/python-events/800/ vs. http://www.pycon.org/#calendar

The problem is that full day events should be displayed using times (or time zones for that matter), but just the dates, which are correctly entered into the Google calendars.

berkerpeksag:
IIRC, this hasn't been deployed to production yet. Could you please check it at staging.python.org?

malemburg:
Looks fine for EuroPython on staging: https://staging.python.org/events/python-events/794/

So I guess all that needs to be done is deploy on production :-)

Thanks, Berker !

hugovk:
This is now in production, for example https://www.python.org/events/python-events/1738/ reads:

> From 18 Sept. through 20 Sept., 2024

with no clock times, as expected.

Issue #375, opened 2014-04-04
Events are shown one day longer than scheduled


malemburg:
See e.g. https://www.python.org/events/python-events/?page=2

PyCon UK is from Sept 19-22, but the listing shows Sept 19-23.

I guess this has to do with timezones again. Note that most events in the events calendar are full day events.

jacobb:
This actually seems to be a misunderstanding with the ics format than timezones -- from what I can tell, the .ics file stores full-day events with a DTEND value of (to use the given example):

DTSTART;VALUE=DATE:20140919
DTEND;VALUE=DATE:20140923

That is, they "end" (exclusively) at midnight the day after the last day of the event.

Best way to solve this, I think, is to see if DTEND is a date (as opposed to a datetime), and if so, backtrack the end of the event by one day.

Would like to get a confirmation on this before merging, but I'll write up code for it and tag this issue tonight.

malemburg:
This is most likely related to #560.

CianciuStyles:
I have submitted a PR with a possible fix in https://github.com/python/pythondotorg/pull/1161.

PR #1161, opened 2017-10-10
Fix issue 375

CianciuStyles:
This PR is for a possible fix for the isse #375.

I chose to implement a filter for removing the ending day under some circumstances, so that if the events already present in the database are setup correctly (especially the all_day property) no models should be touched at all.

### DIFF
--- a/blogs/tests/test_parser.py
+++ b/blogs/tests/test_parser.py
@@ -11,7 +11,7 @@ class BlogParserTest(unittest.TestCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.test_file_path = get_test_rss_path()
-        cls.entries = get_all_entries("file://{}".format(cls.test_file_path))
+        cls.entries = get_all_entries("file:///{}".format(cls.test_file_path))
 
     def test_entries(self):
         self.assertEqual(len(self.entries), 25)
--- a/events/importer.py
+++ b/events/importer.py
@@ -4,7 +4,7 @@ import pytz
 import requests
 
 from .models import EventLocation, Event, OccurringRule
-from .utils import convert_dt_to_aware
+from .utils import extract_date_or_datetime
 
 DATE_RESOLUTION = timedelta(1)
 TIME_RESOLUTION = timedelta(0, 0, 1)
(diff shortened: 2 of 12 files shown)


berkerpeksag (commented):
Thank you! Your approach looks good to me, but using a template filter might be problematic in the future (e.g. someone may forget to use it) Can't we modify `dt_start` and `dt_end` properties instead if `all_day` is `True`? That way we will also avoid moving more logic to Django templates.

CianciuStyles:
That was my idea initially, but I went this road instead to avoid a migration for the events which have been already parsed and stored in the database. How would you modify dt_start and dt_end to achieve the same result?

Another way to possibly make the change a little bit more future-proof is to extract the part of the template dealing with the event start and end times as a different partial template (similar to the time_tag one), so that we should only include it wherever we need it (obviously, this won't address your comment on moving more logic to the templates).

Let me know what you think about this.

berkerpeksag:
I was thinking moving the logic in templates and templatetag into the properties in the models, for example:

```py
def dt_end(self):
    ...
    if self.all_day:
        return recurrence - timedelta(days=1)
    ...
```

Then we would do the following in the templates:

```jinja
{% if next_time.all_day %}
    {{ next_time.dt_end|date:"d N" }}
{% else %}
    {{ next_time.dt_end|date:"d N" }} at {{ next_time.dt_end|date:"fA"|lower }} {{ next_time.dt_end|date:"e" }}
{% endif %}
```

I don't mind running a data migration if needed. Currently, there 601 events in the database and a data migration would run quickly without any problem.

berkerpeksag:
Also, looking at this again, perhaps we should fix the importer instead. We probably need to do the `dt - timedelta(days=1)` step here if `all_day` returns `True`:

https://github.com/python/pythondotorg/blob/fea2f399557ec32e3b094901692ed110099ebf34/events/importer.py#L28-L47

CianciuStyles:
If we fix the importer we should also run a migration for applying the same logic to the already existing events in the database (subtract `timedelta(days=1)` from `dt_end` for all the events for which `all_day is True`), as `dt_end` on `OccurringRule` is a field, not a property).

Other than that, this sounds sensible to me. I'll try and change the code according to what we have discussed here.

berkerpeksag:
Could you also add a test for #560?

```ical
BEGIN:VEVENT
DTSTART;VALUE=DATE:20150328
DTEND;VALUE=DATE:20150330
DTSTAMP:20150202T092425Z
```

Unless I'm missing something, this will also fix #560.

berkerpeksag:
Superseded by #1167.

PR #1164, opened 2017-10-21
remove membership roster

MarkMangoba:
incorrect link listed, remove from sponsors block, membership roster is showing, although the sponsors link is listed.


MarkMangoba:
@bwaliszewski per issue reported by @malemburg - re: doesn't directly link to the list of sponsors.

PR #1165, opened 2017-10-25
Fix: non featured events are always included in get_events_upcoming

mariocj89:
Just realized that `qs.filter(featured=True)` does not modify `qs` but returns the filtered result instead.


berkerpeksag (approved):
Good catch, thank you!

Issue #1162, opened 2017-10-18
New events workflow


datapythonista:
At the moment, the process to create new events to python.org is as described next:
* Events are sent to events@python.org with all the information
* A moderator reviews it, and if appropriate, the event is manually created in a Google calendar
* A cron job exists that pull the events from the calendars into the website

This can be made more efficient, by changing the workflow to:
* Events are created using a form, or imported from trusted sources (e.g. certain meetup.com groups)
* The events have a flag "Pending approval" that make them not visible
* A dashboard exists where moderators can approve events with a single click
* Once an event is approved:
    * It becomes visible in the website
    * It is exported to the appropriate Google calendar
    * An email is sent to events@python.org

To implement this new workflow, next tasks are required:
* Add a new "Approved" flag to the Events model (TICKET PENDING)
* Create a form in python.org to submit an event (TICKET PENDING)
* Create a dashboard that lists pending events, where a user with permissions can approve them (TICKET PENDING)
* Create a command to export an event to a Google calendar (TICKET PENDING)
* Create a command to import events from meetup.com (TICKET PENDING)

datapythonista:
@berkerpeksag, @malemburg, I created this ticket with what we discussed, please feel free to add comments if I misunderstood anything, or if you think we can do better than what I described.

Once we're all happy with this, I'll create individual tickets for each task.

Thanks!

malemburg:
This won't really work, since for review we have to have easy
access to email for feedback to the submitter to fix submissions.
The job subsystem is a good example of the web based review
system not really working in practice.

I would prefer to keep the approval workflow via the mailing list
and only have the subsystem on python.org make this more admin
friendly. We could, for example, add a link to the email to
trigger the addition to the calendar.

It's not a good idea to make the events tables on python.org
the main source of truth - this should always be the calendars.

malemburg:
Some more feedback from emails...

Regarding the events: We are managing these using Google Calendars
outside the python.org website. The website just pulls the entries
from those calendars in using a cronjob.

Since the Google calendars are much better to integrate elsewhere,
it is probably not wise to add events directly and only to the
python.org website database.

Now, as for automatically adding events: this will not be possible,
since we have to review events listed on the website before switching
them live. We do have to do some quality control on these to make
sure the events are placed on the right calendars, we don't get
lots of commercial training events listed or other abuse for commercial
interests.

What would help is a form on the python.org which takes the
event listing as required by us and sends us an email to the
[...]

PR #1166, opened 2017-10-25
New flag that allows to control the visibility of events

mariocj89:
The new flag "visible" allows controlling whether an event is listed in the events section of the site. This allows to temporary unlist events. This feature allows to build on to a better workflow where events can be created automatically and approvers only have to make them visible.

This issue is a blocker for other issues that will be tackling in the next London Python Sprint as part of #1162 

I've split some long lines in `events/views.py` happy to undo it if you prefer those.

Thanks!


mariocj89:
This might actually not be needed anymore after th conversation via email. It can still have some value in case you want to temporarily hide an event. But otherwise we can just close it

berkerpeksag:
Thank you for the PR, @mariocj89. I agree that we can close this one for now.

PR #1167, opened 2017-10-29
Fix issue 375 #2

CianciuStyles:
Hello,

I'm sorry this took so long. I am opening a new pull request for closing issue #375, in which I addressed the approach we discussed in the other PR (https://github.com/python/pythondotorg/pull/1161).

I have also added a test for issue #560, so hopefully this should fix that as well.


blogs/tests/test_parser.py
```diff
@@ -11,7 +11,7 @@ class BlogParserTest(unittest.TestCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.test_file_path = get_test_rss_path()
-        cls.entries = get_all_entries("file://{}".format(cls.test_file_path))
+        cls.entries = get_all_entries("file:///{}".format(cls.test_file_path))
```
berkerpeksag: Did you get a test failure without this change?

events/migrations/0006_auto_20171029_1524.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('events', '0005_auto_20170821_2000'),
+    ]
+
+    operations = [
+        migrations.RunPython(exclude_ending_day, include_ending_day)
```
berkerpeksag: Style nit: Please add trailing comma.

events/tests/test_importer.py
```diff
@@ -20,7 +20,7 @@ def setUpClass(cls):
 
     def test_injest(self):
         importer = ICSImporter(self.calendar)
-        with open(EVENTS_CALENDAR) as fh:
+        with open(EVENTS_CALENDAR, encoding='utf-8') as fh:
```
berkerpeksag: This shouldn't be necessary. What's the output of the following command in your system?

```py
locale.getpreferredencoding(False)
```

events/tests/test_importer.py
```diff
@@ -94,3 +103,68 @@ def test_modified_event(self):
...
+        self.assertEqual(
+            make_aware(datetime(year=2016, month=4, day=2)),
+            e.next_or_previous_time.dt_start
+        )
+        self.assertEqual(
+            make_aware(datetime(year=2016, month=4, day=3)),
+            e.next_or_previous_time.dt_end
+        )
+
+        ical = """\
```
berkerpeksag: Can we move the new tests to separate test methods?

events/tests/test_importer.py
```diff
@@ -94,3 +103,68 @@ def test_modified_event(self):
...
+        self.assertTrue(e4.next_or_previous_time.single_day)
+        self.assertEqual(
+            make_aware(datetime(year=2013, month=8, day=2, hour=20)),
+            e4.next_or_previous_time.dt_start
+        )
+        self.assertEqual(
+            make_aware(datetime(year=2013, month=8, day=2, hour=20, minute=30)),
+            e4.next_or_previous_time.dt_end
+        )
+
```
berkerpeksag: Style nit: Please delete the extra line.

peps/converters.py
```diff
@@ -33,7 +33,7 @@ def convert_pep0():
     for a Python.org Page returns the core body HTML necessary only
     """
     pep0_path = os.path.join(settings.PEP_REPO_PATH, 'pep-0000.html')
-    pep0_content = open(pep0_path).read()
+    pep0_content = open(pep0_path, encoding="utf-8 ").read()
```
berkerpeksag: I don't think this change is in scope of #375.

events/migrations/0006_auto_20171029_1524.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.5 on 2017-10-29 15:24
+from __future__ import unicode_literals
+
+import datetime
+
+from django.db import migrations
+
+
+def exclude_ending_day(apps, _):
```
berkerpeksag: I'd replace `_` with `schema_editor` and use it in:

```py
...
db_alias = schema_editor.connection.alias
...
Model.objects.using(db_alias).filter(...)
```

events/migrations/0006_auto_20171029_1524.py
```diff
@@ -0,0 +1,32 @@
...
+    OccurringRule = apps.get_model('events', 'OccurringRule')
+    for occurring_rule in OccurringRule.objects.filter(all_day=True):
```
berkerpeksag: Can't we do this without a for loop?

```py
from django.db.models import F

OccurringRule.objects.filter(all_day=True).update(dt_end=F('dt_end') - datetime.timedelta(days=1))
```

berkerpeksag (changes requested):
Could you also rename `0006_auto_20171029_1524.py` to a more readable name so we can easily see that it's a data migration?

blogs/tests/test_parser.py
CianciuStyles: I do indeed...
```
  File "C:\Users\cianciustyles\Documents\Code\Python.org\blogs\tests\test_parser.py", line 17, in test_entries
    self.assertEqual(len(self.entries), 25)
AssertionError: 0 != 25
```
but I can completely drop the first commit with the changes to make the tests pass on my host, if this already works for you.

events/tests/test_importer.py
CianciuStyles: On my Windows 10 laptop, the result is the following:
```
>>> import locale
>>> locale.getpreferredencoding(False)
'cp1252'
```
As in a previous comment, I think it would be better for the time being to drop completely these changes, and only commit the ones about the event importer with tests.

events/migrations/0006_auto_20171029_1524.py
CianciuStyles: I was not aware this was possible, I'm happy to change it.

events/tests/test_importer.py
berkerpeksag: I've never tried to run python.org tests under Windows so feel free to send those changes as a separate PR :)

CianciuStyles:
I think I have addressed all the changes you requested. If I forgot anything or some more changes are required, just let me know!

berkerpeksag (approved):

berkerpeksag:
Thanks!

Issue #1124, opened 2017-08-12
Issue on dropdown list


ghost:
I found that on [https://www.python.org/jobs](https://www.python.org/jobs) there is dropdown list **Types** on main navigation bar which is not fixed. If there are more than 10-12 items in the list, then links in list goes out of screen. please hover the mouse on **types** option on **_navigation bar._**

levlaz:
I am not sure what the issue that you are describing is. If there are more items than fit on the screen they still show up as you scroll down. The *Locations* list is very similar since there are many entries.

Am I missing something?

ghost:
There are two ways to scroll: 1) By using a mouse 2) By using touchpad on a laptop
Consider a situation, when a user is using a laptop and using a touchpad. To scroll the browser's window, either user can use scrollbar or can use **two finger scrolling** on a laptop. If two finger scrolling is not working then user has only one option **"use scrollbar to scroll".**

This drop-down menu appears only then when we hover on it. So if a user hovers on these options then he cannot scroll and see more items that don't fit on the screen because he can scroll only then, when he uses the scrollbar.

So if we make fit all the menu items on the window, then this problem can be resolved.

We can use a scrollbar in the drop-down menu to make fit all of them on the browser's window.

ghost:
Now I have fixed it. Extract [fixedDropDown.zip](https://github.com/python/pythondotorg/files/1432153/fixedDropDown.zip) and  see **index.html** .

**Thank you**

levlaz:
Thank you for explaining the use case. I admit I have been a bit spoiled by two finger scroll and I have not considered this situation. It makes prefect sense.

levlaz:
I think the fix is great. You should open a PR with those same changes against the code base since that is not a static page.

berkerpeksag:
Thank you for the fix, @prateek89760. @levlaz is correct, could you please convert your changes to a pull request? You can find our installation guide at https://pythondotorg.readthedocs.io/install.html

PR #1183, opened 2017-11-04
Fixed issue #1124

ghost:
Fixed drop-down menu. The description can be found here.

 issue https://github.com/python/pythondotorg/issues/1124

### DIFF
--- a/static/sass/mq.css
+++ b/static/sass/mq.css
@@ -2528,3 +2528,11 @@ html[xmlns] .slides { display: block; }
 
 @viewport {
   width: device-width; }
+
+
+.element-4 ul,
+.element-2 ul{
+  height:288px;
+  overflow:hidden;
+  overflow-y:scroll;
+}


berkerpeksag (changes requested):
Thank you for the PR. `mq.css` is an automatically generated file. Please modify `static/sass/mq.scss` instead.

static/sass/mq.css
```diff
@@ -2528,3 +2528,11 @@ html[xmlns] .slides { display: block; }
 
 @viewport {
   width: device-width; }
+
+
+.element-4 ul,
+.element-2 ul{
+  height:288px;
```
berkerpeksag, line 2535: We don't use px as unit. Please use the `px2em` helper.

Also, is there a way to fix this without set `<ul>`'s height to a fixed value?

static/sass/mq.css
```diff
@@ -2528,3 +2528,11 @@ html[xmlns] .slides { display: block; }
 
 @viewport {
   width: device-width; }
+
+
+.element-4 ul,
```
berkerpeksag, line 2533: We support multiple devices so this should be moved into one of the applicable media blocks like

```scss
@media (min-width: px2em( ... )) {

}
```

berkerpeksag (changes requested):

ghost:
I have made changes and made a pull request, but it has gone to the new pull request.
Link: https://github.com/python/pythondotorg/pull/1209

berkerpeksag:
Superseded by #1209.

PR #1187, opened 2017-11-05
Add docker support

pablogsal:
This PR adds support for Docker. This is intended to easy the initial set-up of the environment and requirements for recurring and first-time contributors alike. The advantage of using this approach is that the developer does not need to install and configure postgresql, ruby, python3.4 and all python packages locally, effectively simplifying the installation and set up of the environment. Also, if the environment is not needed, is very easy to clean.

To install all dependencies and start the containers:

1. Install docker and docker-compose

2. `docker-compose up --build`

The first set up will take a while, but after that everything will be cached and therefore new instantiation of the containers will take much less time.

3. Access the webpage on `localhost:8000`.

4. Profit! :moneybag: 

If you want to change anything in the project and see the result, just re-run `docker-compose up --build` again. Take into account that this will NOT reset the database. If you want to start **truly** clean, run `docker-compose rm` first to remove the previous containers.

------

One of the main targets of using this approach is to make much easier to set up everything for new developers and contributors, no mater if they use Windows, Linux or MacOS as long as they can run docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+FROM ubuntu:14.04
+
+RUN apt-get update \
+    && apt-get install -y software-properties-common \
+    && add-apt-repository ppa:deadsnakes/ppa
+
+RUN apt-get update \
+    && apt-get upgrade -y \
+    && apt-get install -y \
+    build-essential \
+    ca-certificates \
+    gcc \
+    git \
+    libpq-dev \
+    make \
+    mercurial \
+    pkg-config \
+    python3.4 \
+    python3.4-dev \
+    ssh \
+    libxml2-dev \
+    libxslt1-dev \
+    ruby \
+    && apt-get autoremove \
+    && apt-get clean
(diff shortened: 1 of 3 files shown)


berkerpeksag (commented):
Thank you for the PR, but we already have a way to install development environment without installing all dependencies in contributor's system: https://pythondotorg.readthedocs.io/install.html#easy-setup-using-vagrant

I personally prefer Vagrant + Ansible setup since it doesn't require users to upgrade their Kernel version frequently (I can't run Docker in my system because its Kernel version is too old for Docker -- I definitely won't upgrade it just because of Docker) and they are more stable than Docker.

Nathanator:
Yeah, I too think Docker is too heavyweight for most tasks.  Ever tried to get Docker working out of the box on Ubuntu?  It's not trivial or easy.  If fixing a typo is harder than git clone && nano, we're going to suffer, IMHO

PR #1199, opened 2017-11-21
adding view analytics.txt

MarkMangoba:
adding view analytics.txt for Google url, python.org/analytics.txt


berkerpeksag (commented):
Could you also move analytics.txt into templates/ directory?

PR #1219, opened 2018-01-24
Remove python2 cta

### LINKED ISSUE #1218: Python 3 only for Call To Action download buttons.
#RE: https://www.python.org/downloads/

I don't think it should be done immediately... but perhaps when https://pythonclock.org read 2 years or less.

screen shot 2018-01-23 at 9 02 40 pm

The current scenario only breeds confusion and indicates that "choosing" between Python 2 and Python 3 is even a thing anymore.

Perhaps removing the 2.7.x button, and replacing 

> Wondering which version to use? Here’s more about the difference between Python 2 and 3.

with 

> Looking for Python 2.7 releases? <a href="link to 2.7 releases">Click here</a>

mpirnat:
Fixes #1218...

**Remove the Python 2 CTA buttons from the downloads page**

As described in the original issue.

**Remove 'which version' text from download-banner box fixture**

As described in the original issue, this may have to be manually effected in production since it's in the "box" stuff that appears to be managed in the CMS (and was how I had to do it locally for the change to take effect). It seemed reasonable to refresh the fixture though.

With these changes, it the banner area on the downloads page looks like this:

downloads-banner

**Remove the Python 2 CTA buttons from the downloads supernav**

Not described in the original issue but seemed like a good idea for consistency.

With this change, the downloads supernav looks like this:

downloads-supernav

**Remove Python 2 from homepage downloads box**

Not described in the original issue, this also seemed like a good idea for consistency since the "wondering which version" text that the original issue is asking to remove also currently appears in the downloads box on the homepage. Again, I had to update the box in the CMS for the change to take effect, but the fixture has been updated to remove both the "which version" text as well as the link to the latest Python 2 version--this last seemed in the spirit of the CTA button removal.

With this change, the downloads box on the homepage looks like this:

homepage-downloads-box

**Not Undertaken**

There may be some opportunities to clean up the downloads view code to remove queries to get the latest Python 2 versions, which would trim a few ms, but I didn't feel comfortable ripping those out yet in case something else needed them or if the resulting change would be more extensive than desired. This can certainly be explored though.

### DIFF
--- a/fixtures/boxes.json
+++ b/fixtures/boxes.json
@@ -387,7 +387,7 @@
 {
     "fields": {
         "label": "homepage-downloads",
-        "content": "<h2 class=\"widget-title\"><span aria-hidden=\"true\" class=\"icon-download\"></span>Download</h2>\n<p>Python source code and installers are available for download for all versions! Not sure which version to use? <a href=\"https://wiki.python.org/moin/Python2orPython3\">Check here</a>.</p>\n<p>Latest: <a href=\"/downloads/release/python-352/\">Python 3.5.2</a> - <a href=\"/downloads/release/python-2712/\">Python 2.7.12</a></p>",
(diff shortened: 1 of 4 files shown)


ewdurbin:
By the decree of @gvanrossum I hereby merge this. I'll update the CMS content after deploy @mpirnat.

PR #1234, opened 2018-03-23
Some HTTP to HTTPS.

JulienPalard:
While shaving a yak I though it was a good idea to proofread some old links.


users/tests/test_models.py
```diff
@@ -60,12 +60,12 @@ def test_create_usergroup(self):
         group = UserGroup.objects.create(
             name='PLUG',
             location='London, UK',
-            url='http://meetup.com/plug',
+            url='https://meetup.com/plug',
```
berkerpeksag: This is just a test, no need to change.

templates/python/inner.html
```diff
@@ -623,11 +623,11 @@ <h1 id="form">Form Example</h1>
 
 {% block left_sidebar %}
 <aside class="left-sidebar" role="secondary">
-  
+
     {% include "components/navigation-widget.html" %}
 
     <div class="twitter-widget sidebar-widget">
-        <script type="text/javascript" src="http://widgets.twimg.com/j/2/widget.js"></script>
+        <script type="text/javascript" src="https://widgets.twimg.com/j/2/widget.js"></script>
```
berkerpeksag: It would better to use the same style in both this file and `templates/waitforit.html` (`//...` was used instead of `https://...` there)

templates/python/inner.html
```diff
@@ -623,11 +623,11 @@ <h1 id="form">Form Example</h1>
 
 {% block left_sidebar %}
 <aside class="left-sidebar" role="secondary">
-  
+
```
berkerpeksag: Unrelated change.

pages/tests/fake_svn_content_checkout/about/success/dlink/content.rst
```diff
@@ -8,7 +8,7 @@ D-Link Australia Uses Python to Control Firmware Updates
 :Author: How Wong
 :Contact: grandsonata at aol dot com
 :Date: 2007/08/02 10:00:00
-:Web site: http://www.dlink.com.au/
+:Web site: https://www.dlink.com.au/
```
berkerpeksag: This is used by the test suite, no need to change.

pages/tests/fake_svn_content_checkout/about/content.ht
```diff
@@ -31,29 +31,29 @@ Python is powerful... and fast
 ------------------------------
 
 Fans of Python use the phrase "batteries included" to describe the
-`standard library <http://docs.python.org/3/library/>`_, which covers
+`standard library <https://docs.python.org/3/library/>`_, which covers
```
berkerpeksag: This file is used by the test suite, no need to change.

codesamples/factories.py
```diff
@@ -42,7 +42,7 @@ def initial_data():
             <p>Experienced programmers in any other language can pick up Python very
             quickly, and beginners find the clean syntax and indentation structure
             easy to learn.
-            <a href=\"http://docs.python.org/3.3/tutorial/\">Whet your appetite</a>
+            <a href=\"https://docs.python.org/3/tutorial/\">Whet your appetite</a>
```
berkerpeksag, line 45: This file creates data for development purposes. No need to change.

blogs/tests/test_models.py
```diff
@@ -9,7 +9,7 @@ class BlogModelTest(TestCase):
     def test_translation(self):
         t = Translation.objects.create(
             name='Swiss',
-            url='http://python.ch',
+            url='https://python.ch',
```
berkerpeksag: This is just a test, no need to change.

blogs/tests/test_models.py
```diff
@@ -21,7 +21,7 @@ def test_blog_entry(self):
             title='Test Entry',
             summary='Test Summary',
             pub_date=now,
-            url='http://www.revsys.com',
+            url='https://www.revsys.com',
```
berkerpeksag: This is just a test, no need to change.

berkerpeksag (changes requested):

users/tests/test_models.py
JulienPalard: I agree, I however though it would be good to have some https in tests just in case it break something. But it should better be separated tests in this case so I revert this one OK.

codesamples/factories.py
JulienPalard: Why not having up-to-date links in dev too?

berkerpeksag: The test data in codesamples/factories.py is basically for frontend developers to see the slider in action, but sure let's update it.

berkerpeksag (approved):

berkerpeksag:
Thanks.

PR #1241, opened 2018-04-18

stevepiercy:
In the footer/sitemap of the site, there are right borders only on "Python Books" and "Python Logo". Removing the probably unintended line break makes this style consistent with the other `.tier-1.element-N` styles.

### DIFF
--- a/static/sass/mq.css
+++ b/static/sass/mq.css
@@ -1888,8 +1888,7 @@ html[xmlns] .slides { display: block; }
       clear: none;
       border-left: 0;
       border-right: 0; }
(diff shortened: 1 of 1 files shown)


stevepiercy:
See PR #1242

berkerpeksag (changes requested):
Thanks for the PR!

It would be better to modify both files (`mq.css` and `no-mq.css`) in the same PR.

Also, note that `mq.css` and `no-mq.css` are generated files. We'll need to make these changes in `_layout.scss` instead.

stevepiercy:
Yeah, I hoped to do it all via GitHub's UI so I wouldn't have to clone it locally. Oh, well!

Regardless, I didn't know where the .css file came from, so I'll update `_layout.scss` instead in a single patch. I've submitted a separate PR #1244, so I will close these two related PRs. Thank you!

PR #1242, opened 2018-04-18

stevepiercy:
In the footer/sitemap of the site, there are right borders only on "Python Books" and "Python Logo". Removing the probably unintended line break makes this style consistent with the other `.tier-1.element-N` styles.


stevepiercy:
See PR #1241

PR #1244, opened 2018-04-18

stevepiercy:
In the footer/sitemap of the site, there are right borders only on "Python Books" and "Python Logo". Removing the probably unintended line break makes this style consistent with the other .tier-1.element-N styles.

Closes #1241 and #1242.


berkerpeksag (approved):

berkerpeksag:
Thank you for the PR, @stevepiercy. Sorry for my late response, I've tested this and generated CSS file.

PR #1245, opened 2018-04-19
Update URL to PyPI (hopefully)

### LINKED ISSUE #1243: Update PyPI URL
There are many instances of `https://pypi.python.org/` in the code, but I'm not sure what they all do. The new URL is `https://pypi.org/`. Making updates would save a redirect.

Anyway, the one I wanted to update is that which appears in the uppermost navigation bar. I think it's this one based on its context, but I'm not positive.
https://github.com/python/pythondotorg/blob/master/cms/views.py#L6

If someone can verify, I can update that one. I don't know whether others need to be updated.

stevepiercy:
- Partially resolve #1243. There may be other places where the URL for PyPI needs to change from https://pypi.python.org/ to https://pypi.org/

### DIFF
--- a/cms/views.py
+++ b/cms/views.py
@@ -3,7 +3,7 @@ from django.shortcuts import render
 from urllib.parse import urljoin
 
 LEGACY_PYTHON_DOMAIN = 'http://legacy.python.org'
-PYPI_URL = 'https://pypi.python.org/'
+PYPI_URL = 'https://pypi.org/'
 
 
 def legacy_path(path):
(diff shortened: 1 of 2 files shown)


pradyunsg:
Another place: https://github.com/python/pythondotorg/blob/0d99645d28b8e3935c0a5e83fd096a3ab59af70b/docs/source/administration.rst#L9

(9th line)

stevepiercy:
This search found a couple more instances, but I don't know whether I should submit commits for them. Please let me know:
https://github.com/python/pythondotorg/search?utf8=%E2%9C%93&q=%22pypi.python.org%22&type=

berkerpeksag (approved):

berkerpeksag:
Thank you and I'm sorry for missing this PR earlier.

PR #1272, opened 2018-06-04
Change default database for local development to sqlite. Changed defa…

Tomasz-Kluczkowski:
Modified default database in local.py to sqlite. This should allow users who do not wish to set up a fully fledged postgres database to quickly set up the project with sqlite database which will sit in the root of the project.
For access to previous settings the postgres setting is left commented out should the user want to choose this type of the database.

We should also add the following to the developers instructions as some people get lost when they need to create a postgres database.

We assume postgres is already installed.

1. Switch to base postgres account to create a database
- $ sudo su - postgres
2. Get into postgres shell
- postgres@<your pc name>:~$ psql
3. Create pythondotorg database
- postgres=#  createdb pythondotorg -E utf-8 -l en_US.UTF-8

4. Quit postgres shell
postgres-# /q
5. Go back to your own user account
- postgres@<your pc name>:~$ exit

### DIFF
--- a/pydotorg/settings/local.py
+++ b/pydotorg/settings/local.py
@@ -9,9 +9,12 @@ INTERNAL_IPS = ['127.0.0.1']
 # For example,
 # PYTHON_ORG_CONTENT_SVN_PATH = '/Users/flavio/working_copies/beta.python.org/build/data'
 PYTHON_ORG_CONTENT_SVN_PATH = ''
+DATABASE_URL = os.path.join(BASE, 'test_sqlite.db')
 
 DATABASES = {
-    'default': dj_database_url.config(default='postgres:///pythondotorg')
+    # 'default': dj_database_url.config(default='postgres:///pythondotorg')
+    # 'default': dj_database_url.parse('postgres:///pythondotorg')
+    'default': dj_database_url.parse('sqlite:////' + DATABASE_URL)
 }
 
 HAYSTACK_CONNECTIONS = {
--- a/pydotorg/settings/local.py
+++ b/pydotorg/settings/local.py
@@ -9,7 +9,7 @@ INTERNAL_IPS = ['127.0.0.1']
 # For example,
 # PYTHON_ORG_CONTENT_SVN_PATH = '/Users/flavio/working_copies/beta.python.org/build/data'
 PYTHON_ORG_CONTENT_SVN_PATH = ''
-DATABASE_URL = os.path.join(BASE, 'test_sqlite.db')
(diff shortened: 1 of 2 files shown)


pydotorg/settings/local.py
```diff
@@ -9,9 +9,12 @@
 # For example,
 # PYTHON_ORG_CONTENT_SVN_PATH = '/Users/flavio/working_copies/beta.python.org/build/data'
 PYTHON_ORG_CONTENT_SVN_PATH = ''
+DATABASE_URL = os.path.join(BASE, 'test_sqlite.db')
```
berkerpeksag: Django will add the `test_` prefix when running the test suite so I'd suggest renaming the database name to `pythondotorg.db`.

berkerpeksag (commented):
> For access to previous settings the postgres setting is left commented out should the user want to choose this type of the database.

Unfortunately, this will make the tests fail on master. What do you think about converting this to a documentation PR? We could add subsection in https://github.com/python/pythondotorg/blob/master/docs/source/install.rst#manual-setup and say something like "use sqlite3 if you want to work on frontend and/or stuff like the downloads app." and add the following snippet:

```py
DATABASE_URL = os.path.join(BASE, 'pythondotorg.db')

DATABASES = {
    # 'default': dj_database_url.config(default='postgres:///pythondotorg')
    # 'default': dj_database_url.parse('postgres:///pythondotorg')
    'default': dj_database_url.parse('sqlite:////' + DATABASE_URL)
}
```

Or we can add another setting file named `pydotorg/settings/local_sqlite3.py`:

```py
from .local import *  # We just need override the DATABASES setting.

DATABASE_URL = os.path.join(BASE, 'pythondotorg.db')

DATABASES = {
    'default': dj_database_url.parse('sqlite:////' + DATABASE_URL),
}
```

Tomasz-Kluczkowski:
I assume then that setting sqlite as the default option in the code is not good.
Ok I will add some more modifications to install.rst with better instructions on how to create a postgres database and put this snippet in there as an alternative for easy setup.

berkerpeksag:
> I assume then that setting sqlite as the default option in the code is not good.

Unfortunately, yes :( The only way to do it by default is to replace all `distinct()` calls in `jobs/` with something that works on both PostgreSQL and SQLite. Otherwise, we need to provide SQLite as an alternative for people who don't plan to work on the job board.

berkerpeksag:
@Tomasz-Kluczkowski can we now close this and continue our discussion over #1274?

Tomasz-Kluczkowski:
Closed as addding information to docs is preffered.
See #1274

PR #1274, opened 2018-06-04
Modify install instructions on how to create a postgres database.

Tomasz-Kluczkowski:
Added necessary steps required to create a postgres database using the standard postgres user account.
Should be useful for users less familiar with postgres database. 
I think we could also add links on installing postgres on main platforms and performing database creation on them as the instructions in my update are for linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ static-root
 tmp/*
 media/*
 pydotorg/settings/mylocal.py
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,3 @@ static-root
 tmp/*
 media/*
 pydotorg/settings/mylocal.py
-.idea/
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -56,9 +56,27 @@ And then you'll need to install dependencies::
     $ pip install -r dev-requirements.txt
 
 *pythondotorg* will look for a PostgreSQL database named ``pythondotorg`` by
-default. Run the following command to create a new database::
+default. Run the following commands to create a new database.
 
-    $ createdb pythondotorg -E utf-8 -l en_US.UTF-8
+Change to postgres user to perform postgres database administrative tasks::
+
+    $ sudo su - postgres
+
+You should now be in a shell session for the postgres user. Log into a Postgres session by typing::
+
+    $ psql
+
+Finally create a postgres database::
+
+    postgres=# createdb pythondotorg -E utf-8 -l en_US.UTF-8
+
+Exit the SQL prompt to get back to the postgres user's shell session::
+
+    postgres=# \q
+
+Exit out of the postgres user's shell session to get back to your regular user's shell session::
+
+    $ exit
 
 To change database configuration, you can add the following setting to
 ``pydotorg/settings/local.py`` (or you can use the ``DATABASE_URL`` environment
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -86,7 +86,18 @@ variable)::
         'default': dj_database_url.parse('postgres:///your_database_name')
     }
 
-Now it's time to run migrations::
+If you prefer to use a simpler setup for your database you can use sqlite engine.
+To change database configuration to sqlite go to ``pydotorg/settings/local.py`` and:
+ - add DATABASE_URL variable
+ - modify DATABASES variable
+ Copy/paste the following snippet to replace the default settings::
+
+    DATABASE_URL = os.path.join(BASE, 'pythondotorg.db')
+    DATABASES = {
+        'default': dj_database_url.parse('sqlite:////' + DATABASE_URL)
+    }
+
+Whichever database type you chose, now it's time to run migrations::
 
     $ ./manage.py migrate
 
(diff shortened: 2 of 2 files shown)


docs/source/install.rst
```diff
@@ -56,9 +56,27 @@ And then you'll need to install dependencies::
     $ pip install -r dev-requirements.txt
 
 *pythondotorg* will look for a PostgreSQL database named ``pythondotorg`` by
-default. Run the following command to create a new database::
+default. Run the following commands to create a new database.
 
-    $ createdb pythondotorg -E utf-8 -l en_US.UTF-8
+Change to postgres user to perform postgres database administrative tasks::
+
+    $ sudo su - postgres
```
berkerpeksag: Out of curiosity, did you get an error when you ran `createdb pythondotorg -E utf-8 -l en_US.UTF-8`? If you need to be "postgres" user, perhaps passing `-O postgres` might work. Could you please try the following command?

```sh
$ createdb pythondotorg -O postgres -E utf-8 -l en_US.UTF-8
```

Tomasz-Kluczkowski: Indeed i get an error:
createdb: database creation failed: ERROR:  permission denied to create database
Even using createdb pythondotorg -O postgres -E utf-8 -l en_US.UTF-8
I have to log into posgres user to be able to create a database.
This is on Ubuntu 18.04 with the postgres 10.3 installation.

berkerpeksag: Thanks for the info. I will make another review tonight, sorry for the delay!

berkerpeksag: I think this can be done a little shorter with the following command:

```sh
$ sudo -u postgres createdb pythondotorg -E utf-8 -l en_US.UTF-8
```

Can you try this command?

docs/source/install.rst
```diff
@@ -56,9 +56,27 @@ And then you'll need to install dependencies::
     $ pip install -r dev-requirements.txt
 
 *pythondotorg* will look for a PostgreSQL database named ``pythondotorg`` by
-default. Run the following command to create a new database::
+default. Run the following commands to create a new database.
 
-    $ createdb pythondotorg -E utf-8 -l en_US.UTF-8
```
berkerpeksag: Can we keep this command as default and convert your additional commands to a note?

For example:

```rst
default. Run the following command to create a new database::

    $ createdb pythondotorg -E utf-8 -l en_US.UTF-8

.. note::

   .. You can change the wording, I just use it as an example.

   If you get an error message by running the command above::

       createdb: could not connect to database template1: FATAL:  Peer authentication failed for user "postgres"


   ...

.. Rest of the documentation here.
```

docs/source/install.rst
```diff
@@ -68,7 +86,18 @@ variable)::
         'default': dj_database_url.parse('postgres:///your_database_name')
     }
 
-Now it's time to run migrations::
+If you prefer to use a simpler setup for your database you can use sqlite engine.
+To change database configuration to sqlite go to ``pydotorg/settings/local.py`` and:
```
berkerpeksag: What do you think about converting this to mention the `DATABASE_URL` environment variable? Contributors can use SQLite without editing `pydotorg/settings/local.py`:

```sh
$ DATABASE_URL="sqlite:///pythondotorg.db" ./manage.py migrate
```

or

```sh
$ export DATABASE_URL="sqlite:///pythondotorg.db"
$ ./manage.py migrate
```

.gitignore
```diff
@@ -15,4 +15,5 @@ static-root
 /static/stylesheets/style.css
 tmp/*
 media/*
-pydotorg/settings/mylocal.py
\ No newline at end of file
+pydotorg/settings/mylocal.py
+.idea/
```
berkerpeksag: We usually prefer adding things like this to a global .gitignore. See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore for details.

Tomasz-Kluczkowski: Thanks. You never stop learning.

Tomasz-Kluczkowski:
I am going to use your suggestions and rewrite the instructions.

berkerpeksag (approved):

berkerpeksag:
Thanks! I just made some tweaks and merged.

berkerpeksag:
@Tomasz-Kluczkowski I forgot to ask whether the following command worked in your system:

```sh
$ sudo -u postgres createdb pythondotorg -E utf-8 -l en_US.UTF-8
```

Tomasz-Kluczkowski:
If I change the language to GB it works.
$ sudo -u postgres createdb pythondotorg -E utf-8 -l en_GB.UTF-8
I was able to attempt creating the database (but it failed since I already have pythondotorg database made).

berkerpeksag:
So if you change en_US.UTF-8 to en_GB.UTF-8 and pythondotorg to pythondotorg42 it works, right? Should we replace it with the steps in https://github.com/python/pythondotorg/commit/c30ebf59d5ffb781f112725955c14d1bffd8f960#diff-a2e1e3b391e0e24fcd7e517d40d9bba8R72 ?

Tomasz-Kluczkowski:
I believe we can leave it as:
$ sudo -u postgres createdb pythondotorg -E utf-8 -l en_US.UTF-8
and mention that a change of collation (language) may be necessary to match user's locale.
Shall I just amend this on my branch and push again ?

berkerpeksag:
Yeah, let's continue the discussion in a separate PR. Thanks!

Issue #1177, opened 2017-11-01
Add an app_label option to create_initial_data command


berkerpeksag:
Currently, there is no way to specify a specific application by passing its `app_label`. Example usage:

```sh
$ ./manage.py create_initial_data boxes
```

PR #1276, opened 2018-06-04
specify a specific application by passing app_label

fdroessler:
This addresses issue: #1177

### DIFF
--- a/cms/management/commands/create_initial_data.py
+++ b/cms/management/commands/create_initial_data.py
@@ -12,8 +12,8 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            '--app-label',
-            dest='app_label',
+            'app-label',
+            nargs='?',
             help='Provide an app label to create app specific data (e.g. --app_label boxes)',
             default=False,
         )
@@ -40,7 +40,7 @@ class Command(BaseCommand):
                         break
         return functions
 
-    def output(self, app_name, verbosity, done=False, result=False):
+    def output(self, app_name, verbosity, *, done=False, result=False):
         if verbosity > 0:
             if done:
                 self.stdout.write(self.style.SUCCESS('DONE'))
@@ -51,9 +51,8 @@ class Command(BaseCommand):
 
 
     def handle(self, **options):
-        # collect relevant cli arguments
         verbosity = options['verbosity']
-        app_label = options['app_label']
+        app_label = options['app-label']
         msg = (
             'Note that this command won\'t cleanup the database before '
             'creating new data.\n'
@@ -74,7 +73,7 @@ class Command(BaseCommand):
             except Exception as exc:
                 self.stdout.write(self.style.ERROR('{}: {}'.format(type(exc).__name__, exc)))
             else:
-                self.output('sitetree', verbosity, True)
+                self.output('sitetree', verbosity, done=True)
 
         for app_name, function in functions.items():
             self.output(app_name, verbosity)
@@ -84,4 +83,4 @@ class Command(BaseCommand):
                 self.stdout.write(self.style.ERROR('{}: {}'.format(type(exc).__name__, exc)))
                 continue
             else:
-                self.output(app_name, verbosity, True, result)
+                self.output(app_name, verbosity, done=True, result=result)
--- a/cms/management/commands/create_initial_data.py
+++ b/cms/management/commands/create_initial_data.py
@@ -12,8 +12,8 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            'app-label',
-            nargs='?',
+            '--app-label',
+            dest='app_label',
             help='Provide an app label to create app specific data (e.g. --app_label boxes)',
             default=False,
         )
@@ -52,7 +52,7 @@ class Command(BaseCommand):
 
     def handle(self, **options):
         verbosity = options['verbosity']
-        app_label = options['app-label']
+        app_label = options['app_label']
         msg = (
             'Note that this command won\'t cleanup the database before '
             'creating new data.\n'
--- a/cms/management/commands/create_initial_data.py
+++ b/cms/management/commands/create_initial_data.py
@@ -49,7 +49,6 @@ class Command(BaseCommand):
         if verbosity >= 2 and result:
             pprint.pprint(result)
 
-
     def handle(self, **options):
         verbosity = options['verbosity']
         app_label = options['app_label']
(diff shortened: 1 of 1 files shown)


cms/management/commands/create_initial_data.py
```diff
@@ -10,9 +10,21 @@ class Command(BaseCommand):
...
+            dest='app_label',
+            help='Provide an app label to create app specific data (e.g. --app_label boxes)',
+            default=False
+        )
+
+    def collect_initial_data_functions(self, app_label=None):
         functions = {}
-        for app in apps.get_app_configs():
+        if app_label:
+            app_list = [apps.get_app_config(app_label)]
```
berkerpeksag: Could you please catch `LookupError` and print a user friendly message if there is no app named `app_label` (for example, "foo")?

We can use `self.stdout.write(self.style.ERROR('...'))` to print the message. I'm not a native English speaker so I left that part to you :)

cms/management/commands/create_initial_data.py
```diff
@@ -10,9 +10,21 @@ class Command(BaseCommand):
 
     help = 'Create initial data by using factories.'
 
-    def collect_initial_data_functions(self):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--app_label',
```
berkerpeksag: Nit: app_label -> app-label (or we can just use what Django uses, I don't remember)

cms/management/commands/create_initial_data.py
```diff
@@ -10,9 +10,21 @@ class Command(BaseCommand):
...
+            dest='app_label',
+            help='Provide an app label to create app specific data (e.g. --app_label boxes)',
+            default=False
```
berkerpeksag: Style nit: default=False -> default=False, (add trailing comma)

cms/management/commands/create_initial_data.py
```diff
@@ -24,8 +36,9 @@ def collect_initial_data_functions(self):
                         break
         return functions
 
-    def handle(self, **options):
+    def handle(self, *args, **options):
```
berkerpeksag: Do we need `*args` here?

cms/management/commands/create_initial_data.py
```diff
@@ -43,7 +56,7 @@ def handle(self, **options):
         else:
             if verbosity > 0:
                 self.stdout.write(self.style.SUCCESS('DONE'))
-        functions = self.collect_initial_data_functions()
+        functions = self.collect_initial_data_functions(app_label)
```
berkerpeksag: I think we also need something like

```py
if not functions:
    return
```

in case when there is no app found.

berkerpeksag (commented):
Thank you, this looks great. I just have a question: Should we disable the following snippet when we passed `--app-label`?:

https://github.com/python/pythondotorg/blob/4d8d9728eea7c7b2fef32eb6f24fda409cf24a06/cms/management/commands/create_initial_data.py#L37-L45

Currently, the output of `./manage.py create_initial_data --app_label foo` is:

```
Note that this command won't cleanup the database before creating new data.
Type 'y' or 'yes' to continue, 'n' or 'no' to cancel: y
Creating initial data for 'sitetree'... DONE
... traceback ...
```

fdroessler:
Thanks for the comments, I applied them and also reformatted the code slightly. Happy to revert that if necessary.

cms/management/commands/create_initial_data.py
```diff
@@ -24,8 +40,20 @@ def collect_initial_data_functions(self):
                         break
         return functions
 
+    def output(self, app_name, verbosity, done=False, result=False):
```
berkerpeksag: Can we make boolean parameters keyword-only so we pass `self.output(..., done=True, result=False)` instead of `self.output(..., True, False)`?

cms/management/commands/create_initial_data.py
```diff
@@ -24,8 +40,20 @@ def collect_initial_data_functions(self):
...
+            if done:
+                self.stdout.write(self.style.SUCCESS('DONE'))
+            else:
+                self.stdout.write('Creating initial data for {!r}... '.format(app_name), ending='')
+        if verbosity >= 2 and result:
+            pprint.pprint(result)
+
+
     def handle(self, **options):
+        # collect relevant cli arguments
```
berkerpeksag: Style nit: IMO, the following lines are already self-descriptive. We can drop this comment.

cms/management/commands/create_initial_data.py
```diff
@@ -10,9 +10,25 @@ class Command(BaseCommand):
...
-    def collect_initial_data_functions(self):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--app-label',
+            dest='app_label',
+            help='Provide an app label to create app specific data (e.g. --app_label boxes)',
+            default=False,
+        )
+
+    def collect_initial_data_functions(self, app_label=None):
```
berkerpeksag: Can `app_label` be a positional argument now? We pass `app_label` to `collect_initial_data_functions()` in line 66 so it seems like we can make it a positional argument.

Edit: positional-only parameter is not a thing in Python yet (see PEP 570) :) I meant `def collect_initial_data_functions(self, app_label):`.

berkerpeksag (approved):
Your cleanup definitely made the code cleaner, thanks! This is almost ready to go, I just left some questions about minor details.

berkerpeksag:
@fdroessler is there any other contributor I need to give credit in the commit message?

fdroessler:
added the changes, hope I did not misunderstood the last question/comment.

berkerpeksag:
> added the changes, hope I did not misunderstood the last question/comment.

The last comment in my review? I was thinking maybe we can replace `app_label=None` with just `app_label` since we always call `collect_initial_data_functions(app_label)` so no need to define a default value (`None` in this case) for `app_label`.

fdroessler:
Ok I see, changed that. Just to make sure should `--app-label` be a optional argument used via `python manage.py create_initial_data --app-label boxes` or like I have changed it (due misunderstanding your comment) now to `python manage.py create_initial_data boxes`?

berkerpeksag:
> Just to make sure should --app-label be a optional argument used via python manage.py create_initial_data --app-label boxes or like I have changed it (due misunderstanding your comment) now to python manage.py create_initial_data boxes?

I'd prefer the first one: `create_initial_data --app-label boxes`

berkerpeksag (approved):

berkerpeksag:
Thanks!

@fdroessler would you like to submit another PR to document `--app-label` in `docs/source/install.rst`?

fdroessler:
Yes will do that

PR #1277, opened 2018-06-04
added two boolean variables for relocation and visa sponsorship

buketkonuk:
added two boolean variables for relocation and visa sponsorship

### DIFF
--- a/jobs/models.py
+++ b/jobs/models.py
@@ -139,6 +139,15 @@ class Job(ContentManageable):
     telecommuting = models.BooleanField(
         verbose_name='Telecommuting allowed?',
         default=False)
+
+    relocation = models.BooleanField(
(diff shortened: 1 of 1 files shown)


berkerpeksag (commented):
Thank you for the PR. This looks good, but we also add a new migration (see https://docs.djangoproject.com/en/2.0/topics/migrations/ for more details) and a test case.

Tests for the Job model can be found at https://github.com/python/pythondotorg/blob/master/jobs/tests/test_models.py

A simple test case for the `relocation` field would look like:

```py
def test_relocation(self):
    job = self.create_job()
    self.assertFalse(job.relocation)
    job.relocation = True
    job.save()
    self.assertTrue(job.relocation)
```

berkerpeksag:
Closing this due inactivity.

Issue #1175, opened 2017-11-01
Add a --flush option to create_initial_data command


berkerpeksag:
`./manage.py create_initial_data` currently prints the following message:

https://github.com/python/pythondotorg/blob/c3fc1f47405e93d97136757603ff77a9bf778bdd/cms/management/commands/create_initial_data.py#L29-L33

It would be great if we could add a `--flush` to delete existing records in the database.

lamar-frankie:
I'd like to claim this issue. Is there a contributor guide somewhere with instructions?

berkerpeksag:
@lamar-frankie great! Installation guide is at https://pythondotorg.readthedocs.io/install.html and a short contribution guide is at https://pythondotorg.readthedocs.io/contributing.html#code Let me know if you have any specific questions.

fdroessler:
Sorry @lamar-frankie I had been working on a related issue and just continued before seeing your comment here....

PR #1280, opened 2018-06-05
--flush option for create_initial_data

fdroessler:
This addresses #1175. 

At the moment it will display a similar message as before depending if `--flush` is provided or not. The alternative would be to use `interactive=True` in the `call_command` and let this handle if the user wants to flush the database or not. The advantage of the current implementation is that a no will abort the whole process of creating data while when only using `interactive=True` and not adding the current confirmation step, the database would not be flushed and the data generated anyway which might not be desirable. This way the user is "forced" to be explicit with the command arguments she/he is providing.

### DIFF
--- a/cms/management/commands/create_initial_data.py
+++ b/cms/management/commands/create_initial_data.py
@@ -21,7 +21,7 @@ class Command(BaseCommand):
             '--flush',
             action='store_true',
             dest='do_flush',
-            help='Remove existing data in the database before creating new data.'
+            help='Remove existing data in the database before creating new data.',
         )
 
     def collect_initial_data_functions(self, app_label):
@@ -57,7 +57,8 @@ class Command(BaseCommand):
 
     def flush_handler(self, do_flush, verbosity):
         if do_flush:
-            msg = ('You have provided the --flush parameter, this will cleanup '
+            msg = (
+                'You have provided the --flush parameter, this will cleanup '
                 'the database before creating new data.\n'
                 'Type \'y\' or \'yes\' to continue, \'n\' or \'no\' to cancel: '
                 )
@@ -66,20 +67,16 @@ class Command(BaseCommand):
                 'Note that this command won\'t cleanup the database before '
                 'creating new data.\n'
                 'If you would like to cleanup the database before creating '
-                'new data, \nplease type \'n\' or \'no\' and call the '
-                'create_initial_data command adding the --flush parameter\n'
+                'new data, \ncall create_initial_data with --flush\n'
                 'Type \'y\' or \'yes\' to continue, \'n\' or \'no\' to cancel: '
             )
         confirm = input(self.style.WARNING(msg))
-        if do_flush and (confirm in ('y', 'yes')):
+        if do_flush and confirm in ('y', 'yes'):
             try:
-                cmd = flush.Command()
-                call_command(cmd, verbosity=verbosity, interactive=False)
+                call_command('flush', verbosity=verbosity, interactive=False)
             except Exception as exc:
                 self.stdout.write(self.style.ERROR('{}: {}'.format(type(exc).__name__, exc)))
-            return confirm
-        else:
-            return confirm
+        return confirm
 
     def handle(self, **options):
         verbosity = options['verbosity']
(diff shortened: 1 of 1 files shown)


fdroessler:
@berkerpeksag would be great if you could review this one as well 😄

berkerpeksag:
@fdroessler I've just tested this in my dev environment and I think your current approach (`interactive=False`) is pretty reasonable.

I'm starting to make code review now.

cms/management/commands/create_initial_data.py
```diff
@@ -16,6 +17,12 @@ def add_arguments(self, parser):
             dest='app_label',
             help='Provide an app label to create app specific data (e.g. --app-label boxes)',
         )
+        parser.add_argument(
+            '--flush',
+            action='store_true',
+            dest='do_flush',
+            help='Remove existing data in the database before creating new data.'
```
berkerpeksag: Style nit: missing trailing comma

cms/management/commands/create_initial_data.py
```diff
@@ -48,15 +55,38 @@ def output(self, app_name, verbosity, *, done=False, result=False):
         if verbosity >= 2 and result:
             pprint.pprint(result)
 
+    def flush_handler(self, do_flush, verbosity):
+        if do_flush:
+            msg = ('You have provided the --flush parameter, this will cleanup '
```
berkerpeksag: Style nit: can we use hanging indentation?

```py
msg = (
    'line 1'
    'line 2'
)
```

cms/management/commands/create_initial_data.py
```diff
@@ -48,15 +55,38 @@ def output(self, app_name, verbosity, *, done=False, result=False):
...
+                'the database before creating new data.\n'
+                'Type \'y\' or \'yes\' to continue, \'n\' or \'no\' to cancel: '
+                )
+        else:
+            msg = (
+                'Note that this command won\'t cleanup the database before '
+                'creating new data.\n'
+                'If you would like to cleanup the database before creating '
+                'new data, \nplease type \'n\' or \'no\' and call the '
```
berkerpeksag: Is it possible to reword this without repeating the "'n' or 'no'" part? Perhaps we can just say something like "call create_initial_data with --flush."

cms/management/commands/create_initial_data.py
```diff
@@ -48,15 +55,38 @@ def output(self, app_name, verbosity, *, done=False, result=False):
...
+                'create_initial_data command adding the --flush parameter\n'
+                'Type \'y\' or \'yes\' to continue, \'n\' or \'no\' to cancel: '
+            )
+        confirm = input(self.style.WARNING(msg))
+        if do_flush and (confirm in ('y', 'yes')):
```
berkerpeksag: Style nit: (confirm in ('y', 'yes')) -> confirm in ('y', 'yes') (parentheses can be removed)

cms/management/commands/create_initial_data.py
```diff
@@ -48,15 +55,38 @@ def output(self, app_name, verbosity, *, done=False, result=False):
...
+            try:
+                cmd = flush.Command()
+                call_command(cmd, verbosity=verbosity, interactive=False)
```
berkerpeksag: We don't need to import `flash.Command()` here. Passing `'flush'` as first argument should do the trick.

cms/management/commands/create_initial_data.py
```diff
@@ -48,15 +55,38 @@ def output(self, app_name, verbosity, *, done=False, result=False):
...
+            except Exception as exc:
+                self.stdout.write(self.style.ERROR('{}: {}'.format(type(exc).__name__, exc)))
+            return confirm
```
berkerpeksag: It looks like we can simply this part a bit:

```py
            return confirm
        else:
            return confirm
```

What do you think about the following snippet?

```py
confirm = input(self.style.WARNING(msg))
if do_flush and confirm in ('y', 'yes'):
    ...
return confirm

fdroessler: yeah this seemed a bit weird when I wrote it.

berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1282, opened 2018-06-07
Fix for the distinct error when running website over sqlite

datapythonista:
This code should be equivalent to the existing, but compatible with all backends.

Had some problems testing the change, I rely on the CI.

### DIFF
--- a/jobs/views.py
+++ b/jobs/views.py
@@ -47,8 +47,9 @@ class JobMixin:
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        active_locations = Job.objects.visible().distinct(
-            'location_slug'
+        active_locations = Job.objects.visible(
+        ).values_list('location_slug', flat=True
+        ).distinct(
         ).order_by(
             'location_slug',
         )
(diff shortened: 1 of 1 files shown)


jobs/views.py
```diff
@@ -47,8 +47,9 @@ class JobMixin:
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        active_locations = Job.objects.visible().distinct(
-            'location_slug'
+        active_locations = Job.objects.visible(
+        ).values_list('location_slug', flat=True
```
berkerpeksag, line 51: Unfortunately, this will return

```py
<JobQuerySet ['lawrence-ks-us']>
```

instead of

```py
<JobQuerySet [<Job: Job Listing #6>]>
```

The output of `active_locations` will be used in `templates/jobs.base.html` so we also need to get `city`, `region`, and `country` fields.

Perhaps we change this line to

```py
values_list('location_slug', 'city', 'region', 'country')
```

?

berkerpeksag:
By the way, I don't see this PR on Travis. I'm going to close and reopen and see if it's going to be picked up by Travis.

jobs/views.py
datapythonista: You're right, I misunderstood that.

What you propose makes sense, but as the data is not normalized, we'll have duplicate `location_slug` values if they have different `city`, `country`...

I think the only other option is to fetch all data from the database, and remove duplicate `location_slug` in Python.

Closing this PR.

PR #1283, opened 2018-06-08
Fix the bug that the `dump_pep_pages` command was broken

mozillazg:


peps/tests/test_commands.py
```diff
@@ -33,3 +38,20 @@ def test_image_generated(self):
         img = Image.objects.get(page__path='dev/peps/pep-3001/')
         soup = BeautifulSoup(img.page.content.raw, 'lxml')
         self.assertIn(settings.MEDIA_URL, soup.find('img')['src'])
+
+    @override_settings(PEP_REPO_PATH=FAKE_PEP_REPO)
+    def test_dump_pep_pages_work(self):
+        call_command('generate_pep_pages')
+        stdout = io.StringIO()
+        sys.stdout = stdout
```
berkerpeksag: We can replace monkeypatching with the following snippet:

```py
stdout = io.StringIO()
call_command('dump_pep_pages', stdout=stdout)
output = stdout.getvalue()
```

peps/management/commands/dump_pep_pages.py
```diff
@@ -4,14 +4,14 @@
 from pages.models import Page
 
 
-class Command(BaeeCommand):
+class Command(BaseCommand):
```
berkerpeksag, line 7: Nice catch!

peps/tests/test_commands.py
```diff
@@ -33,3 +38,20 @@ def test_image_generated(self):
...
+        try:
+            call_command('dump_pep_pages')
+        finally:
+            sys.stdout = sys.__stdout__
+        stdout.seek(0)
+        output = stdout.read()
+        stdout.close()
+
+        json.loads(output)
```
berkerpeksag: Do we need to call `json.loads()` here?

peps/tests/test_commands.py
```diff
@@ -33,3 +38,20 @@ def test_image_generated(self):
         img = Image.objects.get(page__path='dev/peps/pep-3001/')
         soup = BeautifulSoup(img.page.content.raw, 'lxml')
         self.assertIn(settings.MEDIA_URL, soup.find('img')['src'])
+
+    @override_settings(PEP_REPO_PATH=FAKE_PEP_REPO)
+    def test_dump_pep_pages_work(self):
```
berkerpeksag: `_work` can be removed from the test name.

peps/tests/test_commands.py
mozillazg: No, it's not necessary.

peps/tests/test_commands.py
mozillazg: 👍

berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1287, opened 2018-06-13
Simplify postgres database creating instructions.

Tomasz-Kluczkowski:
Simplified instructions for the contributors when they want to create a
new database and must use postgres user account as the owner.
Now it is a one liner.
Also added suggestion of what to do if the languages do not match.


berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1292, opened 2018-06-27
must use from import

bushuev-a:


berkerpeksag (commented):
Thanks for the PR. Can you explain why we need to use the `from ... import ...` form?

berkerpeksag:
I'm closing this one for now. @kolayv we can discuss this further if you reply my earlier question.

PR #1295, opened 2018-07-12
Update readme, clarify that this is not CPython's issue tracker

Mariatta:
We have people mistakenly reporting issues with CPython here.
Figured it will be good to explain that this is just for python.org, not CPython.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,4 +21,3 @@ https://bugs.python.org.
 * IRC: `#pydotorg` on Freenode
 * Staging site: https://staging.python.org/ (`master` branch)
 * License: Apache License
-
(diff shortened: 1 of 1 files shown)


berkerpeksag (commented):
Perhaps it would be better if we create an issue template? I doubt people would read the README before filing an issue here :)

Mariatta:
Issue template sounds good too. Ok I'll try to come up with one.

Mariatta:
Issue templates created in https://github.com/python/pythondotorg/pull/1296

I think having that info in readme is useful too though.

berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1305, opened 2018-08-01
Simplify ReleaseManager

jaap3:
Defer most manager methods to the `ReleaseQueryset` by using `Manager.from_queryset(ReleaseQuerySet)` as the base for `ReleaseManager`. 

Keep only the special cases `latest_python2` and `latest_python3`


berkerpeksag (approved):
Nice cleanup, thanks!

PR #1306, opened 2018-08-01
Add TestCase as base to DownloadModelTests

jaap3:
The DownloadModelTests weren't actually running because they didn't inherit from TestCase

Before: `Ran 246 tests` 
After: `Ran 257 tests`

### DIFF
--- a/downloads/tests/base.py
+++ b/downloads/tests/base.py
@@ -25,7 +25,7 @@ class DownloadMixin:
         cls.linux, _ = OS.objects.get_or_create(name='Linux')
 
 
-class BaseDownloadTests(DownloadMixin):
+class BaseDownloadTests(DownloadMixin, TestCase):
 
     def setUp(self):
         self.release_275_page = Page.objects.create(
--- a/downloads/tests/test_views.py
+++ b/downloads/tests/test_views.py
@@ -28,7 +28,7 @@ TEST_THROTTLE_RATES = {
 }
 
 
-class DownloadViewsTests(BaseDownloadTests, TestCase):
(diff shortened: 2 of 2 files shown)


downloads/tests/test_models.py
```diff
@@ -1,8 +1,10 @@
+from django.test import TestCase
+
 from ..models import Release
 from .base import BaseDownloadTests
 
 
-class DownloadModelTests(BaseDownloadTests):
+class DownloadModelTests(BaseDownloadTests, TestCase):
```
berkerpeksag: I'd rather add `TestCase` to `BaseDownloadTests` in `downloads/tests/base.py`:

```py
class BaseDownloadTests(DownloadMixin, TestCase):
    ...
```

This way we don't have to remember adding both `BaseDownloadTests` and `TestCase` classes as base classes everytime we create a new test case.

berkerpeksag:
We can drop `TestCase` from the following class as well:

```py
class DownloadApiV1ViewsTest(BaseDownloadApiViewsTest, BaseDownloadTests, TestCase):
    ...
```

jaap3:
With this change `DownloadApiV2ViewsTest` will inherit from `django.test.TestCase` twice, once through `BaseDownloadTests` and once through `rest_framework.test.APITestCase`.

Not sure if that's OK (seems to work).

berkerpeksag (approved):

berkerpeksag:
Yes, I think Python's MRO algorithm is smart enough to handle that case. If not, we can do a bit refactoring make it work.

berkerpeksag:
Thanks.

Issue #1289, opened 2018-06-21
Python org shell window console iframe height not enough for 1366x768


mohit9896807610:
Tried using the shell for the first time and noticed the height was not enough when you've typed enough command to fill up the iframe, the problem is , then the output or even the next line content is not properly visible.

On inspecting in chrome (v67 ) found it to be 280px, which was just a bit shorter than adequate. Changed it to 300px and works great, can really suggest this to be changed to 300px or any other appropriate value.

PR #1312, opened 2018-08-22
Shell slightly under-sized

gpotter2:
The shell is slightly under-sized, meaning that the shell is missing the last 1-2 lines

https://www.python.org/shell/

image

A few more pixels and we're ok

image

### DIFF
--- a/templates/python/shell.html
+++ b/templates/python/shell.html
@@ -5,7 +5,7 @@
 {% block header_content %}
         <div class="slideshow">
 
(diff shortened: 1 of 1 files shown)


berkerpeksag (approved):

berkerpeksag:
Thanks! This was reported in #1289, so I've just closed it as well.

gpotter2:
Nice ! Thanks

PR #1313, opened 2018-08-27
Make sure create_initial_data is executed in the same order each time

jaap3:
This could prevent random success/failure caused by the order of the function dict (I'm unsure if there are fixtures that depend on other fixtures to run before them).

This change is not necessary for Python < 3.7 (technically 3.6) as dicts will keep their insertion order.

### DIFF
--- a/cms/management/commands/create_initial_data.py
+++ b/cms/management/commands/create_initial_data.py
@@ -1,3 +1,4 @@
+import collections
 import importlib
 import inspect
 import pprint
@@ -24,7 +25,7 @@ class Command(BaseCommand):
         )
 
     def collect_initial_data_functions(self, app_label):
-        functions = {}
+        functions = collections.OrderedDict()
         if app_label:
             try:
                 app_list = [apps.get_app_config(app_label)]


berkerpeksag (commented):
I've been using `create_initial_data` without any problems with Python 3.4 and Python 3.5 since I wrote it in last year. I'd like keep the current code as-is unless you've noticed a bug.

jaap3:
I just noticed that things were being created in a different order each time. This was causing issues  with the downloads factories I've created since they rely on certain Boxes to be available. This would only happen when the downloads factories were executed before the boxes, so it was somewhat of a heisenbug.

I've also created #1315 to work around that issue so that one might be preferable over this one.

berkerpeksag:
Thanks for the explanation. Closing this in favor of PR #1315 then.

Note that the use of `OrderedDict` wouldn't %100 solve this problem either. I think the proper way to solve the dependency problem would be to design a proper dependency mechanism.

jaap3:
I agree with closing this PR. I'm aware that fixing the order wouldn't really solve the dependency issue, it would just make fixtures reliably fail/work if they are interdependent. 

I'd also argue against designing a dependency mechanism, and in favor of not introducing fixtures that depend on other fixtures ;-).

PR #1314, opened 2018-08-27
Use timezone.now as default for updated field

jaap3:
This makes it possible to `bulk_create` `ContentManageable` and `Membership` objects.

Since `bulk_create` skips `save` it is not possible to `bulk_create` objects that inherit from `ContentManageable` because the `updated` field doesn't get a value. By adding a default it becomes possible to `bulk_create`. The functionality remains the same, `updated` will always be set to the current time.

### DIFF
--- a/blogs/migrations/0001_initial.py
+++ b/blogs/migrations/0001_initial.py
@@ -34,7 +34,7 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.AutoField(primary_key=True, auto_created=True, serialize=False, verbose_name='ID')),
                 ('created', models.DateTimeField(db_index=True, default=django.utils.timezone.now, blank=True)),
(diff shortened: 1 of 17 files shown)


berkerpeksag (approved):
This looks good to me. I'll try it out locally and merge this week. Thanks!

berkerpeksag:
@jaap3 by the way, did you or could you try to run migrations on a fresh database?

jaap3:
Good question, I don't think I have. I'll provision a fresh Vagrant box
tomorrow and will let you know if it works.

jaap3:
OK, just rebuilt the vagrant box (making sure I was using the correct branch) and ran `create_initial_data`. Everything seems to work just fine. Come to think of it, Travis runs the migrations before running the tests as well, so these are covered by a normal test run.

berkerpeksag:
Thanks!

PR #1315, opened 2018-08-27
Use Box.objects.get_or_create

jaap3:
This makes it possible to create a release without installing the Box fixtures.

This should be perfectly safe as the box content is set immediately after getting/creating the box.

### DIFF
--- a/blogs/parser.py
+++ b/blogs/parser.py
@@ -40,6 +40,10 @@ def update_blog_supernav():
     latest_entry = BlogEntry.objects.filter(feed_id=1).latest()
     rendered_box = _render_blog_supernav(latest_entry)
 
-    box = Box.objects.get(label='supernav-python-blog')
-    box.content = rendered_box
-    box.save()
+    box, _ = Box.objects.update_or_create(
+        label='supernav-python-blog',
+        defaults={
+            'content': rendered_box,
+            'content_markup_type': 'html',
+        }
+    )
--- a/downloads/models.py
+++ b/downloads/models.py
@@ -158,9 +158,13 @@ def update_supernav():
         'python_files': python_files,
     })
 
-    box = Box.objects.get(label='supernav-python-downloads')
-    box.content = content
-    box.save()
+    box, _ = Box.objects.update_or_create(
+        label='supernav-python-downloads',
+        defaults={
+            'content': content,
+            'content_markup_type': 'html',
+        }
+    )
 
     # Update latest Sources box on Download landing page
     if latest_python2:
@@ -173,13 +177,17 @@ def update_supernav():
     else:
         latest_python3_source = None
 
-    source_box = Box.objects.get(label='download-sources')
     source_content = render_to_string('downloads/download-sources-box.html', {
         'latest_python2_source': latest_python2_source,
         'latest_python3_source': latest_python3_source,
     })
-    source_box.content = source_content
-    source_box.save()
+    source_box, _ = Box.objects.update_or_create(
+        label='download-sources',
+        defaults={
+            'content': source_content,
+            'content_markup_type': 'html',
+        }
+    )
 
 
 def update_homepage_download_box():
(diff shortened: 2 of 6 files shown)


jaap3:
Looking at this again I think this could even be changed to `update_or_create`. Let me know if that's preferred.

berkerpeksag (commented):
You're right, changing the code to use `update_or_create()` would be better.

jaap3:
I've updated the PR to use `update_or_create`, I've also changed the use of `Box.objects.get` / `Box.objects.get_or_create` in other places

blogs/parser.py
```diff
@@ -40,6 +40,10 @@ def update_blog_supernav():
...
     rendered_box = _render_blog_supernav(latest_entry)
 
-    box = Box.objects.get(label='supernav-python-blog')
-    box.content = rendered_box
-    box.save()
+    box, _ = Box.objects.update_or_create(
+        label='supernav-python-blog',
+        defaults={
+            'content': rendered_box,
+            'content_markup_type': 'html'
```
berkerpeksag: Nit: Please add a trailing comma.

downloads/models.py
```diff
@@ -158,9 +158,13 @@ def update_supernav():
...
     })
 
-    box = Box.objects.get(label='supernav-python-downloads')
-    box.content = content
-    box.save()
+    box, _ = Box.objects.update_or_create(
+        label='supernav-python-downloads',
+        defaults={
+            'content': content,
+            'content_markup_type': 'html'
```
berkerpeksag: Nit: Please add a trailing comma.

downloads/models.py
```diff
@@ -173,13 +177,17 @@ def update_supernav():
...
         'latest_python2_source': latest_python2_source,
         'latest_python3_source': latest_python3_source,
     })
-    source_box.content = source_content
-    source_box.save()
+    source_box, _ = Box.objects.update_or_create(
+        label='download-sources',
+        defaults={
+            'content': source_content,
+            'content_markup_type': 'html'
```
berkerpeksag: Nit: Please add a trailing comma.

downloads/models.py
```diff
@@ -198,9 +206,13 @@ def update_homepage_download_box():
...
     })
 
-    box = Box.objects.get(label='homepage-downloads')
-    box.content = content
-    box.save()
+    box, _ = Box.objects.update_or_create(
+        label='homepage-downloads',
+        defaults={
+            'content': content,
+            'content_markup_type': 'html'
```
berkerpeksag: Nit: Please add a trailing comma.

successstories/models.py
```diff
@@ -100,9 +100,13 @@ def update_successstories_supernav(sender, instance, created, **kwargs):
...
         })
 
-        box, _ = Box.objects.get_or_create(label='supernav-python-success-stories')
-        box.content = content
-        box.save()
+        box, _ = Box.objects.update_or_create(
+            label='supernav-python-success-stories',
+            defaults={
+                'content': content,
+                'content_markup_type': 'html'
```
berkerpeksag: Nit: Please add a trailing comma.

berkerpeksag (commented):
This looks like a good cleanup, thanks! I only have a comment about using trailing commas after the last item of the dict.

jaap3:
I've added the trailing commas. 

On a related note: It might be interesting to run [black](https://black.readthedocs.io/en/stable/) on this codebase at some point. That would fix a lot of existing code style inconsistencies.

berkerpeksag (approved):

berkerpeksag:
Thanks!

> On a related note: It might be interesting to run black on this codebase at some point. That would fix a lot of existing code style inconsistencies.

Sorry, I dislike some of the style choices of black :) Also, we don't accept cosmetic-only patches in CPython, so I'd like to follow that rule in python.org as well.

PR #1316, opened 2018-08-27
Ignore venv

jaap3:
The `venv` directory created during Vagrant provisioning was not ignored. I also decided to cleanup/sort the .gitignore file a bit.


berkerpeksag (changes requested):
Please use a global .gitignore file for this. `venv` is a pretty common name for virtual environments and I think it's better adding it into a global .gitignore, not a project-specific one. I'd prefer cleaning up `.gitignore` instead.

jaap3:
Do you want me to submit a PR that removes other commonly ignored things like `__pycache__`, `*.pyc` etc?

berkerpeksag:
Yeah, that would be nice if you have time to prepare a PR. I think the first nine entries can be removed. Use of coveragepy is not that common, perhaps we can keep `.coverage` and `htmlcov`.

`tmp/*` and `pydotorg/settings/mylocal.py` can be removed as well.

Issue #956, opened 2016-06-30
Redirect /download/releases/X.Y.Z to /downloads/release/python-XYZ


berkerpeksag:
Currently, `/download/releases/X.Y.Z` links are stored in the `pages.Page` model but we already have same feature in the `downloads` application. It would be nice to
1. merge their content
2. redirect `/download/releases/X.Y.Z` to `/downloads/releases/python-XYZ`

for all 2.7 and 3.x releases. We should keep the old URL active for older releases like 1.6, 2.3, etc.

**Pros**
- No need to duplicate release notes anymore
- One less step for release managers

ned-deily:
+1. Actually, we're no longer consistently creating the duplicated page in `/download/releases`, for example, for 2.7.11.  I, for one, had forgotten about the old pattern of links. I'm not sure how important it is to maintain them forever but if the redirects could be generated automatically that would be great.  ( @benjaminp @larryhastings )

PR #1317, opened 2018-08-27
Add downloads factories

jaap3:
Recreate the downloads section by importing data from the python.org API

### DIFF
--- a/downloads/factories.py
+++ b/downloads/factories.py
@@ -43,7 +43,9 @@ class APISession(requests.Session):
 
     def request(self, method, url, **kwargs):
         url = urljoin(self.base_url, url)
-        return super().request(method, url, **kwargs)
+        response = super().request(method, url, **kwargs)
+        response.raise_for_status()
+        return response
 
 
 def _get_id(obj, key):
--- a/downloads/factories.py
+++ b/downloads/factories.py
@@ -1,0 +1,111 @@
+from urllib.parse import urljoin
+
+import factory
+import requests
+
+from users.factories import UserFactory
+
+from .models import OS, Release, ReleaseFile
+
+
+class OSFactory(factory.DjangoModelFactory):
+
+    class Meta:
+        model = OS
+        django_get_or_create = ('slug',)
+
+    creator = factory.SubFactory(UserFactory)
+
+
+class ReleaseFactory(factory.DjangoModelFactory):
+
+    class Meta:
+        model = Release
+        django_get_or_create = ('slug',)
+
+    creator = factory.SubFactory(UserFactory)
+    is_published = True
+
+
+class ReleaseFileFactory(factory.DjangoModelFactory):
+
+    class Meta:
+        model = ReleaseFile
+        django_get_or_create = ('slug',)
+
+    creator = factory.SubFactory(UserFactory)
+    release = factory.SubFactory(ReleaseFactory)
+    os = factory.SubFactory(OSFactory)
+
+
+class APISession(requests.Session):
+    base_url = 'https://www.python.org/api/v1/'
+
+    def request(self, method, url, **kwargs):
+        url = urljoin(self.base_url, url)
+        return super().request(method, url, **kwargs)
+
+
+def _get_id(obj, key):
+    """
+    Get the ID of an object by extracting it from the resource uri.
+    """
+    return (obj.pop(key, '') or '').rstrip('/').rpartition('/')[-1]
+
+
+def initial_data():
+    """
+    Create the data for the downloads section by importing
+    it from the python.org API.
+    """
+    objects = {
+        'oss': {},
+        'releases': {},
+        'release_files': {},
+    }
+
+    with APISession() as session:
+        for key, resource_uri in [
+            ('oss', 'downloads/os/'),
+            ('releases', 'downloads/release/'),
+            ('release_files', 'downloads/release_file/')
+        ]:
+            while resource_uri:
+                response = session.get(resource_uri, params={'limit': 1000}).json()
+
+                for obj in response['objects']:
+                    objects[key][_get_id(obj, 'resource_uri')] = obj
+
+                resource_uri = response['meta']['next']
+
+    # Create the list of operating systems
+    objects['oss'] = {k: OSFactory.build(**obj) for k, obj in objects['oss'].items()}
+    OS.objects.bulk_create(objects['oss'].values())
+
+    # Create all the releases
+    for key, obj in objects['releases'].items():
+        obj.pop('release_page')  # Ignore release pages
+        objects['releases'][key] = ReleaseFactory.build(**obj)
+    Release.objects.bulk_create(objects['releases'].values())
+
+    # Create all release files
+    for key, obj in tuple(objects['release_files'].items()):
+        release_id = _get_id(obj, 'release')
+        try:
+            release = objects['releases'][release_id]
+        except KeyError:
+            # Release files for draft releases are available through the API,
+            # the releases are not.
+            # https://github.com/python/pythondotorg/issues/1308
+            objects['release_files'].pop(key)
+        else:
+            obj['release'] = release
+            obj['os'] = objects['oss'][_get_id(obj, 'os')]
+            objects['release_files'][key] = ReleaseFileFactory.build(**obj)
+    ReleaseFile.objects.bulk_create(objects['release_files'].values())
+
+    return {
+        'oss': list(objects.pop('oss').values()),
+        'releases': list(objects.pop('releases').values()),
+        'release_files': list(objects.pop('release_files').values()),
+    }
--- a/downloads/factories.py
+++ b/downloads/factories.py
@@ -76,14 +76,14 @@ def initial_data():
     it from the python.org API.
     """
     objects = {
-        'oss': {},
+        'oses': {},
         'releases': {},
         'release_files': {},
     }
 
     with APISession() as session:
         for key, resource_uri in [
-            ('oss', 'downloads/os/'),
+            ('oses', 'downloads/os/'),
             ('releases', 'downloads/release/'),
             ('release_files', 'downloads/release_file/')
         ]:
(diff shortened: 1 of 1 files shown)


jaap3:
This depends on the changes in #1314 and #1315

downloads/factories.py
```diff
@@ -0,0 +1,112 @@
...
+        model = ReleaseFile
+        django_get_or_create = ('slug',)
+
+    creator = factory.SubFactory(UserFactory)
+    release = factory.SubFactory(ReleaseFactory)
+    os = factory.SubFactory(OSFactory)
+
+
+class APISession(requests.Session):
+    base_url = 'https://www.python.org/api/v1/'
```
berkerpeksag: The new API will be deployed to production on this Friday. I would like to use v2 here. If you have some time before Friday, you can try it out at https://staging.python.org/api/v2/

jaap3:
Tested this with the new API and made some corrections to support it. While I was at it I also fixed some other issues.

downloads/factories.py
```diff
@@ -0,0 +1,120 @@
...
+        response = super().request(method, url, **kwargs)
+        response.raise_for_status()
+        return response
+
+
+def _get_id(obj, key):
+    """
+    Get the ID of an object by extracting it from the resource uri.
+    """
+    return (obj.pop(key, '') or '').rstrip('/').rpartition('/')[-1]
```
berkerpeksag: Do we need the `or ''` part?

downloads/factories.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def initial_data():
+    """
+    Create the data for the downloads section by importing
+    it from the python.org API.
+
+    """
+    objects = {
+        'oss': {},
```
berkerpeksag: Typo? os -> oses

downloads/factories.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class ReleaseFactory(factory.DjangoModelFactory):
+
+    class Meta:
+        model = Release
+        django_get_or_create = ('slug',)
+
+    creator = factory.SubFactory(UserFactory)
+    is_published = True
```
berkerpeksag, line 27: Could you add `is_latest` as well? We need it to make `update_supernav()` work. https://github.com/python/pythondotorg/commit/5ad3ebf26ff09848557d85587afdf223d57ce20c should expose it in both APIs.

downloads/factories.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class APISession(requests.Session):
+    base_url = 'https://www.python.org/api/v2/'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.headers.update(
+            {
+                "Accept": "application/json",
```
berkerpeksag: Style nit: Please use single quotes everywhere.

downloads/factories.py
```diff
@@ -0,0 +1,120 @@
...
+    Get the ID of an object by extracting it from the resource uri.
+    """
+    return (obj.pop(key, '') or '').rstrip('/').rpartition('/')[-1]
+
+
+def initial_data():
+    """
+    Create the data for the downloads section by importing
+    it from the python.org API.
+
```
berkerpeksag: Style nit: Newline can be dropped.

downloads/factories.py
```diff
@@ -0,0 +1,120 @@
...
+            for obj in object_list:
+                objects[key][_get_id(obj, 'resource_uri')] = obj
+
+    # Create the list of operating systems
+    objects['oss'] = {k: OSFactory(**obj) for k, obj in objects['oss'].items()}
+
+    # Create all the releases
+    for key, obj in objects['releases'].items():
+        obj.pop('release_page')  # Ignore release pages
```
berkerpeksag: Did you skip adding release pages for now or did you get blocked by something? I think we can add a `# TODO` comment if it takes too long to implement it.

berkerpeksag (commented):
This is great! I needed this when I was working on #1370 and I noticed that I didn't expose the `is_latest` field.

berkerpeksag:
I think there's a bug here. After I ran `create_initial_data` the supernav for Downloads menu didn't change:

Screen Shot 2019-04-07 at 6 30 16 AM

I created a temporary `update_boxes` command:

```py
from django.core.management import BaseCommand, call_command


class Command(BaseCommand):

    def handle(self, **options):
        from downloads.models import update_supernav
        update_supernav()
```

After I ran it:

Screen Shot 2019-04-07 at 6 30 56 AM

I also applied this patch:

```diff
diff --git a/downloads/models.py b/downloads/models.py
index 305143b..72f40e4 100644
--- a/downloads/models.py
+++ b/downloads/models.py
@@ -126,35 +126,34 @@ class Release(ContentManageable, NameSlugModel):
 
 
 def update_supernav():
-    try:
-        latest_python2 = Release.objects.latest_python2()
-    except Release.DoesNotExist:
-        latest_python2 = None
-
     try:
         latest_python3 = Release.objects.latest_python3()
     except Release.DoesNotExist:
-        latest_python3 = None
+        # If we don't have a Python 3 release, we don't need to bother updating supernav.
+        return
 
     python_files = []
     for o in OS.objects.all():
         data = {
             'os': o,
-            'python2': None,
             'python3': None,
         }
 
-        if latest_python2:
-            data['python2'] = latest_python2.download_file_for_os(o.slug)
-
         if latest_python3:
-            data['python3'] = latest_python3.download_file_for_os(o.slug)
+            release_file = latest_python3.download_file_for_os(o.slug)
+            if not release_file:
+                return
+            data['python3'] = release_file
 
         python_files.append(data)
 
+    if not python_files:
+        return
+
+    if not all(f['python3'] for f in python_files):
+        return
+
     content = render_to_string('downloads/supernav.html', {
-        'latest_python2': latest_python2,
-        'latest_python3': latest_python3,
         'python_files': python_files,
     })
 
@@ -166,7 +165,20 @@ def update_supernav():
         }
     )
 
+
+def update_download_sources_box():
     # Update latest Sources box on Download landing page
+
+    try:
+        latest_python2 = Release.objects.latest_python2()
+    except Release.DoesNotExist:
+        latest_python2 = None
+
+    try:
+        latest_python3 = Release.objects.latest_python3()
+    except Release.DoesNotExist:
+        latest_python3 = None
+
     if latest_python2:
         latest_python2_source = latest_python2.download_file_for_os('source')
     else:
@@ -177,6 +189,9 @@ def update_supernav():
     else:
         latest_python3_source = None
 
+    if not latest_python2_source or not latest_python3_source:
+        return
+
     source_content = render_to_string('downloads/download-sources-box.html', {
         'latest_python2_source': latest_python2_source,
         'latest_python3_source': latest_python3_source,
@@ -260,14 +275,14 @@ def purge_fastly_download_pages(sender, instance, **kwargs):
 
 
 @receiver(post_save, sender=Release)
-def update_download_supernav(sender, instance, **kwargs):
-    """ Update download supernav """
+def update_download_supernav_and_boxes(sender, instance, **kwargs):
     # Skip in fixtures
     if kwargs.get('raw', False):
         return
 
     if instance.is_published:
         update_supernav()
+        update_download_sources_box()
         update_homepage_download_box()
 ```

**Comment by berkerpeksag** [comment]
I think I figured it out. Due to how `initial_data()` functions and Django signals work, `release.files.all()` will always return no records when a `Release` instance is saved. This can be solved by registering a `post_save` callback for `ReleaseFile`. However, I think it's unnecessary and making our signal handlers more robust would be a better way to solve this, so supernav boxes won't be updated unnecessarily (as shown below, without using my inline patch, against current master with this PR applied, after I ran `create_initial_data`):

Screen Shot 2019-04-07 at 2 45 54 PM

I will work on my inline patch to make signal handlers more robust.

File: downloads/factories.py (hunk 7 above)
Comment (jaap3) [reply downloads/factories.py]: I think I got blocked on something, but I cannot remember what. I'll add a `TODO` for now

Comment (berkerpeksag) [reply downloads/factories.py]: Great! Feel free to file an issue if you remember what was the blocker.

File: downloads/factories.py (hunk 4 above)
Comment (jaap3) [reply downloads/factories.py:27]: There's no need to add `is_latest` to the factory, if it's on the model then it's also on the factory. So in this case `is_latest` will default to `False`. In `initial_data` the release is created like this `ReleaseFactory(**obj)`, so all fields that are available in the API are set on the newly created instance. So this should already set `is_latest` correctly.

File: downloads/factories.py (hunk 2 above)
Comment (jaap3) [reply downloads/factories.py]: I've changed the implementation slightly, should be clearer now

File: downloads/factories.py (hunk 3 above)
Comment (jaap3) [reply downloads/factories.py]: yeah, I didn't know how to pluralize OS :)

File: downloads/factories.py (hunk 4 above)
Comment (berkerpeksag) [reply downloads/factories.py:27]: Ah, yes, I missed you've added `is_published = True` because its default value is `False`. Good point!

**Review by berkerpeksag — APPROVED** [review]
LGTM! I already tested this last week and will merge once Travis is happy.

**Comment by jaap3** [comment]
Thanks for merging, glad you find it useful :)

I just did a run where I added a `print(obj['release_page'])` (not sure why I did not do that saturday) and noticed that the field never has a value (always `None`). I'm fairly sure it used to have a value (maybe it was API V1?).

**Comment by berkerpeksag** [comment]
> I'm fairly sure it used to have a value (maybe it was API V1?).

It could be a bug I introduced while I was porting API v1 to DRF. Could you check https://github.com/python/pythondotorg/blob/master/downloads/tests/test_views.py whether there's a test for the `release_page` field?

**Comment by jaap3** [comment]
There's `test_post_release` that asserts `self.assertIn(data['release_page'], content['release_page'])`.

However, I just did a run against API V1 and the `release_page` field is always `None` there as well. So I'm not sure what's going on.

To use API v1 replace these bits:

```
class APISession(requests.Session):
    base_url = 'https://www.python.org/api/v1/'
```

and

```
def initial_data():
    [...]
    with APISession() as session:
        for key, resource_uri in [
            ('oses', 'downloads/os/'),
            ('releases', 'downloads/release/'),
            ('release_files', 'downloads/release_file/'),
        ]:
            while resource_uri:
                response = session.get(resource_uri, params={'limit': 1000}).json()
                for obj in response['objects']:
                    objects[key][_get_id(obj, 'resource_uri')] = obj
                resource_uri = response['meta']['next']
```

berkerpeksag:
We don't use the release_page field much these days. You may need to create a release with a release page locally to verify it.

jaap3:
Yeah, seems like release_page is a legacy field, the fix to #956 actually removes a lot of release pages.

PR #1319, opened 2018-08-28
Cleanup .gitignore

jaap3:
Recommend using a global git ignore file to ignore common files.


jaap3:
As you can see, I've removed everything that isn't project specific and added a link to an example global gitignore file that would cover all the things that were in the gitignore file (including coverage files, vagrant and sass).

.gitignore
```diff
@@ -1,18 +1,16 @@
...
+#
+# https://www.gitignore.io/api/windows,linux,osx,python,vagrant,sass,idea,visualstudiocode
+#
+# Store the content of this file in your home directory as ~/.gitignore_global
+# and configure git using the following command:
+#
+# $ git config --global core.excludesfile ~/.gitignore_global
 
-static-root
-.sass-cache
```
berkerpeksag, line 12: `.sass-cache` is an implementation detail of the Ruby implementation of sass. I'd guess many Python projects don't depend on the Ruby gem (personally python.org is the only one I've worked on) and use libsass-python instead. It would be convenient if we keep `.sass-cache` for now.

.gitignore
```diff
@@ -1,18 +1,16 @@
-*.log
-*.pot
-*.pyc
-__pycache__
-.vagrant
-build
-ENV
-.coverage
-htmlcov
+# Consider using a global git ignore file for files and directories that
```
berkerpeksag: Nit: "git ignore" -> .gitignore

.gitignore
```diff
@@ -1,18 +1,16 @@
...
+# are commonly generated by your OS, IDE, Python, development tools etc.
+# A fairly extensive starting point would be:
+#
+# https://www.gitignore.io/api/windows,linux,osx,python,vagrant,sass,idea,visualstudiocode
```
berkerpeksag: I didn't know this service, awesome!

.gitignore
jaap3: It's included in the list that's linked in the comment (part of the sass set). It's fairly common in older projects that use compass.

I'll add it back and remove the sass set from the global gitignore example link.

.gitignore
jaap3: Me neither before today, I found it linked from the GitHub docs (https://help.github.com/articles/ignoring-files/).

berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1334, opened 2018-09-18
Upgrade vagrant box to run on python 3.6

### LINKED ISSUE #1333: Upgrade Vagrant box to run on Python 3.6
We probably need to use a newer Ubuntu box in https://github.com/python/pythondotorg/blob/master/provisioning/pythondotorg.yml to install Python 3.6 from the official Ubuntu repositories.

Also, https://pythondotorg.readthedocs.io/install.html#easy-setup-using-vagrant documentation needs to be updated.

Tomasz-Kluczkowski:
Changed vagrant box to bento/ubuntu-18.04 which gives Python 3.6.5.

Fixes #1333

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -18,7 +18,7 @@ You should then be able to provision the Vagrant box.
 
     $ vagrant up
 
-The box will be provisioned with Python 3.5, a virtual environment with all
+The box will be provisioned with Python 3.6.5, a virtual environment with all
 the requirements installed, and a database ready to use.
 
 Once this is done it's time to create some data and run the server::
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -18,8 +18,8 @@ You should then be able to provision the Vagrant box.
 
     $ vagrant up
 
-The box will be provisioned with Python 3.6.5, a virtual environment with all
-the requirements installed, and a database ready to use.
+The box will be provisioned with a latest Python 3.6 version, a virtual
+environment with all the requirements installed, and a database ready to use.
 
 Once this is done it's time to create some data and run the server::
 
--- a/provisioning/pythondotorg.yml
+++ b/provisioning/pythondotorg.yml
@@ -6,11 +6,6 @@
   vars_files:
     - env_vars.yml
   tasks:
-    # TODO: Remove this when we have a better 16.04 LTS box.
-    - name: Make /etc/apt/sources.list useful
-      become: yes
-      shell: "sed -i -- 's/#deb-src/deb-src/g' /etc/apt/sources.list && sudo sed -i -- 's/# deb-src/deb-src/g' /etc/apt/sources.list"
-
     - name: Run apt-get update
       become: yes
       apt: update_cache=yes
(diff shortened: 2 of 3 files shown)


provisioning/pythondotorg.yml
```diff
@@ -36,13 +31,6 @@
       changed_when:
         "result.stdout is defined and '0 upgraded, 0 newly installed, 0 to remove and' not in result.stdout"
 
-    - name: Install python-lxml deps
-      become: yes
-      apt: pkg=python-lxml state=build-dep
```
berkerpeksag, line 41: Does this get installed by BeautifulSoup now?

berkerpeksag (commented):
This looks good, but I haven't tested it locally yet. Could you also remove the commented-out line at line 29:

```yml
# - acl
```

docs/source/install.rst
```diff
@@ -18,7 +18,7 @@ You should then be able to provision the Vagrant box.
 
     $ vagrant up
 
-The box will be provisioned with Python 3.5, a virtual environment with all
+The box will be provisioned with Python 3.6.5, a virtual environment with all
```
berkerpeksag: Nit: How about saying something like "the latest Python 3.6 release" or "version" so we don't need to update the version information every six month :)

provisioning/pythondotorg.yml
Tomasz-Kluczkowski: Sure I will remove # - acl in the next amendment.
Regading lxml:
I have done local tests yesterday during the Python Sprint of the new box and was able to launch the site with initial data.
Also built the docs from txt to html and run the management command to do the same.
It looks like lxml required works now without specifying those dependencies.
I will retest all of that though just to make sure. Is there anything else that lxml is used for?

docs/source/install.rst
Tomasz-Kluczkowski: Ok, will do.
I need a day or two though as was really busy today - work / training :).

provisioning/pythondotorg.yml
berkerpeksag: > Is there anything else that lxml is used for?

I think BeautilfulSoup + lxml is only used by the PEP converter, so running `./manage.py test peps` may help to confirm this.

docs/source/install.rst
berkerpeksag: Sure, there is no rush :)

provisioning/pythondotorg.yml
Gautier: `lxml` is now available as a wheel, so `pip install` does not need to build it any-more which means we don't need the `pkg=python-lxml state=build-dep` step and we can also remove `libxml2-dev` and `libxslt-dev` from the `Install base packages` step.

I have tested locally, `./manage.py test peps` passes and I can generate the PEPs.

berkerpeksag: Awesome, thanks for testing @Gautier! @Tomasz-Kluczkowski could you remove those packages as well?

Tomasz-Kluczkowski: Working on it.

Tomasz-Kluczkowski: Added Gautier to the commit as a co-autor - he helped a lot during the sprint.

berkerpeksag (approved):

berkerpeksag:
Thanks, @Tomasz-Kluczkowski and @Gautier!

PR #1336, opened 2018-09-25
Cleanup blogs app

jaap3:
I might have gotten slightly carried away with cleaning this up, so feel free to reject this wholesale or ask me to create separate PRs for the things that are actually OK.

What I did:

- [x] Removed seemingly unused Contributor and Translation models (and related admin classes)
- [x] Used admin.register decorator to register the remaining admin classes
- [x] Add a "initial_data" factory for the Python Insider feed
- [x] Simplified the update_blogs management commands (It was almost possible to use `update_or_create`, but just not quite)
- [x] Made update_blog_supernav more robust (I'm not sure if that box is actually used though)
- [x] Strip tags in blog_supernav because I saw the "email this" etc. links inserted by feedburner in the summary field
- [x] Switch to https for the blog urls
- [x] Fix/tweak the test_feed_list test


blogs/management/commands/update_blogs.py
```diff
@@ -13,24 +13,14 @@ def handle(self, **options):
...
-                        url=entry['url'],
-                        feed=feed,
-                    )
+                url = entry.pop('url')
+                e, created = BlogEntry.objects.get_or_create(
+                    feed=feed, url=url, defaults=entry
+                )
+
+                # Update our info if it's changed
+                if not created and e.pub_date < entry['pub_date']:
```
jaap3: This `pub_date` check prevented me from just using `update_or_create`. I'm not sure if this check actually adds much, but I didn't want to mess with the logic too much.

blogs/parser.py
```diff
@@ -37,13 +38,20 @@ def _render_blog_supernav(entry):
...
+        pass
+    else:
+        rendered_box = _render_blog_supernav(latest_entry)
+        box, _ = Box.objects.update_or_create(
+            label='supernav-python-blog',
+            defaults={
+                'content': rendered_box,
+                'content_markup_type': 'html',
+            }
+        )
```
jaap3, line 57: This change makes it not rely on a hardcoded primary key and it doesn't crash when the feed doesn't exist or there are no BlogEntries at all. Which will happen on a "fresh" installation.

However, I'm not sure if the 'supernav-python-blog' is actually used anymore. I don't see it on the live site. So maybe the entire supernav stuff can be removed?

blogs/tests/test_templatetags.py
```diff
@@ -33,9 +29,6 @@ def test_get_latest_entries(self):
             entries[0].pub_date.isoformat(),
             '2013-03-04T15:00:00+00:00'
         )
-        b = Box.objects.get(label='supernav-python-blog')
-        rendered_box = _render_blog_supernav(BlogEntry.objects.latest())
-        self.assertEqual(b.content.raw, rendered_box)
```
jaap3, line 38: I removed this because it didn't test much and relied on a side effect of running the `update_blogs` and management command and the previously hardcoded primary key in the `update_blog_supernav` function. All while claiming to be a test for a template tag.

If updating the supernav is still required I can add proper tests.

blogs/tests/test_templatetags.py
```diff
@@ -61,14 +54,14 @@ def test_feed_list(self):
             summary='',
             pub_date=now(),
             url='path/to/foo',
-            feed=f1
+            feed=f2
```
jaap3, line 57: The second feed never got a blog entry, so this test was only testing that items from f1 were returned. It seems to me the intention was to test that entries of both feeds are returned.

templates/blogs/supernav.html
```diff
@@ -2,6 +2,6 @@
     <p class="date-posted"><small>{{ entry.pub_date|date:"l, F j, Y" }}</small></p>
     <h4>{{ entry.title }}</h4>
     <p class="excerpt">
-        <small>{{ entry.summary|truncatewords_html:50|safe }}<a class="readmore" href="{{ entry.url }}">Read more</a></small>
+        <small>{{ entry.summary|truncatewords_html:50|striptags|safe }}<a class="readmore" href="{{ entry.url }}">Read more</a></small>
```
jaap3, line 5: There was a serious potential for xss here, locally I could see the images FeedBurner inserts to share a blogpost.

blogs/management/commands/update_blogs.py
```diff
@@ -13,24 +13,14 @@ def handle(self, **options):
...
-                    BlogEntry.objects.create(
-                        title=entry['title'],
-                        summary=entry['summary'],
-                        pub_date=entry['pub_date'],
-                        url=entry['url'],
-                        feed=feed,
-                    )
+                url = entry.pop('url')
+                e, created = BlogEntry.objects.get_or_create(
+                    feed=feed, url=url, defaults=entry
```
berkerpeksag: Nit: Please add trailing comma.

blogs/models.py
```diff
@@ -57,48 +57,6 @@ class FeedAggregate(models.Model):
     def __str__(self):
         return self.name
 
-class Translation(ContentManageable):
```
berkerpeksag, line 60: Thanks for the cleanup!

blogs/parser.py
berkerpeksag: 'supernav-python-blog' is used in `static/js/script.js`.

Your rewrite is much better than the original code.

blogs/tests/test_templatetags.py
berkerpeksag: I think having more tests wouldn't hurt :) But it can be done in a separate PR.

blogs/parser.py
```diff
@@ -37,13 +38,20 @@ def _render_blog_supernav(entry):
...
-        label='supernav-python-blog',
-        defaults={
-            'content': rendered_box,
-            'content_markup_type': 'html',
-        }
-    )
+    try:
+        latest_entry = BlogEntry.objects.filter(
+            feed=Feed.objects.get(
+                feed_url=settings.PYTHON_BLOG_FEED_URL
```
berkerpeksag: Nit: Please add trailing comma.

blogs/management/commands/update_blogs.py
berkerpeksag: I think `e.pub_date < entry['pub_date']` was used because `get_or_create()` wasn't added to Django (just a guess, of course) at the time of writing. We can safely remove that check now. `created` should cover the same use case.

blogs/factories.py
```diff
@@ -0,0 +1,14 @@
+from django.conf import settings
+
+from .models import Feed
+
+
+def initial_data():
+    feed, created = Feed.objects.get_or_create(id=1, defaults={
```
berkerpeksag: Nit: We can rename `created` to `_` since it's unused.

berkerpeksag: Nit: Could you please move `id=1` to a new line?

```py
feed, _ = Feed.objects.get_or_create(
    id=1,
    defaults={
        ...
    },  # We probably don't need to add a trailing comma here as defaults is usually the last parameter.
)
```

berkerpeksag:
I think there's a bug here. After I ran `create_initial_data` the supernav for Downloads menu didn't change:

Screen Shot 2019-04-07 at 6 30 16 AM

I created a temporary `update_boxes` command:

```py
from django.core.management import BaseCommand, call_command


class Command(BaseCommand):

    help = 'Create initial data by using factories.'

    def handle(self, **options):
        from downloads.models import update_supernav
        update_supernav()
```

After I ran it:

Screen Shot 2019-04-07 at 6 30 56 AM

berkerpeksag:
I also applied this patch:

```diff
diff --git a/downloads/models.py b/downloads/models.py
index 305143b..72f40e4 100644
--- a/downloads/models.py
+++ b/downloads/models.py
@@ -126,35 +126,34 @@ class Release(ContentManageable, NameSlugModel):
 
 
 def update_supernav():
-    try:
-        latest_python2 = Release.objects.latest_python2()
-    except Release.DoesNotExist:
-        latest_python2 = None
-
     try:
         latest_python3 = Release.objects.latest_python3()
     except Release.DoesNotExist:
-        latest_python3 = None
+        # If we don't have a Python 3 release, we don't need to bother updating supernav.
+        return
 
     python_files = []
     for o in OS.objects.all():
         data = {
             'os': o,
-            'python2': None,
             'python3': None,
         }
 
-        if latest_python2:
-            data['python2'] = latest_python2.download_file_for_os(o.slug)
-
         if latest_python3:
-            data['python3'] = latest_python3.download_file_for_os(o.slug)
+            release_file = latest_python3.download_file_for_os(o.slug)
+            if not release_file:
+                return
+            data['python3'] = release_file
 
         python_files.append(data)
 
+    if not python_files:
+        return
+
+    if not all(f['python3'] for f in python_files):
+        return
+
     content = render_to_string('downloads/supernav.html', {
-        'latest_python2': latest_python2,
-        'latest_python3': latest_python3,
         'python_files': python_files,
     })
 
@@ -166,7 +165,20 @@ def update_supernav():
         }
     )
 
+
+def update_download_sources_box():
     # Update latest Sources box on Download landing page
+
+    try:
+        latest_python2 = Release.objects.latest_python2()
+    except Release.DoesNotExist:
+        latest_python2 = None
+
+    try:
+        latest_python3 = Release.objects.latest_python3()
+    except Release.DoesNotExist:
+        latest_python3 = None
+
     if latest_python2:
         latest_python2_source = latest_python2.download_file_for_os('source')
     else:
@@ -177,6 +189,9 @@ def update_supernav():
     else:
         latest_python3_source = None
 
+    if not latest_python2_source or not latest_python3_source:
+        return
+
     source_content = render_to_string('downloads/download-sources-box.html', {
         'latest_python2_source': latest_python2_source,
         'latest_python3_source': latest_python3_source,
@@ -260,14 +275,14 @@ def purge_fastly_download_pages(sender, instance, **kwargs):
 
 
 @receiver(post_save, sender=Release)
-def update_download_supernav(sender, instance, **kwargs):
-    """ Update download supernav """
+def update_download_supernav_and_boxes(sender, instance, **kwargs):
     # Skip in fixtures
     if kwargs.get('raw', False):
         return
 
     if instance.is_published:
         update_supernav()
+        update_download_sources_box()
         update_homepage_download_box()
 ```

berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1337, opened 2018-09-25
Fix comment about the creator field

jaap3:
The creator ForeignKey sets blank=True, so the comment isn't correct. Since setting creator is enforced by ContentManageableAdmin anyway I don't think it makes sense to change blank to False.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -22,11 +22,7 @@ class ContentManageable(models.Model):
 
     # We allow creator to be null=True so that we can, if we must, create a
     # ContentManageable object in a context where we don't have a creator (i.e.
-    # where there isn't a request.user sitting around). However, we still leave
-    # it blank=False so that any time we try to *validate* a ContentManageable
-    # object we'll get an error. This is a reasonable compromise that lets us
-    # track creators fairly well without necessarily over-enforcing it in places
-    # where it'd be invasive.
+    # where there isn't a request.user sitting around).
     creator = models.ForeignKey(
         settings.AUTH_USER_MODEL,
         related_name='%(app_label)s_%(class)s_creator',


cms/models.py
```diff
@@ -22,11 +22,7 @@ class ContentManageable(models.Model):
 
     # We allow creator to be null=True so that we can, if we must, create a
     # ContentManageable object in a context where we don't have a creator (i.e.
-    # where there isn't a request.user sitting around). However, we still leave
-    # it blank=False so that any time we try to *validate* a ContentManageable
-    # object we'll get an error. This is a reasonable compromise that lets us
-    # track creators fairly well without necessarily over-enforcing it in places
-    # where it'd be invasive.
+    # where there isn't a request.user sitting around).
```
berkerpeksag, line 25: Isn't `request.user` always defined? Even if we don't have a `User` instance, it would still return an `AnonymousUser` instance. I'd like to do some digging on this before making a decision about this use case.

jaap3: I think that this comment is talking about the situation where you create a `ContentManageable` object outside of the request/response cycle (so there would no `request` object around, and therefore also no `request.user`.

In those cases there might not be any user responsible for the creation of the object anyway. e.g. creating an object through a fixture, or as a result of some automated action, like an importer that creates `ContentManageable` objects. In those cases it makes sense to omit the creator field entirely.

Now it seems that the original intention was to enforce setting a creator when creating `ContentManageable` objects when using a form. By setting `blank=False` a form would raise a validation error when no creator was provided with the form data.

But this is not how setting a creator is enforced. Instead it's done during the saving of the object in the admin (which makes more sense than selecting on in a form). I don't know if any `ContentManageable` object is created by users *outside* of the admin, but in those cases the `creator` field might be left empty unintentionally.

All in all, the comment claims that a flag is False, when it's actually set to True so the claim it's making doesn't hold.

berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1338, opened 2018-09-25
Use update_or_create instead of custom create_or_update_model

jaap3:
This custom implementation is probably a holdover of the days before Django gained a `update_or_create` method


berkerpeksag (approved):

berkerpeksag:
Good catch, thanks!

PR #1340, opened 2018-09-25
Tweak purge_url

jaap3:


jaap3:
I've updated this PR so purging multiple urls will re-use the same `requests.Session`, which was my intention in the first place but I somehow forgot to actually do.

fastly/utils.py
```diff
@@ -1,22 +1,20 @@
 import requests
 
+from urllib.parse import urljoin
+
 from django.conf import settings
 
 
-def purge_url(path):
+def purge_url(*paths):
```
berkerpeksag: What do you think about renaming this to `purge_urls()`?

jaap3: Done

berkerpeksag (approved):

berkerpeksag:
Nice cleanup, thank you very much!

PR #1343, opened 2018-09-25
Reuse convenient model properties in job feed

jaap3:
Both `display_name` and `display_location` were more or less duplicated. 

These changes result is slightly different output: 

`display_name` joins `job_title` and `company_name` with a comma instead of a dash.

`display_location` joins the location parts with a ', ' instead of just a ',' and has a less strict check on the each location part (just `if location_part`).


berkerpeksag (approved):

berkerpeksag:
Nice, thanks!

PR #1386, opened 2019-02-01
Update Freenode's "register your nickname" link on Community page

m-aciek:
Current link returns HTTP 404. It changed to `https://freenode.net/kb/answer/registration` from `http://freenode.net/faq.shtml#registering`.

I'm not sure if fixture change is enough to fix it in production (maybe just database change is needed?).

[Community page.](https://www.python.org/community/)

obraz

### DIFF
--- a/fixtures/boxes.json
+++ b/fixtures/boxes.json
@@ -128,7 +128,7 @@
 {
     "fields": {
         "label": "community-irc-channels",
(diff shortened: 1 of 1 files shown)


Mariatta (approved):
Thanks!

berkerpeksag (approved):
You're right that we need to update the content via Django admin as well, but it's good to fix this in the fixture. Thank you!

m-aciek:
@berkerpeksag Actually, the link haven't been corrected via Django admin yet. How should I ping maintainers for it to be done?

berkerpeksag:
@m-aciek updated! Thank you for pinging us.

PR #1404, opened 2019-04-05
Change state to present (from installed) on Install PostgresSQL

eamanu:
The installed state is deprecated.

### DIFF
--- a/provisioning/pythondotorg.yml
+++ b/provisioning/pythondotorg.yml
@@ -32,7 +32,7 @@
       gem: name=bundler user_install=no
 
     - name: Install PostgreSQL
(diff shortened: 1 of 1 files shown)


berkerpeksag (commented):
Do you know when 'present' was added? I'd like keep `pythondotorg.yml` compatible with earlier 2.0 releases, so if it was added in 2.6 or 2.7, I prefer to keep 'installed' for now.

berkerpeksag (approved):
It was deprecated in 2014 in https://github.com/ansible/ansible-modules-core/commit/bbebdefacb749bc9b5203da9b96a98e1dac3354b.

LGTM, thank you!

Issue #1402, opened 2019-03-30
Install.rst has an incorrect step


eamanu:
**Describe the bug**
On https://github.com/python/pythondotorg/blob/master/docs/source/install.rst
one of the step is activate the virtual environment (sound how the venv is created)
but venv don't exist.

**To Reproduce**
1. Go to `https://github.com/python/pythondotorg/blob/master/docs/source/install.rst`
2. Read and run:
```
# Go to the pythondotorg/ directory and activate the virtual environment.
$ cd ~/pythondotorg
$ . venv/bin/activate
```
3. See venv does not exist:
![image](https://user-images.githubusercontent.com/7605307/55282665-a9da5980-5327-11e9-8bd8-0b9f96f247cb.png)

berkerpeksag:
Thank you for the report! This is weird. I can see `venv` in my setup:

```sh
vagrant@vagrant:~/pythondotorg$ ls | grep venv
venv
```

Do you have some logs from the provisioning step?

berkerpeksag:
Just to make sure, I've started a fresh clone (without `venv` directory etc.) and it worked for me. It would be helpful if we know what versions of Vagrant and Ansible are installed on your machine and see logs of `vagrant up` (or `vagrant provision`) command.

berkerpeksag:
By the way, I've made some tweaks to installation instructions (https://github.com/python/pythondotorg/commit/e12cdbf9cfa6caead9f4f2535dd9a09f0d30442a) and fixed a bug (https://github.com/python/pythondotorg/commit/34119ad27e29f785918bcbfb9422c510502d2125) that may annoy or confuse new contributors.

eamanu:
Hello, 

Sorry for the delay of my response.

My versions are:

```
Vagrant 2.2.4
ansible 2.7.9
```
I am new on the use of vagrant, the ```vagrant provision``` run automatically?

Despite this,  when I run ```vagrant provision``` I have this error:

```
TASK [Install dependencies for the Django app inside virtual environment] ****** 
fatal: [default]: FAILED! => {"changed": false, "msg": "No setuptools found in remote host, please install it first."}
```

eamanu:
The solution was set the vars `ansible_python_interpreter: "/usr/bin/python3"` I wil make the PR adding this vars.

[...]

PR #1405, opened 2019-04-05
Add ansible_python_interpreter on "Install dependencies for the Djangoapp inside virutal environment"

eamanu:
Add ansible_python_interpreter : "/usr/bin/python3"  to avoid
"setuptools is not installed" error message if python2 is installed.

### DIFF
--- a/provisioning/pythondotorg.yml
+++ b/provisioning/pythondotorg.yml
@@ -76,6 +76,8 @@
            update=no
 
     - name: Install dependencies for the Django app inside virtual environment
+      vars:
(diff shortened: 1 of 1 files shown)


berkerpeksag (changes requested):
We are already installing setuptools globally in line 22 and using Python 3 interpreter when creating a new virtual environment (see line 80: `virtualenv_command="/usr/bin/python3 -m venv"`). I don't think this PR should be merged. We need to investigate your problem first.

berkerpeksag:
Closing per https://github.com/python/pythondotorg/issues/1402#issuecomment-482464564.

PR #1410, opened 2019-04-08
Add 'vagrant up' install.rst on "next time" section

eamanu:
Before run ```vagrant ssh``` the VM must be power on.
So, the step ```vagrant up``` is added.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -41,11 +41,13 @@ Now use your favorite browser to go to http://localhost:8001/. The admin pages
 can be found at http://localhost:8001/admin/. You can use your superuser
 credentials to log in to Django admin.
 
-You will only need to run the following two commands the next time you want to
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -44,6 +44,7 @@ credentials to log in to Django admin.
 You will only need to run the following two commands the next time you want to
```
berkerpeksag: two -> four

Could you also add

```sh
$ . venv/bin/activate
```

to the command list? I forgot to add it.

berkerpeksag (approved):

docs/source/install.rst
eamanu: Ready, I add two more command. 1. ```cd ~/pythondotorg``` and 2. ```. venv/bin/activate```

berkerpeksag:
Thanks!

PR #1411, opened 2019-04-08
Add new entries on .gitignore

eamanu:
Add the words ```__pycache__/``` and ```docs/build/``` to
ignore on .gitignore


berkerpeksag (commented):
Thank you, but we suggest using a [global `.gitignore`](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) for these.

PR #1412, opened 2019-04-09
Fix can't generate pep pages with local pep repo on local

mozillazg:
That issue is the result of https://github.com/python/pythondotorg/pull/1389 .

### DIFF
--- a/peps/management/commands/generate_pep_pages.py
+++ b/peps/management/commands/generate_pep_pages.py
@@ -51,26 +51,7 @@ class Command(BaseCommand):
         verbose("== Starting PEP page generation")
 
         with ExitStack() as stack:
-            verbose(f"== Fetching PEP artifact from {settings.PEP_ARTIFACT_URL}")
-            peps_last_updated = get_peps_last_updated()
-            with requests.get(settings.PEP_ARTIFACT_URL, stream=True) as r:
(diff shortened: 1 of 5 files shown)


peps/management/commands/generate_pep_pages.py
```diff
@@ -129,3 +110,30 @@ def verbose(msg):
                     verbose("- Skipping non-PEP related image '{}'".format(img))
 
         verbose("== Finished")
+
+    def get_artifacts_path(self, stack, verbose):
+        if settings.DEBUG and settings.PEP_REPO_PATH:
```
berkerpeksag, line 115: I don't think we should go back to the old method. Downloading and extracting an artifact from S3 sounds like a better idea to me.

I think we can set `settings.PEP_ARTIFACT_URL` to `os.path.join(settings.BASE, 'peps/tests/peps.tar.gz')` in `settings/local.py`. This way contributors wouldn't need to clone python/peps repo if they would like to work on our PEP converter code.

berkerpeksag (commented):
Thank you! Just to clarify, this PR reverts to old Git-based implementation in development mode and https://github.com/python/peps/pull/982 fixes the problem with reST links, right?

peps/management/commands/generate_pep_pages.py
mozillazg: 👌

PR #1413, opened 2019-04-11
implement naive self-service banners for python.org

ewdurbin:
Allows management of banners displayed across python.org via Django admin.

Screen Shot 2019-04-11 at 10 24 09 AM
Screen Shot 2019-04-11 at 10 24 12 AM
Screen Shot 2019-04-11 at 10 24 25 AM


banners/templatetags/banners.py
```diff
@@ -0,0 +1,23 @@
...
+from django.template.loader import render_to_string
+
+from banners.models import Banner
+
+register = template.Library()
+
+
+@register.simple_tag
+def render_active_banner(psf_pages_only=True):
+    if not psf_pages_only:
```
berkerpeksag: If I understood this correctly, the following line would cover both if and else statements:

```py
banner = Banner.objects.filter(active=True, psf_pages_only=psf_pages_only).first()
```

banners/admin.py
```diff
@@ -0,0 +1,9 @@
+from django.contrib import admin
+
+from banners.models import Banner
+
+
+class BannerAdmin(admin.ModelAdmin):
+    list_display = ("title", "active", "psf_pages_only")
+
+admin.site.register(Banner, BannerAdmin)
```
berkerpeksag: Style nit: We are trying to use the `@admin.site.register()` decorator version in new code.

banners/migrations/0002_auto_20190411_1521.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 2.0.9 on 2019-04-11 15:21
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
berkerpeksag: Do you plan to squash these two migrations when design of the `Banner` model is finished?

templates/downloads/index.html
```diff
@@ -38,6 +39,8 @@ <h1 class="call-to-action">Download the latest version of Python</h1>
 {% block content %}
                 <div class="row download-list-widget">
 
+                    {% render_active_banner False %}
```
berkerpeksag: Nit: Passing `psf_pages_only=False` would increase readability here.

templates/psf/default.html
```diff
@@ -26,6 +27,8 @@
 {% block breadcrumbs %}
     {% load sitetree %}
 
+    {% render_active_banner True %}
```
berkerpeksag: Nit: Did Django complained without passing an argument? I don't have a computer to test at the moment, but we might just use `{% render_active_banner %}` here.

banners/templatetags/banners.py
```diff
@@ -0,0 +1,23 @@
+from django import template
+from django.template.loader import render_to_string
+
+from banners.models import Banner
+
+register = template.Library()
+
+
+@register.simple_tag
+def render_active_banner(psf_pages_only=True):
```
jaap3: What about having two template tags `render_active_banner` and `render_active_psf_banner` instead of adding a boolean argument? This would also make it clearer what is going on in the template.

banners/templatetags/banners.py
```diff
@@ -0,0 +1,23 @@
...
+    if not psf_pages_only:
+        banner = Banner.objects.filter(active=True, psf_pages_only=psf_pages_only).first()
+    else:
+        banner = Banner.objects.filter(active=True).first()
```
jaap3: It would be nice to have a custom `QuerySet`/`Manager` that provides an `active()` method (i.e. `Banner.objects.active().first()`)

banners/templatetags/banners.py
```diff
@@ -0,0 +1,23 @@
...
+    if banner is not None:
+        tmpl = template.loader.get_template('banners/banner.html')
+        ctx = {
+            'message': banner.message,
+            'title': banner.title,
+            'link': banner.link,
+        }
+        return tmpl.render(ctx)
```
jaap3: `render_to_string` does the same in less lines:

https://docs.djangoproject.com/en/2.2/topics/templates/#django.template.loader.render_to_string

banners/templatetags/banners.py
ewdurbin: great suggestion, thank you!

templates/psf/default.html
ewdurbin: with suggestion from jaap3, this isn't necessary. I was just being explicit :)

templates/downloads/index.html
ewdurbin: again, will be addressed with excellent suggestion from jaap3

banners/migrations/0002_auto_20190411_1521.py
ewdurbin: aye

banners/templatetags/banners.py
ewdurbin: little trickier than that, since we accept either value for the PSF pages. so filtering on psf_pages_only makes general banners _not_ show.

PR #1415, opened 2019-04-13
Always create a new venv when provisioning

jaap3:
I still had an old vagrant box and venv laying around, but master now only works with Python 3.6.

Running `vagrant destroy && vagrant up` resulted in a strange error during provisioning. This was cause by the old `venv` that remained.

With this PR any `venv` in the root of the project is removed when setting up vagrant box for the first time or when running `vagrant up --provision`

### DIFF
--- a/provisioning/pythondotorg.yml
+++ b/provisioning/pythondotorg.yml
@@ -75,6 +75,11 @@
            clone=no
            update=no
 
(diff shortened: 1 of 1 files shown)


berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1422, opened 2019-04-15
Misc cleanup

jaap3:
I noticed that I unintentionally reused the key argument when I partially rewrote the `_get_id` function for the downloads factory. I didn't want to create a PR for it, but then I found this old branch where I removed some unused stuff, so maybe it can go in as some general spring cleaning?

### DIFF
--- a/codesamples/views.py
+++ b/codesamples/views.py
@@ -1,1 +1,0 @@
-# Create your views here.
--- a/community/models.py
+++ b/community/models.py
@@ -12,17 +12,6 @@ from .managers import PostQuerySet
 
 DEFAULT_MARKUP_TYPE = 'html'  # getattr(settings, 'DEFAULT_MARKUP_TYPE', 'restructuredtext')
 
-'''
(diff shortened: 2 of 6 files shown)


community/models.py
```diff
@@ -12,17 +12,6 @@
 
 DEFAULT_MARKUP_TYPE = 'html'  # getattr(settings, 'DEFAULT_MARKUP_TYPE', 'restructuredtext')
```
berkerpeksag: Could you delete the commented-out code as well?

codesamples/views.py
```diff
@@ -1 +0,0 @@
-# Create your views here.
```
berkerpeksag, line 1: We can delete these too:

```
membership/models.py:1:# Create your models here.
work_groups/views.py:1:# Create your views here.
```

berkerpeksag (approved):

berkerpeksag:
Thanks!

PR #1424, opened 2019-04-15
Replace calls to admin.site.register with @admin.register decorator 

jaap3:
Started with a commit cherry picked from #1336, but then I just continued and replaced all calls to `admin.site.register` :)

### DIFF
--- a/events/admin.py
+++ b/events/admin.py
@@ -9,10 +9,6 @@ class EventInline(admin.StackedInline):
     extra = 0
 
 
-class CalendarAdmin(ContentManageableModelAdmin):
-    pass
-
-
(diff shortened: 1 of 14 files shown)


events/admin.py
```diff
@@ -9,10 +9,6 @@ class EventInline(admin.StackedInline):
     extra = 0
 
 
-class CalendarAdmin(ContentManageableModelAdmin):
```
jaap3, line 12: Removed this because it didn't do anything

berkerpeksag (approved):
Yes, replacing all instances in single PR is much better especially if diff is small :)

berkerpeksag:
Thanks!

PR #1427, opened 2019-04-17
[WIP] Add purge_fastly_cache management command

berkerpeksag:


Procfile
```diff
@@ -1,2 +1,2 @@
 release: python manage.py migrate --noinput
-web: bin/start-nginx gunicorn -c gunicorn.conf pydotorg.wsgi
+web: python manage.py purge_fastly_cache && bin/start-nginx gunicorn -c gunicorn.conf pydotorg.wsgi
```
berkerpeksag, line 2: I think we may need to move this into a `run.sh` in case `Procfile` doesn't support `&&`.

downloads/models.py
```diff
@@ -248,19 +264,9 @@ def purge_fastly_download_pages(sender, instance, **kwargs):
...
-        # Purge our common pages
-        purge_url('/downloads/')
-        purge_url('/downloads/latest/python2/')
-        purge_url('/downloads/latest/python3/')
-        purge_url('/downloads/mac-osx/')
-        purge_url('/downloads/source/')
-        purge_url('/downloads/windows/')
+        _purge_download_pages()
         if instance.get_version() is not None:
             purge_url('/ftp/python/{}/'.format(instance.get_version()))
```
berkerpeksag, line 269: It would be nice to find a way to purge these URLs in the command as well. I think using surrogate keys may help here: https://docs.fastly.com/guides/api-caching/purging-api-cache-with-surrogate-keys (I'm not sure whetler I can create them without a Fastly account?)

fastly/management/commands/purge_fastly_cache.py
```diff
@@ -0,0 +1,11 @@
+from django.core.management import BaseCommand
+
+from fastly import register
+
+
+class Command(BaseCommand):
+
+    def handle(self, **options):
+        for cmd in register:
```
berkerpeksag, line 9: We may need to check for network exceptions here.

fastly/__init__.py
```diff
@@ -0,0 +1,6 @@
+registry = []
+
+
+def register_function(func):
```
berkerpeksag, line 4: I'm open to suggestions here. I think `register_function()` is too generic.

downloads/models.py
ewdurbin: Surrogate Keys are set by simply including the header `Surrogate-Key`, no need to register them.

ewdurbin: Ah, I see the rub here. Since the /ftp urls are served by [another host](https://github.com/python/psf-salt/blob/196d03fe8b2b6d846032e8940a4b172d1fd82215/salt/downloads/config/nginx.downloads-backend.conf#L19-L23)... It is unfortunate that we already have a Surrogate-Key there that is static... and somewhat painful if we ever purge it (downloads box can only serve _so much_ bandwidth).

Procfile
ewdurbin: kinda stinks that normal heroku apps don't have a "post deploy" phase.

it'd be ideal to issue the command after the app is up and running. this solution is a bit prone to a race condition, as the purge is issued before the app is up. traffic may still get the old app.

fastly/__init__.py
ewdurbin: naming things... hmmmm. I don't have any improvement.

Procfile
berkerpeksag: Yes, unfortunately I couldn't find a way to do that. I'm not sure if changing the order would help.

dogweather:
(Is there a better place to discuss the strategy? Please let me know if so.)

It looks like we're doing full page caching. Do we know that it'd be prohibitively expensive to simply not cache the pages? The savings in maintenance and complexity are big.

In my experience with even slower languages than Python, this kind of caching isn't necessary. All hits to the app's paths are handled by the app. Assets are fingerprinted, cached and CDN-served, like we're already doing, plus other auto-expiring methods.

Or even, use a static HTML generator for pages like these that don't change for each user.

PR #1428, opened 2019-04-20
Support use local file path as value of PEP_ARTIFACT_URL

mozillazg:
> I think we can set settings.PEP_ARTIFACT_URL to os.path.join(settings.BASE, 'peps/tests/peps.tar.gz') in settings/local.py. This way contributors wouldn't need to clone python/peps repo if they would like to work on our PEP converter code. 
> ---- @berkerpeksag  https://github.com/python/pythondotorg/pull/1412#discussion_r273556541

@berkerpeksag What do you think about this PR? Is it necessary?

### DIFF
--- a/peps/management/commands/generate_pep_pages.py
+++ b/peps/management/commands/generate_pep_pages.py
@@ -52,20 +52,7 @@ class Command(BaseCommand):
 
         with ExitStack() as stack:
             verbose(f"== Fetching PEP artifact from {settings.PEP_ARTIFACT_URL}")
-            peps_last_updated = get_peps_last_updated()
-            with requests.get(settings.PEP_ARTIFACT_URL, stream=True) as r:
-                artifact_last_modified = parsedate(r.headers['last-modified'])
-                if peps_last_updated > artifact_last_modified:
-                    verbose(f"== No update to artifacts, we're done here!")
-                    return
-
-                temp_file = stack.enter_context(TemporaryFile())
-                for chunk in r.iter_content(chunk_size=8192):
-                    if chunk:
-                        temp_file.write(chunk)
-
-            temp_file.seek(0)
-
+            temp_file = self.get_artifact_tar_ball(stack, verbose)
             temp_dir = stack.enter_context(TemporaryDirectory())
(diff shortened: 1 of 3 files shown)


peps/management/commands/generate_pep_pages.py
```diff
@@ -129,3 +116,24 @@ def verbose(msg):
                     verbose("- Skipping non-PEP related image '{}'".format(img))
 
         verbose("== Finished")
+
+    def get_artifact_tar_ball(self, stack, verbose):
+        artifact_url = settings.PEP_ARTIFACT_URL
+        if not (artifact_url.startswith('http://') or
```
berkerpeksag:
```suggestion
        if not artifact_url.startswith(('http://', 'https://')):
```

peps/management/commands/generate_pep_pages.py
```diff
@@ -129,3 +116,24 @@ def verbose(msg):
                     verbose("- Skipping non-PEP related image '{}'".format(img))
 
         verbose("== Finished")
+
+    def get_artifact_tar_ball(self, stack, verbose):
```
berkerpeksag:
```suggestion
    def get_artifact_tarball(self, stack, verbose):
```

berkerpeksag (commented):
I think we need to change the following line:

https://github.com/python/pythondotorg/blob/f1747e7d704987d7ca7f9f29112a97f0a66a67f4/pydotorg/settings/local.py#L27-L28

to

```py
PEP_ARTIFACT_URL = os.path.join(BASE, 'peps/tests/peps.tar.gz')
```

peps/management/commands/generate_pep_pages.py
```diff
@@ -129,3 +116,24 @@ def verbose(msg):
...
+        artifact_url = settings.PEP_ARTIFACT_URL
+        if not (artifact_url.startswith('http://') or
+                artifact_url.startswith('https://')):
```
berkerpeksag:
```suggestion

```

berkerpeksag (approved):

berkerpeksag:
Thank you!

PR #1432, opened 2019-05-01
Add link to Github website repository

Master29f:


eamanu (commented):
Hi @Master29f 

I think that is a good improve on the website. But, IMHO I don't think that 
```Edit Website at Github``` is the correct phrase.  Because, we can not
edit the website directly  without control.

I would like something like: ```Get  the source code``` or something  like this.

Is just an opinion :-)

Master29f:
But I think, that then the visitor think, that they get the source code of python and not of the website

Master29f:
Maybe "web source code" is a good idea

berkerpeksag (commented):
Thank you for your PR, but unfortunately this can cause more confusion than it solves. The content of python.org isn't editable via GitHub, and it's easy to confuse 'source code' with different Python projects and even Python itself (as you already said) The "Submit Website Bug" should give users enough hint to find python.org's source code.

Master29f:
Perhaps there could be "Python@Github" and this link go direct to the official python acc

PR #1434, opened 2019-05-09
[WIP] started adding django 3.2 support

auvipy:
some packages like Django-pipeline lack Django 2.2 support. Will address them one by one.

### DIFF
--- a/base-requirements.txt
+++ b/base-requirements.txt
@@ -1,11 +1,11 @@
 dj-database-url==0.5.0
 django-pipeline==1.6.14
 django-sitetree==1.10.0
-Django==2.0.13
+Django==2.2.1
 docutils==0.12
 Markdown==2.5.2
 Pillow==5.1.0
-psycopg2==2.7.3.2
+psycopg2==2.8.2
 python3-openid==3.1.0
 # lxml used by BeautifulSoup.
 lxml==4.1.1
--- a/base-requirements.txt
+++ b/base-requirements.txt
@@ -26,12 +26,12 @@ python-dateutil==2.6.1
 
 requests[security]>=2.20.0
 
-django-honeypot==0.6.0
-django-markupfield==1.4.3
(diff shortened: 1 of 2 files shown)


berkerpeksag (changes requested):
We shouldn't upgrade from 2.0 to 2.2. The proper way to do this is upgrade to 2.1 first. And I'd suggest to create a some sort of plan before starting to work on this. For example, why django-pipeline doesn't work with Django 2.2? What's the action plan? Add Django 2.2 support to django-pipeline? Or use something else?

auvipy:
https://github.com/jazzband/django-pipeline/pull/693

auvipy:
thanks for your input. Some of them are certainly valid. on the 2.0-2.1 upgrade, it's really not necessary as Django LTS to LTS upgrade is much easier than before. So I would like to stand on the ground of django 2.0 to 2.2 upgrade. definitely with proper plans and rigorous testing.

auvipy:
django-pipeline is now on @jazzbrand/django-pipeline so I am free to contribute and release. Regarding using something else rather than django-pipeline I am open to looking for more options.

berkerpeksag:
Sorry, I'm not going to merge a PR that attempts to upgrade Django from X to X+2 directly. Django release notes aren't cover all behavior changes, so it's much easier to detect regressions since the scope of the changes are much smaller between X and X+1.

auvipy:
I am not accepting to get this PR merged anytime soon. As It's natural to complete many things which need to be addressed first. I can start another pr with X to X+1 upgrade. I do also want to experience your concerns are up to date.

auvipy:
re_path or new simpler regex free URL paths which one do you prefer for this project?

auvipy:
used both

hugovk:
Closed because https://github.com/python/pythondotorg/pull/2520 has upgraded to 4.2. Thanks anyway!

Issue #1267, opened 2018-05-27
Migrate website to Python 3.7


datapythonista:
The website currently works with Python 3.4, which starts to be quite old. CPython 3.7 is already available in beta, and we can make the website work if it (if not already, since Django has been migrated recently to 1.11 and it was the previous version of Django that wasn't compatible with Python > 3.4).

---

**Blockers:**

* [ ] https://github.com/kurtmckee/feedparser/pull/131

berkerpeksag:
Note that we are currently working on to make python.org run on Python 3.6. We can add Python 3.7 to `.travis.yml` once the feedparser PR is merged.

scottilee:
@berkerpeksag it looks like feedparser works with Python 3.7 now: https://github.com/kurtmckee/feedparser/pull/131.

berkerpeksag:
Unfortunately, the master branch of feedparser hasn't been released yet: https://pypi.org/project/feedparser/#history

verhovsky:
@berkerpeksag it's been released now (October 25, 2020).

hugovk:
The codebase was upgraded to Python 3.9 in https://github.com/python/pythondotorg/pull/2208.

@datapythonista Please can you close this issue?

PR #1440, opened 2019-05-26
Add Docker development option

scottilee:
- Added a `Dockerfile` and `docker-compose.yml` to bring up python.org in a container along with a Postgres container.
- Updated installation documentation


Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR /web
+COPY . /web
+
+RUN apk update
+RUN apk add --no-cache --virtual .build-deps \
+    gcc \
```
berkerpeksag, line 8: Can we split core and app dependencies into separate steps?

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM python:3.7-alpine
```
berkerpeksag: feedparser dependency isn't compatible with Python 3.7. It should be 3.6 for now.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
```
berkerpeksag: Why did you set these env variables? İs it just c/p from an existing project or did you have any problems?

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    gcc \
+    python3-dev \
+    libffi-dev \
+    libxml2-dev \
+    libxslt-dev \
+    linux-headers \
+    musl-dev \
+    postgresql-dev \
+    ruby \
+    ruby-rdoc \
```
berkerpeksag, line 14: We don't need ruby-rdoc.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    zlib-dev
+RUN apk add git jpeg-dev libpq
+
+RUN gem install bundler \
+    && bundle install
+
+RUN pip install pip --upgrade \
```
berkerpeksag: pip is pretty prone to introducing regressions. I rather not pass `--upgrade`.

docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: "3.7"
+
+services:
+  web:
+    build: .
+    command: python manage.py runserver 0.0.0.0:8000
+    environment:
+      DATABASE_URL: "${DB_TYPE:-postgres}://${DB_USER:-admin}:${DB_PASSWORD:-password}@${DB_HOST:-db}:${DB_PORT:-5432}/${DB_NAME:-pythondotorg}"
```
berkerpeksag: We already have `DATABASE_URL` environment variable: https://pythondotorg.readthedocs.io/install.html#manual-setup

docs/source/install.rst
```diff
@@ -1,15 +1,42 @@
...
+include docker-compose which can be used to bring up the application:
+
+::
+
+    $ docker-compose up
+
+docker-compose will bring up the application container and a PostgreSQL 
+container. The website will be available at http://localhost:8000.
+
+Each time changes are made to local files the application container
```
berkerpeksag: I don't think this is true. We'd need to rebuild it if we'd updated dependencies, but we shouldn't have to rebuild it each time we changed Python files.

docs/source/install.rst
```diff
@@ -1,15 +1,42 @@
...
+will need to be rebuilt:
+
+::
+
+    $ docker-compose up --build
+    # You can also make changes inside the application container
+    $ docker exec -it $(docker ps | grep "pythondotorg_web" | awk '{print $1}') /bin/ash
```
berkerpeksag: Sorry, but this looks horrible :) I'd rather use something like `docker-compose exec pythondotorg_web sh`.

entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+
+  while ! nc -z $DB_HOST $DB_PORT; do
+    sleep 0.1
+  done
+
+  echo "PostgreSQL started"
+fi
+
+python manage.py migrate
+yes | python manage.py create_initial_data
```
berkerpeksag, line 4: Do we need to invoke this each time? It seems to me that we can make the `postgres_data` volume persistent.

entrypoint.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env sh
+
+if [[ "${DB_TYPE}" == "postgres" ]]; then
```
berkerpeksag: Aren't we already depending on `db` service? I think running `db` with the `-d` option is pretty common and it'd speed setting up environment quite a bit.

entrypoint.sh
berkerpeksag: And we can pass ` --noinput` to `migrate`.

Dockerfile
scottilee: I don't understand what you mean here. Could you explain what core and app dependencies are?

Dockerfile
scottilee: Will fix.

Dockerfile
scottilee: Mostly practice set by other projects I guess. Let me know if you want me to remove them.

Dockerfile
scottilee: I added it because the container exits with an error about rdoc. I also tried `gem install bundler --no-rdoc` but that didn't work. Let me know if you have any other suggestions.

Dockerfile
scottilee: Will remove.

docker-compose.yml
scottilee: Could you explain how it works because if I don't set the `DATABASE_URL` variable the web container will fail.

docs/source/install.rst
scottilee: My understanding is that the Dockerfile will copy the directory over on build (with any changes or updates) but if that's wrong I will remove this line.

docs/source/install.rst
scottilee: Will fix.

entrypoint.sh
scottilee: I can remove the entrypoint script. Could you let me know how to make the postgres_data volume persistent and how to deal with doing the `migrate` and `create_initial_data` separately?

Dockerfile
scottilee: Here is the specific error:
```
ERROR:  While executing gem ... (Gem::DocumentError)
    RDoc is not installed: cannot load such file -- rdoc/rdoc
```

Dockerfile
scottilee: I went ahead and removed them as not having them didn't produce issues.

entrypoint.sh
scottilee: This looks like an interesting idea but I don't know if it would satisfy what you're looking for: https://stackoverflow.com/a/44283611

Dockerfile
berkerpeksag: Sorry, I should have been clearer.

By app deps, I meant python3-dev, postgresql-dev, ruby etc. (i.e. we need them to run python.org inside the container)

By core deps, I meant gcc, linux-headers, musl-dev etc. (i.e. we need them install our app dependencies)

Dockerfile
berkerpeksag: That's weird. I'll look into it, thanks! I thought we've removed it from the Ansible configuration.

docker-compose.yml
berkerpeksag: What I meant is we can just specify `DATABASE_URL: "postgres://admin:password@db:5432/pythondotorg"` instead of introducing `DB_*` env variables.

docs/source/install.rst
berkerpeksag: I'll check this out later. We are using Docker at $WORK and I don't remember if I had to rebuild it each time I made changes to the application code.

entrypoint.sh
scottilee: Will remove.

Dockerfile
scottilee: Will fix.

scottilee:
@berkerpeksag I did the following, let me know what you want to do about `postgres_data` persistence and the entrypoint script.

- [X] Split out "app" and "core" dependencies
- [X] Move Dockerfile from `python:3.7-alpine` to `python:3.6-alpine` for feedparser dependency
- [X] Removed `PYTHONDONTWRITEBYTECODE 1` and `PYTHONUNBUFFERED 1` env vars
- [X] Removed `pip install pip --upgrade` due to possible regressions
- [X] Paired down the `DATABASE_URL` env var
- [X] Removed re-build explanation in "install" docs and simplified the exec command

docs/source/install.rst
```diff
@@ -1,15 +1,39 @@
...
 ==========
 
-Here are two ways to hack on python.org:
+Here are three ways to hack on python.org:
 
-1. :ref:`vagrant-setup`
-2. :ref:`manual-setup`
+1. :ref:`docker-setup`
+2. :ref:`vagrant-setup`
+3. :ref:`manual-setup`
```
eamanu: I would like an alphabetical enumeration :-). First docker, second manual setup and finally vagrant setup

Just an opinion

eamanu (commented):
I like the idea :-)

docs/source/install.rst
scottilee: Do you mean the following?
```
A. Docker Setup
B. Manual Setup
C. Vagrant Setup
```

If so, as long as @berkerpeksag is okay with it I don't mind.

scottilee:
@berkerpeksag You also mentioned that the feedparser dependency isn't compatible with Python 3.7 but it looks like that was fixed here https://github.com/kurtmckee/feedparser/pull/131, thus allowing https://github.com/python/pythondotorg/issues/1267 to be closed as well?

docs/source/install.rst
eamanu: Yes, that I mean :-)

eamanu (approved):

scottilee:
Per #1441 I am removing the Vagrantfile and mentions of it in the documentation.

@berkerpeksag can I get another review please?

Also can the merge of this pull request close https://github.com/python/pythondotorg/issues/1270?

brainwane:
@berkerpeksag If you have time to review this I'd appreciate it! I'm used to using Docker and would love to be able to do that instead of having to set up Vagrant. Thanks.

jezdez:
I would strongly suggest to not use the Alpine image but the new (default) Buster based Python image. Alpine has had some [unpleasant side effects](https://pythonspeed.com/articles/base-image-python-docker-images/) in the past that are simply not worth the small size savings it provides.

Issue #1441, opened 2019-05-27
Remove Vagrant installation option


scottilee:
**Is your feature request related to a problem? Please describe.**
The [docs](https://pythondotorg.readthedocs.io/install.html) specify installation of Vagrant 2.0+ although the `Vagrantfile` lists 1.7.0 as the minimum version

**Describe the solution you'd like**
The line `Vagrant.require_version ">= 1.7.0"` in the `Vagrantfile` should be changed to a more recent version like 2.2.4.

**Describe alternatives you've considered**
Although the `Vagrantfile` works as-is, for security and bug fixes this change should be considered.

berkerpeksag:
I prefer removing the Vagrant setup once we merge #1440. Docker seems to be way more popular than our current choice of tooling and it would be better if we maintain one solution for setting up the development environment.

berkerpeksag:
I've just retitled the issue.

scottilee:
I'll include this with #1440.

PR #1445, opened 2019-06-02
Bump Vagrantfile above 2.0

scottilee:
Per #1441 the Vagrantfile should be bumped to 2.0+. 2.1.2 was released in the last year so I think it's a suitable minimum version.


berkerpeksag (commented):
Thanks for the PR! Closing this as per https://github.com/python/pythondotorg/issues/1441#issuecomment-498023370.

Please submit another PR after #1440 is merged.

PR #1446, opened 2019-06-02
Fix scroll to top and toggle menu

scottilee:
Fixes issue #1278: "Back to top not working properly"

Fixes two scroll to top buttons on smaller screen sizes that now scroll to the top of the screen.

Additionally, this fixes a menu close button at the top that now closes the menu.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -165,7 +165,7 @@
                     <span aria-hidden="true" class="icon-arrow-down"><span>&#9660;</span></span> Close / Open
                 </a>
 
-                <div id="sitetree_menu_trunk">
+                <div id="sitetree_menu_trunk" style="display:none;">
                     {% sitetree_menu from "main" include "trunk" template "sitetree/top.html" %}                    
                 </div>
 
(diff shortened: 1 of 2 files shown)


templates/base.html
```diff
@@ -162,10 +162,12 @@
 
                 {# Fallback for browsers that do not display generated content... HTML entity will be the icon instead #}
                 <a id="close-python-network" class="jump-link" href="#python-network" aria-hidden="true">
-                    <span aria-hidden="true" class="icon-arrow-down"><span>&#9660;</span></span> Close
+                    <span aria-hidden="true" class="icon-arrow-down"><span>&#9660;</span></span> Close / Open
```
berkerpeksag, line 165: Can we make the default state of the network menu closed instead? Personally, I rarely use them and its current implementation is way more buggy. I think it would be much better if we show it to users on demand.

scottilee:
Fixed the menu to be closed by default.

eamanu (approved):
LGTM

PR #1455, opened 2019-06-13
How about add support of any size for favicon?

Skriptach:


JacobCoffee:
I'm not sure about this, and judging by the age I don't know that past maintainers did either... but please feel free to reopen if you feel strongly about it!

Issue #1472, opened 2019-07-10
Add 3.7 version to CI


endormi:
**Describe the solution you'd like**
Add 3.7 to run CI

berkerpeksag:
This is basically a duplicate of #1267.

berkerpeksag:
(And thank you for the report!)

endormi:
Oops, sorry and no problem

PR #1473, opened 2019-07-10
Fix-1472: Add multiversion of python to travis

eamanu:
Add distribution xenial and add python3.7

Send a proposal for #1472

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
+dist: xenial
 language: python
-python: 3.6
+python:
+  - "3.6"
+  - "3.7"
 cache: pip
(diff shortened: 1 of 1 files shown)


berkerpeksag (changes requested):
1. No need to run tests against multiple Python versions. We only have one deployment, so no need to test for different Python 3 versions.

2. #1267 needs to be closed before we start to think about upgrading to Python 3.7.

I'm closing this for now since we are blocked by a new feedparser release.

Thanks!

eamanu:
Sorry for the noise! I will look for #1267 

Thanks

PR #1484, opened 2019-07-18
bpo-37613: Fix "Back to top" on Chrome for Android

potomak:
Update the selector on which to call the `animate` function to fix "Back to top" links on Chrome for Android.

Example session using the remote debugger:

back-to-top


potomak:
Ops, I just found #1446 that solves the same issue, feel free to close this PR.

PR #1508, opened 2019-09-06
clarify Contact Email on jobs submission form

ewdurbin:
Associates submissions with a user for contact information and add help text to clarify that the Contact Email field will be displayed publicly.

Previously this field was prepopulated with the user's email address, that is disabled to ensure they enter something and hopefully read the help text.

Screen Shot 2019-09-06 at 4 50 12 PM


jobs/views.py
```diff
@@ -361,7 +361,7 @@ def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
         kwargs['request'] = self.request
         # We don't allow posting a job without logging in to the site.
-        kwargs['initial'] = {'email': self.request.user.email}
+        kwargs['initial'] = {}
```
berkerpeksag: The following lines can be deleted:

```py
# We don't allow posting a job without logging in to the site.
kwargs['initial'] = {}
```

bwaliszewski: Can the text be more emphasized?

eamanu (approved):

PR #1509, opened 2019-09-11
Remove pubkeys.txt static file.

ned-deily:
As a convenience, the file contained the public keys of all
release managers but proved to be confusing and a potential
security risk.  The webpage that referenced it has been
updated to remove references to it and encourage downloading
public keys from trusted public keyservers or keybase.io.


ned-deily:
BTW, independently of merging this PR, please ensure that the current `pubkeys.txt` file is removed from the running web server instance.

ewdurbin:
as we're running from Heroku, files will not be present on the server post deploy. *however* we will need to remember to purge the cache once this deploys.

PR #1517, opened 2019-10-20
Improve highlighting of code samples

timhoffm:
- Add highlighting for the prompt marker in the code samples on the main page. This increases readability of the examples. The chosen color is the same as the one in the docs. It works quite well also on dark background:
image

- Refactor the html formatting so that the html tags are automatically generated and don't have to be written manually in the code samples sources.

  `format_html()` is actually a poor man's pygments, but I did not want to introduce this dependency. However, switching to pygments would be straight forward, if one wants to have full-fledged highlighting at some point in the future.

### DIFF
--- a/codesamples/factories.py
+++ b/codesamples/factories.py
@@ -1,3 +1,4 @@
+import re
 import textwrap
 
 import factory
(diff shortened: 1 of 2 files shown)


JacobCoffee:
Closing in favor of #2538 due to inability to push to fork. Commit credits will remain. 

Thanks for your contribution all those years ago :)

PR #1556, opened 2020-02-12
Allow modifying pep images with same name

### LINKED ISSUE #824: Edits to an image don't show up on the PEP page
I made two changes to the PNG image file:

```
$ hg log pep-0495-fold.png
changeset:   6056:e6c82824b10f
user:        Alexander Belopolsky <alexander.belopolsky@gmail.com>
date:        Sun Sep 20 20:25:12 2015 -0400
summary:     PEP 495: Added a gap sketch.

changeset:   6024:8c8352ff4668
user:        Alexander Belopolsky <alexander.belopolsky@gmail.com>
date:        Sun Aug 30 22:49:58 2015 -0400
summary:     PEP 495: Added a vertical segment to the graph.

changeset:   6022:db95b4037a82
user:        Alexander Belopolsky <alexander.belopolsky@gmail.com>
date:        Sun Aug 30 22:27:51 2015 -0400
summary:     PEP 495: Added png rasterization of the fold sketch.
```

but the [PEP](https://www.python.org/dev/peps/pep-0495) page still shows the old image.

lukpueh:
Fixes #824

Currently, `peps.converters.add_pep_image` does not allow to add an image that already exists in the database for the same name. As a consequence images cannot be updated.

This PR modifies `peps.converters.add_pep_image` to also update the image db record if the corresponding file has changed as per `filecmp.cmp'` This requires removing the file prior to the db update, because the django `FileField` does not override existing files.

Alternatives to the ad hoc removal of the file in `add_pep_image` include passing a custom `OverwriteStorage` to the used `ImageField`, see e.g.:

https://code.djangoproject.com/ticket/4339
https://stackoverflow.com/questions/9522759/imagefield-overwrite-image-file-with-same-name

### DIFF
--- a/peps/tests/test_converters.py
+++ b/peps/tests/test_converters.py
@@ -1,8 +1,13 @@
+import os
+import filecmp
+import shutil
+
 from django.test import TestCase, override_settings
 from django.core.exceptions import ImproperlyConfigured
 from django.test.utils import captured_stdout
+from django.conf import settings
 
-from peps.converters import get_pep0_page, get_pep_page, add_pep_image
+from peps.converters import get_pep0_page, get_pep_page, add_pep_image, pep_url
 
 from . import FAKE_PEP_REPO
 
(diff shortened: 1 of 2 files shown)


lukpueh:
I just skimmed through the peps git history to see how other image updates were propagated in the past. It generally doesn't seem to have happened often. 

One case that I found is [`pep-0602-example-release-calendar.png`](https://github.com/python/peps/blame/master/pep-0602-example-release-calendar.png). Interestingly,  the html `img` tag on the [pep site](https://www.python.org/dev/peps/pep-0602/) does point to the latest version of the image, i.e. [`pep-0602-example-release-calendar_cTnYawu.png`](https://s3.dualstack.us-east-2.amazonaws.com/pythondotorg-assets/media/dev/peps/pep-0602/pep-0602-example-release-calendar_cTnYawu.png). @ambv, @brettcannon, do you know how this is possible? Did I miss something in the peps conversion script? Or did you patch the html or db manually on the production server? Thanks!

brettcannon:
@lukpueh I'm not aware of any special thing done for that PEP.

ncoghlan:
There was at least one point where we accidentally broke PEP image publication in general, so I was wondering if it could be that the earlier versions of the image didn't make it into the website DB.

However, I don't think the image commit history backs that up.

peps/tests/test_converters.py
```diff
@@ -43,6 +48,41 @@ def test_add_image_not_found(self):
...
+        pep = '3001'
+        pep_img_name = 'pep-3001-1.png'
+        pep_img_path_src = os.path.join(FAKE_PEP_REPO, pep_img_name)
+        pep_img_path_src_backup = pep_img_path_src + ".backup"
+        pep_img_path_dst = os.path.join(
+                settings.MEDIA_ROOT, pep_url(pep), pep_img_name)
+
+        # We need to remove a prior uploaded file to fix an inconsistency
+        # between db and filesystem.
+        # TODO: This should be dealt with in page.model or converters module
```
lukpueh, line 62: Here's some additional info for this line;

A [prior test that calls `add_pep_image`](https://github.com/python/pythondotorg/blob/bc91a39d97a54943205d901c80c8a86b5bc30dde/peps/tests/test_commands.py#L36) might have already created the 'pep-3001-1.png' file, without persisting the db object. Now a new call to `add_pep_image` will [check for objects](https://github.com/python/pythondotorg/blob/bc91a39d97a54943205d901c80c8a86b5bc30dde/peps/converters.py#L217), whose [name ends with 'pep-3001-1.png'](https://github.com/python/pythondotorg/blob/bc91a39d97a54943205d901c80c8a86b5bc30dde/peps/converters.py#L221), and, since it does not find any, create it anew along with the file. But since Django does not allow to override files it duplicates it using a random infix, e.g 'pep-3001-1_0u7tlLY.png', which is also used as a name on the object.

Now due to the `endswith`-check and with the random infix, each subsequent attempt to add 'pep-3001-1.png' with `add_pep_image`, will also duplicate it the object and file.

Maybe we should just override images on upload as suggested as alternative fix in the PR description?

lukpueh (commented):
I just pushed a test for my patch.  While writing it I found another little issue (see comment inline), which suggests to use my alternative solution (override on upload).  What are others' thoughts? 

It would be nice to fix this.  [Pep 458](https://www.python.org/dev/peps/pep-0458/) still references an [outdated file](https://s3.dualstack.us-east-2.amazonaws.com/pythondotorg-assets/media/dev/peps/pep-0458/pep-0458-1.png), although the repo [has the correct one](https://github.com/python/peps/blob/master/pep-0458-1.png). 

Thanks

lastthyme (approved):

hugovk:
This can be closed: after implementing [PEP 676](https://peps.python.org/pep-0676/), PEP pages are now deployed directly from https://github.com/python/peps to https://peps.python.org/.

It does however highlight there's some old PEP-handling code in the repo which is no longer needed and can be removed.

hugovk:
@lukpueh I don't have permission to close things in this repo. Please could you close this PR?

PR #1565, opened 2020-03-08
GH-1564: Add "Last Modified" trailer to PEP HTML pages

### LINKED ISSUE #1564: Add Last-modified: date and metadata to HTML view of PEPs
When looking at a PEP page on `python.org`, it isn't obvious when the PEP was last modified, or how to see recent revisions.

In addition, search engines often aren't displaying useful information. E.g. the Google result for a search of PEPs seems to show the `Created:` date, which may confuse searchers. E.g. the date listed for [PEP 518](https://www.python.org/dev/peps/pep-0518/) in a Google search is `May 10, 2016`, though this Provisional PEP was most recently updated on `Dec 13, 2019`.

It would be great if the footer of the HTML rendering of each PEP, when served by `python.org`, included a `Last-Modified:` date, with a link to the GitHub history for more information.  It could go after the Source link at the bottom of the page. E.g.

 * Source: https://github.com/python/peps/blob/master/pep-0518.txt
 * Last-Modified: [Dec 13, 2019](https://github.com/python/peps/commits/master/pep-0518.txt)

Including relevant dates as metadata such as `datePublished` and `dateModified` should help search engines display more helpful information, as described at [Official Google Webmaster Central Blog: Help Google Search know the best date for your web page](https://webmasters.googleblog.com/2019/03/help-google-search-know-best-date-for.html)

Our documentation at [PEP Page Generation — Python\.org Website 1\.0 documentation](https://pythondotorg.readthedocs.io/pep_generation.html) describes the current process.

Given the `peps` git repo obtained via that process, it seems to take less than 20 seconds to get the latest modification date of all the PEPs via e.g. even using the shell pipeline listed at [How to retrieve the last modification date of all files in a git repository \- Server Fault](https://serverfault.com/a/774574/64546), so I don't think there would be much overhead to maintaining this information.

A discussion of some of the options for making PEP history clearer, or whether version numbers or changelogs for PEPs would be worthwhile, is at [Easier tracking of PEP changes, versions, or modification dates \- PEPs \- Discussions on Python\.org](https://discuss.python.org/t/easier-tracking-of-pep-changes-versions-or-modification-dates/3658/3). The approach described here expands on what I see as the result of that discussion.

Adding `Last-Modified` information to the [PEP 0](http://python.org/dev/peps) page might also be worthwhile, as suggested there.

Mariatta:
Obtain the last modified date by making GET request to GitHub.
Don't render anything if there's any error with the GET request.

Add tests, and mock the network call.

API call: https://api.github.com/repos/python/peps/commits?path=pep-0012.txt
Display the  date from `["commit"]["committer"]["date"]`

The url is the GitHub history of the pep file, e.g. https://github.com/python/peps/commits/master/pep-0012.txt

Closes https://github.com/python/pythondotorg/issues/1564

Screen Shot 2020-03-07 at 6 59 25 PM

### DIFF
--- a/peps/converters.py
+++ b/peps/converters.py
@@ -4,6 +4,7 @@ import re
 import os
 
 from bs4 import BeautifulSoup
+import requests
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
@@ -181,9 +182,11 @@ def get_pep_page(artifact_path, pep_number, commit=True):
         artifact_path, 'pep-{}.rst'.format(pep_number),
     )
     pep_ext = '.rst' if os.path.exists(pep_rst_source) else '.txt'
-    source_link = 'https://github.com/python/peps/blob/master/pep-{}{}'.format(
-        pep_number, pep_ext)
-    pep_content['content'] += """Source: <a href="{0}">{0}</a>""".format(source_link)
+    pep_filename = 'pep-{}{}'.format(pep_number, pep_ext)
+    source_link = 'https://github.com/python/peps/blob/master/{}'.format(pep_filename)
+    pep_footer = """<div>Source: <a href="{0}">{0}</a></div>""".format(source_link)
+    pep_footer += get_commit_history_info(pep_filename)
+    pep_content['content'] += pep_footer
 
     pep_page, _ = Page.objects.get_or_create(path=pep_url(pep_number))
 
@@ -197,6 +200,28 @@ def get_pep_page(artifact_path, pep_number, commit=True):
         pep_page.save()
 
     return pep_page
+
+
+def get_commit_history_info(pep_filename):
+    commit_info = ''
+
+    try:
+        data = requests.get(
+            'https://api.github.com/repos/python/peps/commits?path={}'.format(pep_filename)
+        )
+    except Exception:
+        return commit_info
+
+    json_data = data.json()
+    if len(json_data) > 0:
+        commit_date = json_data[0]['commit']['committer']['date']
+        commit_url = 'https://github.com/python/peps/commits/master/{}'.format(
+            pep_filename
+        )
+        commit_info = """<div>Last modified: <a href="{0}">{1}</a></div>""".format(
+            commit_url, commit_date
+        )
+    return commit_info
 
 
 def add_pep_image(artifact_path, pep_number, path):
--- a/peps/tests/test_commands.py
+++ b/peps/tests/test_commands.py
@@ -9,6 +9,8 @@ from django.core.management import call_command
 
 import responses
 
+from unittest.mock import patch
+
 from pages.models import Image
 
 from . import FAKE_PEP_ARTIFACT
@@ -32,7 +34,11 @@ class PEPManagementCommandTests(TestCase):
         )
 
     @responses.activate
-    def test_generate_pep_pages_real_with_remote_artifact(self):
+    @patch('peps.converters.get_commit_history_info')
+    def test_generate_pep_pages_real_with_remote_artifact(
+        self, mock_get_commit_history
+    ):
+        mock_get_commit_history.return_value = ''
         call_command('generate_pep_pages')
 
     @override_settings(PEP_ARTIFACT_URL=FAKE_PEP_ARTIFACT)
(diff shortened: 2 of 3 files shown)


Mariatta:
Instead of getting the last commit date through `git` as suggested by @nealmcb  in https://github.com/python/pythondotorg/issues/1564, I chose to retrieve the info using GitHub API.
I find it easier and cleaner than having to make a `git` subprocess call (which returns text that we have to parse). Additionally, we already have `requests` library as a dependency of this repo, and the API call does not require any authorization. So we don't have to store GitHub API token. (but it is probably better to do so).

@berkerpeksag please review when you have the chance.

peps/converters.py
```diff
@@ -199,6 +202,28 @@ def get_pep_page(artifact_path, pep_number, commit=True):
...
+    try:
+        data = requests.get(
+            'https://api.github.com/repos/python/peps/commits?path={}'.format(pep_filename)
+        )
+    except Exception:
+        return commit_info
+
+    json_data = data.json()
+    if len(json_data) > 0:
+        commit_date = json_data[0]['commit']['committer']['date']
```
nealmcb, line 217: Are there any failure conditions in which this would generate a `KeyError`?

nealmcb:
Wow - fast work, @Mariatta!

The issue also discusses the benefits of adding metadata such as `datePublished` and `dateModified` so that search engines can display more useful information to users.
If we plan to add that also, then looking for multiple data points per PEP might affect the tradeoff between doing local git operations and using web requests. The former also presumably would make error handling easier, and generation of the full footer more robust.

Mariatta:
@berkerpeksag will you have time to look into this and review, and perhaps help merge?

@nealmcb regarding additional metadata like datePublished or dateModified, I think those will come from the content of PEP file, which we already have access to. It will not come from `git` so I think the current solution will work.

Sorry for the delay in following up on this.

berkerpeksag:
> @berkerpeksag will you have time to look into this and review, and perhaps help merge?

Yes, sorry for my late response. Will try to have a look this weekend.

merwok:
Thanks for working on this Mariatta!  Better links to git repo and history page can help people find more information and address the criticism/confusion in the original mail thread (or forum, I forgot).

It doesn’t feel great that a local build process, working on local files and with all info in a local git repo, now needs to send many external requests to one repo host.

Using git plumbing commands in subprocess wouldn’t be so bad?  Or maybe one of the Python git libs could be nicer to use?  (I could investigate that next week)

peps/converters.py
aeros: As far as I'm aware, github's API will always return `[]` upon finding no results or error, and when results are found for the repository's commits, there will always be an accessible `commit.committer.date` field. So, the `KeyError` would not occur in any situation (that I'm aware of) where `data.json()` returns something with a length > 0 when targeted at that specific URL.

peps/converters.py
```diff
@@ -199,6 +202,28 @@ def get_pep_page(artifact_path, pep_number, commit=True):
...
 
+def get_commit_history_info(pep_filename):
+    commit_info = ''
+
+    try:
+        data = requests.get(
+            'https://api.github.com/repos/python/peps/commits?path={}'.format(pep_filename)
+        )
+    except Exception:
+        return commit_info
```
aeros, line 213:
```suggestion
    try:
        data = requests.get(
            'https://api.github.com/repos/python/peps/commits?path={}'.format(pep_filename)
        )
        data.raise_for_status()
    except Exception:
        return commit_info
```

It might be worthwhile to add `data.raise_for_status()`, to ensure that we always raise an exception here if we get an http error code like 404, 401, etc. Otherwise, it might raise a `JSONDecodeError` when we call `data.json()`. Realistically, I don't expect this will happen anytime soon to the github API under normal circumstances, but I think it would be a good general best practice to avoid making the PEP pages dependent on an external API returning a parse-able JSON response every time.

In general, I think it would also be better to catch `requests.exceptions.RequestException` instead of `Exception`, but I don't think it's an overly substantial concern in this case.

berkerpeksag:
> It doesn’t feel great that a local build process, working on local files and with all info in a local git repo, now needs to send many external requests to one repo host.

That would be nice, but current solution looks good to me as well.

Perhaps we could pass `per_page=1` to avoid fetching and parsing unnecessary data.

I'll try to get this merge this weekend.

hugovk:
The PEPs now have their own repo: https://github.com/python/peps and own website: https://peps.python.org/ and also have "Last modified" footers that link to GitHub. For example: https://peps.python.org/pep-0750/

@Mariatta Please could you close this PR?

Issue #1302, opened 2018-07-31
It would be great if we can clarify which Python versions are still maintained ...


Mariatta:
... and which ones are not.

**Is your feature request related to a problem? Please describe.**
Folks not involved with Core Python development might not know which Python versions are still maintained, and which ones are no longer maintained and not preferred anymore.

We have a release schedule listed at the bottom of python.org/downloads, I guess it was meant to indicate which ones will still receive new fixes, but it was not explicit in saying that the other ones aren't being maintained anymore.

Devguide has [some more info](https://devguide.python.org/#status-of-python-branches) about it, but casual users wouldn't land on that page.

**Describe the solution you'd like**
It would be great if we can clarify which Python versions are still maintained, and which are not.
The downloads page seems like a good place. Perhaps we can add some paragraph indicating which ones are still actively maintained, and which ones are not.
We also should state that unmaintained Python versions should not be used.

**Additional context**
See twitter thread: https://twitter.com/pydanny/status/1024302657161256960

jefftriplett:
Would it make send to update the release grid from the [downloads page](https://www.python.org/downloads/) to show whether or not a release is supported? 

Here is a screenshot: ![](https://www.evernote.com/shard/s27/sh/7c552539-f1d3-49e5-8d98-324847a8b016/ac929c11c083b5e9/res/b035d42f-ae6b-4fd0-95a6-a50e9d01ecf9/skitch.png) This could be an extra column/icon.

jaap3:
What about doing something similar to [Django's download page](https://www.djangoproject.com/download/)?

They make a clear distinction between 

* _Latest releases_ 
    final and RC
* _Previous releases_
    Currently only the LTS version
* _Unsupported previous releases_
   With a clear indication that these releases "no longer receive security updates or bug fixes"

Furthermore they do an excellent job of communicating the timeline of supported versions.

anlutro:
Was directed here from bugs.python.org, want to re-iterate my points because I think this is important. Every time I see people asking questions about this in IRC I have a hard time finding the official list.

There are many pretty version support pages out there but I don't think that's the most important thing. Try searching for the following in google:

- `$THING` versions support
- `$THING` versions lifecycle
- `$THING` versions roadmap
- `$THING` versions end of life

[...]

PR #1569, opened 2020-03-27
Add Active Python Releases to Downloads Page

crwilcox:
#1302 raised that it would be nice to have information, similar to other languages, that clearly states which python versions are actively maintained.

image

### DIFF
--- a/templates/downloads/index.html
+++ b/templates/downloads/index.html
@@ -118,8 +118,6 @@
                 </div>
 
 
-
-
         		<div class="row">
 
         		    <div class="small-widget download-widget1">
--- a/templates/downloads/index.html
+++ b/templates/downloads/index.html
@@ -92,8 +92,6 @@
 
                 <div class="row download-list-widget">
 
-                    {% render_active_banner %}
-
                     <h2 class="widget-title">Looking for a specific release?</h2>
                     <p class="success-quote">Python releases by version number:</p>
 
--- a/templates/downloads/index.html
+++ b/templates/downloads/index.html
@@ -44,11 +44,11 @@
                     <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information please visit the Python Developer's Guide</a>.</p>
 
                     <div class="list-row-headings">
-                        <span class="release-version">Python Version</span>
-                        <span class="release-status">Maintenance Status</span>
-                        <span class="release-start">First Released</span>
-                        <span class="release-end">End of Support</span>
-                        <span class="release-pep">Release Schedule</span>
+                        <span class="release-version">Python version</span>
+                        <span class="release-status">Maintenance status</span>
+                        <span class="release-start">First released</span>
+                        <span class="release-end">End of support</span>
+                        <span class="release-pep">Release schedule</span>
                     </div>
                     <ol class="list-row-container menu">
                         <li>
--- a/templates/downloads/index.html
+++ b/templates/downloads/index.html
@@ -41,7 +41,7 @@
                     {% render_active_banner %}
 
                     <h2 class="widget-title">Active Python Releases</h2>
-                    <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information please visit the Python Developer's Guide</a>.</p>
+                    <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information visit the Python Developer's Guide</a>.</p>
 
                     <div class="list-row-headings">
                         <span class="release-version">Python version</span>
--- a/templates/downloads/index.html
+++ b/templates/downloads/index.html
@@ -37,6 +37,59 @@
 {% endblock header_content %}
 
 {% block content %}
+                <div class="row active-release-list-widget">
+                    {% render_active_banner %}
+
+                    <h2 class="widget-title">Active Python Releases</h2>
+                    <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information please visit the Python Developer's Guide</a></p>
+
+                    <div class="list-row-headings">
+                        <span class="release-version">Python Version</span>
+                        <span class="release-status">Maintenance Status</span>
+                        <span class="release-start">First Released</span>
+                        <span class="release-end">End of Support</span>
+                        <span class="release-pep">Release Schedule</span>
+                    </div>
+                    <ol class="list-row-container menu">
+                        <li>
+                            <span class="release-version">3.8</span>
+                            <span class="release-status">bugfix</span>
+                            <span class="release-start">2019-10-14</span>
+                            <span class="release-end">2024-10</span>
+                            <span class="release-pep"><a href="https://www.python.org/dev/peps/pep-0569">PEP 569</a></span>
+                        </li>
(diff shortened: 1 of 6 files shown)


Mariatta (approved):
The table looks good to me. Thank you! I think this will help the community greatly.

Mariatta:
@berkerpeksag will you be able to review and also deploy this to prod? Or if you can tell me how I can deploy to production, then I'll be able to help.

ewdurbin:
Thank you @crwilcox for the implementation and @Mariatta for the idea and review!

My only concern here is with populating the table staticly in the template will require someone to update the template as new releases of python come into existence.

The @python/python-release-managers would probably be best suited to maintain this information as they already interact with the [`Release`](https://github.com/python/pythondotorg/blob/6902c19434763603f405061c31fb660454b1dbcd/downloads/models.py#L39) objects when performing their duties.

I'm not positive if it's worth blocking this on these concerns, but adding a new `Version` model to replace the very rough [`versions`](https://github.com/python/pythondotorg/blob/6902c19434763603f405061c31fb660454b1dbcd/downloads/models.py#L44-L52) field on `Release` would make this far more maintainable as part of a RM's role, and allow the list to be populated from the database rather than being hard coded in the template.

I'll leave some time for @python/python-release-managers to chime in with any thoughts, but otherwise this can be merged and the above ramblings converted into an issue :)

crwilcox:
Thanks for taking a look @ewdurbin! I raised a similar concern on the issue. I think linking this to releases would be great. I wasn't quite sure the approach that would be most desired by release managers so I sort of stopped myself 😅

The coupling today of releases and updating PEPs/Dev Guide isn't super automated (as best I know anyway), but it all seems to work out. That said, I would be happy to take on getting a model setup for this, though as you said it might be easier to do that as a separate PR.

templates/downloads/index.html
```diff
@@ -37,6 +37,59 @@ <h1 class="call-to-action">Download the latest version of Python</h1>
 {% endblock header_content %}
 
 {% block content %}
+                <div class="row active-release-list-widget">
+                    {% render_active_banner %}
```
berkerpeksag, line 41: Is it possible to move this inside `<div class="row download-list-widget">`? If not, we need to remove the duplicate `render_active_banner` call in line 95.

templates/downloads/index.html
```diff
@@ -37,6 +37,59 @@ <h1 class="call-to-action">Download the latest version of Python</h1>
...
+
+                    <h2 class="widget-title">Active Python Releases</h2>
+                    <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information please visit the Python Developer's Guide</a></p>
```
berkerpeksag: Nitpick:

```suggestion
                    <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information please visit the Python Developer's Guide</a>.</p>
```

templates/downloads/index.html
```diff
@@ -65,6 +118,8 @@ <h2 class="widget-title">Looking for a specific release?</h2>
                 </div>
 
 
+
+
```
berkerpeksag: Nitpick: Please delete the superfluous newlines.

```suggestion
```

berkerpeksag (commented):
Thank you for the PR! In general, I agree with @ewdurbin that it would be better to find a more dynamic solution, but I think this is a good start for linking release schedule PEPs on www.python.org. If RMs will be happy about getting this in, it would be nice to update PEP 101 and include it as part of release process.

I left some comments about HTML and duplicate template tag.

templates/downloads/index.html
berkerpeksag: Actually, you can ignore my comment about moving it inside `<div class="row download-list-widget">`. It looks pretty great! :) We just need to remove the duplicate `render_active_banner` call in line 95.

ned-deily:
Thanks for the PR.  On behalf of the release managers, who maintain this information, I'd like to take a closer review of this over the next few days.

crwilcox:
Thanks for the feedback @berkerpeksag. I think I have addressed all of it. @ned-deily have you had a chance to think about this? Are there other release folks worthing having take a look?

templates/downloads/index.html
```diff
@@ -37,10 +37,61 @@ <h1 class="call-to-action">Download the latest version of Python</h1>
...
 
+                    <h2 class="widget-title">Active Python Releases</h2>
+                    <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information please visit the Python Developer's Guide</a>.</p>
+
+                    <div class="list-row-headings">
+                        <span class="release-version">Python Version</span>
+                        <span class="release-status">Maintenance Status</span>
+                        <span class="release-start">First Released</span>
+                        <span class="release-end">End of Support</span>
+                        <span class="release-pep">Release Schedule</span>
```
hugovk: Use sentence case for consistency with other table style:

```suggestion
                        <span class="release-version">Python version</span>
                        <span class="release-status">Maintenance status</span>
                        <span class="release-start">First released</span>
                        <span class="release-end">End of support</span>
                        <span class="release-pep">Release schedule</span>
```

templates/downloads/index.html
```diff
@@ -37,10 +37,61 @@ <h1 class="call-to-action">Download the latest version of Python</h1>
 {% endblock header_content %}
 
 {% block content %}
-                <div class="row download-list-widget">
-
+                <div class="row active-release-list-widget">
                     {% render_active_banner %}
 
+                    <h2 class="widget-title">Active Python Releases</h2>
+                    <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information please visit the Python Developer's Guide</a>.</p>
```
hugovk: Remove "please" for brevity, cf. https://developers.google.com/style/tone#politeness-and-use-of-please

```suggestion
                    <p class="success-quote"><a href="https://devguide.python.org/#status-of-python-branches">For more information visit the Python Developer's Guide</a>.</p>
```

ned-deily (approved):
Sorry about the review delay!  I think it's a big improvement to the web site.  Thanks for the PR.  Obviously it's not ideal to have the release data hardwired into the templates but it will do for now.  Longer term, I'd like to find a way to extract the release information we have from the database and make it available as, say, a json file on the website that could then be used by multiple consumers including this web page.  But that's not a good reason to hold this up.  I will add a note to PEP 101, the Python Release Process document to remind release managers to update this data on the website at appropriate points in our release cycle.  Thanks again, everyone!

berkerpeksag (approved):
LGTM!

ewdurbin:
Thanks @crwilcox! This is now live.

ned-deily:
You know how you look at something too long and you overlook the obvious??  I've opened #1577 to change 2.7 status from bugfix to end-of-life (and then remove in the future).

berkerpeksag:
Oops, I completely missed it when I looked at it before merging :)

ned-deily:
We all did!  I guess it's still a bit hard to believe that 2.7 is gone :)

crwilcox:
Thanks to everyone for helping get this in.

PR #1572, opened 2020-04-06
Change header donate button link

indepndnt:


Mariatta:
I think the existing url is correct and should not be changed.
Thanks.

ewdurbin:
Hi @Mariatta, @indepndnt is a member of the Accounting team for PSF. We're moving away from the PayPal donation button because it doesn't allow us to collect all the necessary information needed to properly record and acknowledge donations.

PR #1576, opened 2020-04-20
Update table: Python 2.7 is EOL

hugovk:
List it as EOL (or another status?) for a bit longer (how long?) or remove it now?

### DIFF
--- a/templates/downloads/index.html
+++ b/templates/downloads/index.html
@@ -81,7 +81,7 @@
                         </li>
                         <li>
                             <span class="release-version">2.7</span>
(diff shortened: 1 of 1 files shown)


ned-deily:
Sorry, @hugovk, I didn't notice your PR when I opened #1577.  I would prefer to use "end-of-life" as in #1577 as that is the "official" terminology we use in the devguide and, hopefully, elsewhere.  I'm going to close this in favor of #1577.

hugovk:
All good, thank you!

PR #1613, opened 2020-07-14
switch to github flavored markdown for CMS pages

ewdurbin:
More and more becoming defacto which leads to confused users... rather than battle with extensions and the Markdown python module... just use `cmarkgfm`.

### DIFF
--- a/pages/models.py
+++ b/pages/models.py
@@ -9,6 +9,8 @@ positioned into the URL structure using the nav app.
 import os
 import re
 
+from copy import deepcopy
+
 from django.conf import settings
 from django.core import validators
 from django.db import models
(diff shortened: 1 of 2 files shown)


Mariatta (approved):
Makes sense for this change. Thanks.

ewdurbin:
Thank you for the review @Mariatta!

pages/models.py
```diff
@@ -43,13 +48,24 @@
...
+RENDERERS = deepcopy(DEFAULT_MARKUP_TYPES)
+for i, renderer in enumerate(RENDERERS):
+    if renderer[0] == 'markdown':
+        markdown_index = i
+
+RENDERERS[markdown_index] = (
+    'markdown',
+    cmarkgfm.github_flavored_markdown_to_html,
+    'Markdown'
+)
```
di, line 60:
```suggestion
RENDERERS = [
    x, cmarkgfm.github_flavored_markdown_to_html if x == 'markdown' else xx, xxx,
    for x, xx, xxx in deepcopy(DEFAULT_MARKUP_TYPES)
]
```

I don't know what to name these but maybe this would be more clear? idk

di (approved):
Approved regardless

PR #1648, opened 2020-10-02
Update Release Status of Python 3.5 on downloads page

crwilcox:
EOL of Python 3.5 was announced yesterday. Updating this table to reflect that.

Also, could remove this from the table, but left it, as Python 2.7 has been left, to reflect its status.


Mariatta (approved):
Thanks!

PR #1649, opened 2020-10-02
Improvements on how to work with SCSS files

berinhard:
Hi everyone! As my previous PR #1647, this one also comes after my experiences when setting up the project. 

While doing that, I ran into an issue when with the `sass` command. The brief summary is:

1. It askef for another dependency that I had to install with `gem install --version '~> 0.8.8' rb-inotify`;
2. This command raised an error because I didn't have ruby development headers installed;
3. I had to run `sudo apt install ruby-dev` to fix that;
4. Only after that I was able to install the missing dependency and run `sass` as expected;

To help people on advance, I added a small note in the installing instruction so people don't forget to install ruby's dev headers. Also, now `rb-inotify` is listed in `Gemfile`.

I'm also proposing 2 other things with this PR with a new `Makefile`:

1. A `make sass` command to avoid people from remembering how to run `sass`. The [current command](https://github.com/python/pythondotorg/blob/master/docs/source/install.rst#generating-css-files-automatically) can be very complicated to remember and this shortcut can help with that;
2. A `make run` command to allow people to, if they want to, run all the development commands using a single terminal window;

Before this work, I talked briefly with @ewdurbin about moving from Ruby's gems to Node.js' npm to handle the `.scss` files. On second thought, I realized the project would still with external dependencies to perform the same task, but from different technologies, and it didn't sound like a huge benefit IMHO. It would be nice to have second opinions on this as well and, maybe, having a new issue to address this topic.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ group :media do
   gem "compass", "~>0.12.2"
   gem "sass", "~>3.2.5"
   gem "susy", "~>1.0.5"
+  gem "rb-inotify", "~>0.8.8"
 end
 
 group :development do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,9 +6,12 @@ GEM
       chunky_png (~> 1.2)
       fssm (>= 0.2.7)
       sass (~> 3.1)
+    ffi (1.13.1)
     foreman (0.61.0)
       thor (>= 0.13.6)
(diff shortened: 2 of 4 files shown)


JacobCoffee:
Should this be continued? @ewdurbin

ewdurbin:
I’m -1 on a ruby implementation, so probably not!

Issue #1611, opened 2020-07-07
Update Windows download popup for 3.9


zooba:
Getting this request in early: Python 3.9 will drop support for Windows 7, which still makes up about 20% of our downloads, and so we should warn users that it won't work.

Currently, this popup has a warning about Windows XP:
![image](https://user-images.githubusercontent.com/1693688/86855296-f62b5200-c0b1-11ea-9d00-01860c35161d.png)

We should update it to read "Note that Python 3.9+ cannot be used on Windows 7 or earlier"

I think we should also update the link to point at the 64-bit installer. All the best data I can get shows that 32-bit OS makes up less than 10% (less than 1%, depending on relevant filters), and I believe the issues we see with third parties not supporting 32-bit will now outweigh users attempting to install on 32-bit OS.

Alternatively, we can add a second button for the 32-bit installer and clearly label them both as "Python 3.9.x 64-bit" // "Python 3.9.x 32-bit" (adding @ambv in case this bit is under release manager control).

Also see #1592, regarding adding a link to the Store package (and #1370 from last time we did it). The link for 3.9 is https://www.microsoft.com/store/apps/9P7QFQMJRFP7

aasthashukla1:
Hey, Zooba!

I am new to GitHub open-source contribution community and would love to have this as my first contribution.

Do you believe it would be possible for me to do this?

ned-deily:
It's an easy matter to change which one of the release files per platform are associated with the existing download button.  The release manager can easily change existing releases; a minor change to one of the release scripts will update the default.  We can also easily update the warning text.  Adding a second download button or a hierarchy of download links would be much more involved, e.g. real code changes!  If you and @ambv are OK with those first two changes, I can make them.

ned-deily:
@aasthashukla1, thanks for the offer!  The python.org website infrastructure is spread out over GitHub repos and a content management system database.  These proposed changes affect the latter so can't be done via the usual Github PR mechanism.  Keep an eye out for other opportunities!

zooba:
Thanks Ned!

PR #1655, opened 2020-10-05
Update Windows download popup and list for 3.9. (#1611)

ned-deily:
As of Python 3.9, only Windows 7 and later systems are supported. With the release of 3.9.0, change the download popup to state that (since we assume the download popup button will never move back to 3.8.x or earlier) and change the Windows download list to show Windows 7 for 3.9 and later releases and continue to show Windows XP for 3.5 through 3.8..

### DIFF
--- a/templates/downloads/os_list.html
+++ b/templates/downloads/os_list.html
@@ -38,8 +38,12 @@
                     {% for r in releases %}
                     <li>
                         <a href="{{ r.get_absolute_url }}">{{ r.name }} - {{ r.release_date|date }}</a>
-                        {% if os.slug == 'windows' and r.name >= 'Python 3.5' %}
-                        <p><strong>Note that {{ r.name }} <em>cannot</em> be used on Windows XP or earlier.</strong></p>
+                        {% if os.slug == 'windows' %}
+                            {% if r.name >= 'Python 3.9' %}
+                            <p><strong>Note that {{ r.name }} <em>cannot</em> be used on Windows 7 or earlier.</strong></p>
+                            {% elif r.name >= 'Python 3.5' %}
+                            <p><strong>Note that {{ r.name }} <em>cannot</em> be used on Windows XP or earlier.</strong></p>
+                            {% endif %}
                         {% endif %}
                         <ul>
                             {% for f in r.files.all %}
--- a/templates/downloads/supernav.html
+++ b/templates/downloads/supernav.html
@@ -10,7 +10,7 @@
         <p>
             <a class="button" href="{{ data.python3.url }}">{{ data.python3.release.name }}</a>
         </p>
-        {% if data.os.slug == 'windows' %}<p><strong>Note that Python 3.5+ <em>cannot</em> be used on Windows XP or earlier.</strong></p>{% endif %}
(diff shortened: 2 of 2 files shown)


templates/downloads/os_list.html
```diff
@@ -38,8 +38,12 @@ <h2>Stable Releases</h2>
                     {% for r in releases %}
                     <li>
                         <a href="{{ r.get_absolute_url }}">{{ r.name }} - {{ r.release_date|date }}</a>
-                        {% if os.slug == 'windows' and r.name >= 'Python 3.5' %}
-                        <p><strong>Note that {{ r.name }} <em>cannot</em> be used on Windows XP or earlier.</strong></p>
+                        {% if os.slug == 'windows' %}
+                            {% if r.name >= 'Python 3.9' %}
```
berkerpeksag, line 42: Wouldn't this return `False` if `r.name` is `'Python 3.10'`?

ned-deily: Probably!  But that's just what the existing code does.  I was going to open a separate issue to go through the whole repo to fix all similar version checks prior to 3.10.0 a year from now.  Do you have any idea if there are other such version checks?

hugovk: Yep!
```pycon
>>> 'Python 3.10' >= 'Python 3.9'
False
```

A quick grep doesn't show any more of this pattern:

```console
$ git grep -E "[23]\..* [<>][=]? "
$ git grep -E " [<>][=]? .*[23]\."
static/sass/_layout.scss:        & > .column { width: 33.3333%; }
templates/downloads/os_list.html:                        {% if os.slug == 'windows' and r.name >= 'Python 3.5' %}
```

Running https://github.com/asottile/flake8-2020/ doesn't find anything:

```console
$ pip install -qU flake8-2020 && flake8 --select YTT
$
```

Still worth a more thorough review.

ned-deily: If anyone wants to come up with a proper version test, please feel free to do so! My immediate concern is the imminent 3.9.0 release.

hugovk: Agreed, this PR looks good for 3.9.0, and we've a year to fix it properly for 3.10.0.

---

Something like this?
```pycon
>>> from packaging import version  # pip install packaging
>>> def parse_version(version_string):
...     return version.parse(version_string.split()[1])
...
>>> parse_version("Python 3.10") >= parse_version("Python 3.9")
True
>>> parse_version("Python 3.10") <= parse_version("Python 3.9")
False
>>> parse_version("Python 3.10") == parse_version("Python 3.10")
True
>>>

Comment (ned-deily) [reply templates/downloads/os_list.html:42]: Oh, that looks like a good approach.  I forgot about packaging.

Comment (hugovk) [reply templates/downloads/os_list.html:42]: And rather than the `.split()[1]`, looks like we could use `r.get_version()`:

```python
def parse_version(version_string):
     return version.parse(version_string)
```

https://github.com/python/pythondotorg/blob/0f07e9daf14e9f45853b61b7e9724c3741870762/downloads/models.py#L121-L125

https://github.com/python/pythondotorg/blob/0f07e9daf14e9f45853b61b7e9724c3741870762/downloads/tests/test_models.py#L63-L66

ambv:
OK, including this is urgent for 3.9 to not mislead the users to think 3.9 works on Windows Vista and Windows 7. My vote is: let's merge the current state, unblocking 3.9.0, and fix forward for 3.10.

Issue #1656, opened 2020-10-05
Agenda, Reports, and Minuting feature


ewdurbin:
As requested by @brainwane, pulled from: https://mail.python.org/pipermail/pydotorg-www/2020-June/005432.html

Context:

Right now, www.python.org hosts the board meeting minutes 
https://www.python.org/psf/records/board/minutes/ . We talked about 
maybe also making space for working group/project meeting minutes, like 
the ones I write up and currently host on the wiki (example: 
https://wiki.python.org/psf/PackagingWG#Dependency_resolver_and_user_experience_improvements_for_pip 
).

In the call, Ernest asked me to list what would be important in a 
minuting package for www.python.org, distinguishing "required" from 
"heavily encouraged" to "dream features".

Current workflow:

Right now, in order to minute meetings of Packaging WG-funded projects, I:

* use an Etherpad at pad.sfconservancy.org (just because it's a public 
and reliable EtherPad instance) and take notes with bullet points
* after the meeting, use its Export function to export to plain text
* mess with the formatting to adjust to MoinMoin wiki syntax
* go to https://wiki.python.org/psf/PackagingWG and create a new 
placeholder link in the table of meeting notes, annotating with the 
description, type (meeting notes), and date
* copy the text into that page, fix formatting, and hit Save
* erase the meeting notes from the Etherpad and hyperlink to the 
archived notes

What I want:

For the process in general, I have 3 core values:
* Ease of writing: it should be easy for me and for other meeting 
attendees to collaborate on writing minutes, live, during the meeting, 
in a reasonably lightweight syntax (such as Markdown)
* Privacy: as we write the minutes, they should be private to meeting 
attendees, so we can discuss things we then redact before sharing 
(vacations, burnout, criticism of other people or projects)
* Transparency: the final minutes should be public where anyone can read 
them, without having to log in anywhere, and linkable

As I understand it, a minuting system on www.python.org would have a 
hierarchy like PSF -> Working Group -> Project.  Example: PSF -> 
Packaging WG -> Pip dependency resolver & UX improvements.

Required:

* Discoverability: the public minutes should be easy to find from a 
central project info hub, and show up in search engine results.
* Linkability: each meeting should have some unique URL or anchor tag, 
so that it's easy to link to minutes within an issue or mailing list post.
* Ease of formatting: some subset of HTML, Markdown, and 
reStructuredText should be supported.

Heavily encouraged:

* Automated table of contents on the Project page. Should include (by 
[...]

PR #1659, opened 2020-10-09
Automaticallly generates PSF board meetings minutes

berinhard:
This is still a work in progress related to the discussion in #1656.

So far I'm trying to validate the workflow to create a new PSF Board meeting, edit minutes and display them using the existing `minutes.models.Minutes` model.


minutes/admin.py
```diff
@@ -15,3 +18,41 @@ def get_list_filter(self, request):
...
+    extra = 1
+
+
+@admin.register(Concern)
+class ConcernAdmin(admin.ModelAdmin):
+    inlines = [ConcernRoleInline]
+
+
+@admin.register(ConcernedParty)
+class ConcernedPartyAdmin(admin.ModelAdmin):
```
ewdurbin, line 92: One thing to note for admin fields that reference the `User` object, we should use `raw_id_fields` or some other searchable field, otherwise admin is _very_ slow to render as python.org has _many_ users (valid and not)

minutes/models.py
```diff
@@ -41,3 +42,121 @@ def get_date_month(self):
...
+
+    class Meta:
+        verbose_name = "concerned party"
+        verbose_name_plural = "concerned parties"
+
+
+class Meeting(models.Model):
+    title = models.CharField(max_length=500)
+    date = models.DateField(db_index=True)
+    started_at = models.DateTimeField(null=True, blank=True)
```
ewdurbin: We can skip detailed time information for the time being.

minutes/models.py
```diff
@@ -41,3 +42,121 @@ def get_date_month(self):
...
+    parties = models.ManyToManyField(ConcernedParty)
```
ewdurbin, line 121: By default, I think we'd generally be relating "attendees" to the `Meeting` object by `ConcernedRole`. Individually specifying parties is tedious.

Supporting ad hoc attendees is a good idea, but this field should probably be renamed `additional_parties` or `ad_hoc_parties` and be hidden under the fold in Django admin.

minutes/admin.py
berinhard: Ah, sure! Fixed in 7099ae5

minutes/models.py
berinhard: Ok! Removed in ecf1725

minutes/models.py
berinhard: Got it @ewdurbin! I thought about this `parties` m2m as a way to decouple the `Meeting` model from the concerns/roles structure. My initial idea was that the UX to create a new meeting would only allow the user to select many `ConcernedRole`. Then, the backend would fetch for the parties from the roles list and add them to this attribute. I thought that we would be able to enable an easy way to create meetings, but still allow a granular control when editing that. But I'm ok with removing this and leaving a m2m field to `ConcernedRole` model.

About the `ad_hoc_parties` I'm a +1 for keeping it separated from the primary list. But, just to confirm, shouldn't this attribute be a m2m to the `User` model? By doing so, we won't required the user to create a `ConcernedParty` entry for that user who will join a single meeting.

berinhard:
@ewdurbin one question I have about the Ci. [This PR build](https://travis-ci.com/github/python/pythondotorg/builds/189514281) is broken due to lack of coverage, but I'm not sure how to fix that. The output says:

```
TOTAL                                                    3672   3139    683      0    12%

Coverage failure: total of 12 is less than fail-under=75

The command "coverage report -m --fail-under=75" exited with 2.
```

As you see, the coverage is in 12% when it should be 75% percent. But I'm afraid I didn't add an amount of code that had such impact in the coverage rate. Do you know if I missing something in the coverage config or do you have any tip about this error?

minutes/models.py
ewdurbin: Aha! That makes total sense actually, decoupling in that way is a good idea since members of a given `ConcernedRole` will change over time. 👍

As far as relating `ad_hoc_parties`, I think it's best if we keep that related to a `ConcernedParty`. Again this circles back to the fact that there are a _multitude_ of `User` objects in the database.

ewdurbin:
Looks like you didn't declare your new dependency: https://travis-ci.com/github/python/pythondotorg/builds/189514281#L342

JacobCoffee:
Hi @ewdurbin Can this be closed or does it need to be brought up to date and shipped?

ewdurbin:
The board and executive director who were interested in this feature are no longer around... so I think we can let this go.

PR #1679, opened 2020-11-16
Send email notifications once a sponsorship application is submited

berinhard:
Besides the notifications, this PR also introduces a few DB preparations for the application review process, such as `Sponsorship.for_modified_package` and `SponsorBenefit.benefit_internal_value`.


sponsors/notifications.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def notify(self, user, sponsorship):
+        context = {'user': user, 'sponsorship': sponsorship}
+
+        send_mail(
+            subject=self.get_subject(context),
+            message=self.get_message(context),
+            recipient_list=self.get_recipient_list(context),
+            from_email=settings.DEFAULT_FROM_EMAIL,
+        )
```
berinhard: It's very likely that this class will become even more abstract so it can be re-used by the upcoming email notifications in the sponsorship process.

sponsors/notifications.py
```diff
@@ -0,0 +1,43 @@
...
+    def get_recipient_list(self, context):
+        return [settings.SPONSORS_TO_EMAIL]
+
+
+class AppliedSponsorshipNotificationToSponsors(AppliedSponsorshipNotification):
+    subject_template = "sponsors/email/sponsor_new_application_subject.txt"
+    message_template = "sponsors/email/sponsor_new_application.txt"
+
+    def get_recipient_list(self, context):
+        return [context["user"].email]
```
berinhard: @ewdurbin I think I missed something here. Just to make sure, this notification should be sent to:

- the logged user;
- all email flagged as primary contact for that sponsor;

Is this correct?

ewdurbin: Not quite. We want to notify the `User` submitting the `Sponsorship` as well as any `SponsorContact` for which we already have a verified email address, I'll update this shortly.

berinhard: Hm, got it! Both the user and the verified email address would receive the same email, correct?

Also how can I confirm if an email address is a verified one? Should I check if we have users with that email or use some  kind of email verification service?

ewdurbin: see https://github.com/python/pythondotorg/pull/1679/commits/a5f3d854672a66e3ce36e04a31225819ad312aa2

sponsors/notifications.py
```diff
@@ -40,4 +42,10 @@ class AppliedSponsorshipNotificationToSponsors(AppliedSponsorshipNotification):
...
 
     def get_recipient_list(self, context):
-        return [context["user"].email]
+        emails = [context["user"].email]
+        for contact in context["sponsorship"].sponsor.contacts.all():
+            if EmailAddress.objects.filter(
+                email__iexact=contact.email, verified=True
+            ).exists():
+                emails.append(contact.email)
+        return list(set({e.casefold(): e for e in emails}.values()))
```
berinhard: This is is great! But I think we can refactor this logic by moving this to a method in the `Sponsor` object. Something like:

```
return context['sponsorship'].sponsor.get_verified_emails()
```

sponsors/views.py
```diff
@@ -117,13 +116,20 @@ def form_valid(self, form):
...
+        notified = uc.notifications[1].get_recipient_list(
+            {"user": self.request.user, "sponsorship": sponsorship}
+        )
 
-        response = super().form_valid(form)
+        response = render(
+            self.request,
+            "sponsors/sponsorship_application_finished.html",
+            context={"sponsorship": sponsorship, "notified": notified},
+        )
```
berinhard, line 163: Nice refactoring! Definitely redirecting to a success page was me doing some big design up-front

sponsors/notifications.py
berinhard: Introduced in 5a0902c

sponsors/views.py
ewdurbin: yeah, no need for a separate URL, and it's nice to show information back to the user so they build confidence that we "got it"

PR #1684, opened 2020-11-20
Sponsorship review ux

berinhard:
This PR enables admin users to Reject/Approve sponsorship applications;

**Reject/Approve buttons**
Screenshot from 2020-11-20 13-13-53

**Confrim approve form**
Screenshot from 2020-11-20 13-13-02

**Confirm reject form**
Screenshot from 2020-11-20 13-12-56


templates/sponsors/partials/full_sponsorship.txt
```diff
@@ -11,7 +11,9 @@
 # Sponsorship Information
   * Status		{{ sponsorship.status }}
 {% if sponsorship.level_name %}  * Level		{{ sponsorship.level_name }}{% else %}  * Level			Custom Package{% endif %}
-{% if for_modified_package %}  * Sponsorship Fee	To be determined{% else %}  * Sponsorship Fee	${{ sponsorship.sponsorship_fee|intcomma }} USD{% endif %}
-  * Benefits
-{% for benefit in  benefits %}    - {{ benefit.program.name }} - {{ benefit.name }}
-{% endfor %}
+{% if not display_fee %}  * Sponsorship Fee	To be determined{% else %}  * Sponsorship Fee	${{ sponsorship.sponsorship_fee|intcomma }} USD{% endif %}
```
berinhard, line 16: I realized that the new address fields are missing herein the `full_sponsorship.txt` =S

berinhard:
Print screens of detailed mail/billing address information:

**sponsorship admin detail**
Screenshot from 2020-11-24 09-49-25

**full sponsorship partial**
Screenshot from 2020-11-24 09-49-59

templates/sponsors/partials/full_sponsorship.txt
berinhard: Fixed in 
be6273c

sponsors/models.py
```diff
@@ -260,6 +263,9 @@ class Sponsorship(models.Model):
     level_name = models.CharField(max_length=64, default="")
     sponsorship_fee = models.PositiveIntegerField(null=True, blank=True)
 
+    def __str__(self):
+        return f"{self.level_name} ({self.get_status_display()}) for sponsor {self.sponsor.name}"
```
ewdurbin: may want to include start/end date in the string representation, it won't be an issue for first term (initial launch) but will be important for differentiating down the line.

sponsors/models.py
```diff
@@ -310,10 +316,16 @@ def estimated_cost(self):
...
 
     def reject(self):
+        if self.REJECTED not in self.next_status:
+            msg = f"Can't reject a {self.get_status_display()} sponsorship."
+            raise SponsorshipInvalidStatusException(msg)
         self.status = self.REJECTED
         self.rejected_on = timezone.now().date()
 
     def approve(self):
+        if self.APPROVED not in self.next_status:
```
ewdurbin, line 332: we probably want to ensure that a valid start/end date was provided as part of the approval process. this should be enforced, as we'll need it for generating Statement of Work.

ewdurbin (approved):
All looks great. I think there is some business logic I need to verify with sponsorship team, but that shouldn't block us from merging this to move things along.

ewdurbin:
For the purposes of shipping this, I'd like to see how we can display values that need to be provided before approval via admin on the confirmation screen... is that possible? Can we embed a (potentially dynamic) form?

sponsors/models.py
berinhard: Fixed in d134d3c

sponsors/models.py
berinhard: Totally! I'll update this method to do so, but will do in the statement of work's branch, is it ok? I've changed how the accept view works over there, so it'll be better to change this from a more up to date code.

ewdurbin: Makes sense! Thank you.

ewdurbin (approved):

berinhard:
About the dynamic form in the admin, sure we can do that =)

Currently, the approve view is a regular form view which expects a valid post to proceed. Currently there's no form because there's no data. I'll add a simple form for now with with fields for the `start_date` and `end_date`. But, as said [here](https://github.com/python/pythondotorg/pull/1684#discussion_r539513282), I'll base myself in the most recent code base.

PR #1685, opened 2020-11-23
remove access guards on new sponsorship application

ewdurbin:
soft launching in preview for some sponsors today.


berinhard (commented):
This looks good to me! One less `TODO` to be worried about =)

PR #1690, opened 2020-11-25
Better control of sponsor package's price display after customization

berinhard:
This PR fixes the issue with the sponsor cost display when user customizes the benefits but customizes it back to package's ones without changing the package. Under this scenario, the price wasn't being displayed again and this PR fix this. 

To do so I had to do some refactoring in the frontend code because I was getting very confused with the current strategy to get the dom elements via jQuery. 

Here's a demo of the feature working:

demo

### DIFF
--- a/static/js/sponsors/applicationForm.js
+++ b/static/js/sponsors/applicationForm.js
@@ -4,14 +4,15 @@ $(document).ready(function(){
         costLabel: $("#cost_label"),
         clearFormBtn: $("#clear_form_btn"),
         packageInput: $("input[name=package]"),
+        applicationForm: $("#application_form")
     }
 
 
     let cost = 0;
 
     SELECTORS.clearFormBtn.click(function(){
-        $("#application_form").trigger("reset");
-        $("#application_form [class=active]").removeClass("active");
+        SELECTORS.applicationForm.trigger("reset");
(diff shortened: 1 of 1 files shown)


static/js/sponsors/applicationForm.js
```diff
@@ -1,64 +1,94 @@
...
+      let isChangeFromPackageChange = SELECTORS.costLabel().html() == "Updating cost..."
+      if (!isChangeFromPackageChange) {
+        let selectedBenefits = SELECTORS.getSelectedBenefits();
+        selectedBenefits = $.map(selectedBenefits, (b) => $(b).val()).sort();
+        let selectedPackageId = SELECTORS.packageInput().filter(":checked").val()
+        let packageBenefits = SELECTORS.getPackageBenefits(selectedPackageId);
+        packageBenefits = $.map(packageBenefits, (b) => $(b).text()).sort();
+
+        // check same num of benefits and join with string. if same string, both lists have the same benefits
+        if (packageBenefits.length == selectedBenefits.length && packageBenefits.join(',') === selectedBenefits.join(',')){
```
ewdurbin, line 72: clever solution! will the sort order definitely be stable?

ewdurbin: Doh! I see the call to sort on line 69 :-D

PR #1692, opened 2020-11-30
Remove TODO notes in sponsors app

berinhard:
This PR removes 2 `TODO` comments and now the app:

1. The backend now validates if it can read/write cookies before redirecting the user to fill in the sponsor information;
2. Removed legacy code related to an old template tag;

### DIFF
--- a/sponsors/tests/test_views.py
+++ b/sponsors/tests/test_views.py
@@ -135,7 +135,7 @@ class SelectSponsorshipApplicationBenefitsViewTests(TestCase):
         form = r.context["form"]
 
         self.assertIsInstance(form, SponsorshiptBenefitsForm)
-        msg = "You have to allow python.org to use cookies to proceed."
+        msg = "You must allow cookies from python.org to proceed."
         self.assertEqual(form.non_field_errors(), [msg])
 
 
--- a/sponsors/tests/test_views.py
+++ b/sponsors/tests/test_views.py
@@ -2,13 +2,14 @@ import json
 from model_bakery import baker
 from itertools import chain
 
+from django.conf import settings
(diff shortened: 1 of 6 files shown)


berinhard:
There are still 2 open `TODO` in the `sponsors` app, but I don't think we have enough information to fix none of them by now. That's why this PR only covers 2 of the open opens.

sponsors/views.py
```diff
@@ -57,12 +58,21 @@ def get_initial(self):
         return cookies.get_sponsorship_selected_benefits(self.request)
 
     def form_valid(self, form):
+        if not self.request.session.test_cookie_worked():
+            error = ErrorList()
+            error.append("You have to allow python.org to use cookies to proceed.")
```
ewdurbin:
```suggestion
            error.append("You must allow cookies from python.org to proceed.")
```

ewdurbin (approved):

sponsors/tests/test_views.py
```diff
@@ -117,20 +122,22 @@ def test_redirect_to_login(self):
...
 
         self.assertRedirects(response, redirect_url, fetch_redirect_response=False)
 
+    def test_invalidate_post_even_if_valid_data_but_user_does_not_allow_cookies(self):
+        # do not set Django's test cookie
+        r = self.client.post(self.url, data=self.data)
+        form = r.context["form"]
+
+        self.assertIsInstance(form, SponsorshiptBenefitsForm)
+        msg = "You have to allow python.org to use cookies to proceed."
```
ewdurbin:
```suggestion
        msg = "You must allow cookies from python.org to proceed."
```

PR #1698, opened 2020-12-04
Fix broken link

### LINKED ISSUE #1697: Main doc portal has a weird instance of plain text
On https://www.python.org/doc/ I notice that on the right side, under the heading "Porting from Python 2 to Python 3", there is one bullet point that is not linked to anywhere and just sits as plain text: "Determine what projects are blocking you from porting to Python 3". It seems like a weird place to have a bullet point of plain text that's quite a bit more specific than surrounding bullets.

Is this line of text supposed to be on a different page? Is it supposed to be linked to another page that explains how to find which projects are blocking you? If the answer to both is no, why is it sitting somewhere where (to me, at least) it looks out of place, and what can be done to make it less jarring and less confusing?

- OS: Debian bullseye
- Browser: Firefox latest stable

hugovk:
Fixes https://github.com/python/pythondotorg/issues/1697.

### DIFF
--- a/templates/python/documentation.html
+++ b/templates/python/documentation.html
@@ -85,7 +85,7 @@
                           <li>
                             <a href="https://docs.python.org/3/howto/pyporting.html">Porting Python 2 Code to Python 3</a>
                             <ul>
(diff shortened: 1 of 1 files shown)


orsenthil (approved):

Mariatta (approved):
Thanks for looking into this and fixing.

PR #1702, opened 2020-12-09
Create Statement of Work after approving sponsorship + Send PDF document

### LINKED ISSUE #1748: Capture who approved sponsorship in change history
This is not necessarily a problem, but being able to track approvals to staff is important for our upcoming audit. Right now when a sponsorship is updated in /admin, it captures who made changes to fees, dates, and benefit items. It doesn't explicitly call out the change of approval. Can we add that in? 

Here's a snippet of what it looks like for one of our sponsors:

Screen Shot 2021-03-17 at 1 51 47 PM

A clear and concise description of what you want to happen. Ex. It would be great if [...]

We'd like "approved" or "approved sponsorship" added to the action column so we can see which staff person approved a sponsorship.

berinhard:
This PR adds:

- Form to input data such as `start_date`, `end_date` to properly accept a sponsorship applications;
- Automatically creates a statement of work for the approved application;
- Statement of Work's admin management;
- SoW's preview
- Fix #1748


berinhard:
@ewdurbin I updated this branch with the contract's content. 
[Here's a sample](https://github.com/python/pythondotorg/files/6265753/sample-contract.pdf) of a PDF generated by this PR.

ewdurbin:
This is working _excellently_ from my perspective. Next steps will be to ask the right folks to review an example contract.

We still do want to change all references of "Statement of Work" to "Contract" on the backend.

ewdurbin:
I am noting that once Approved, the "Review Document" button still appears in admin, but is not functional.

templates/sponsors/admin/preview-statement-of-work.html
```diff
@@ -0,0 +1,283 @@
...
+hereinafter sometimes individually referred to as a <b>“Party”</b> and collectively as the <b>“Parties”</b>.</p>
+
+<h1 class='center'>RECITALS</h1>
+
+<p class='paragraph'><b>WHEREAS</b>, the PSF is a tax-exempt charitable organization (EIN 04-3594598) whose
+mission is to promote, protect, and advance the Python programming language, and to support
+and facilitate the growth of a diverse and international community of Python programmers (the
+<b>“Programs”</b>);</p>
+
+<p class='paragraph'><b>WHEREAS</b>, Sponsor is {{ sponsor.description }}; and</p>
```
ewdurbin: I don't think we want the "description" here.

berinhard: @ewdurbin I don't have the link for the sample contract I based myself when coding this, but there was some dynamic information that should be displayed here. From what we currently have in the `Sponsor` model, the `description` attribute sounded as a good fit. Maybe are we missing this input in our models?

berinhard:
@ewdurbin I've updated this PR after your comments. The issue with the "Review Document" button was that there was no data migration to create contracts for existing and approved applications. I added a management command instead of a data migration (I wrote comments with a detailed explanation on why) so, you'll have to run the following command before testing this again:

`$ ./manage.py create_contracts`

berinhard:
@ewdurbin I took the liberty to add more things to this PR. I've worked in the fixes to #1748 and also added a similar log entry when the contract is sent. Here are print screens with results from my local env:

**Sponsorship**
Screenshot from 2021-04-16 16-20-17

**Contract**
Screenshot from 2021-04-16 16-41-31

ewdurbin:
Thank you @berinhard! I think this is an excellent start. Just being able to generate PDF contracts is a big win for the PSF staff.

I'm going to push to get final approval from the necessary folks, and then this is ready to merge.

ejodlowska:
Something that just came to mind: we should not generate a contract/SOW for OSS booths!

ewdurbin:
They will be generated, but do not need to be sent, signed, or reuploaded.

berinhard:
@ewdurbin I finished getting back to the context of this PR to finalize it. I pushed a few new commits addressing some changes.

For some reason the benefits --> program FK was nullable. This was breaking the tests and, since all the existing benefits are linked to a program, I thought it was reasonable to change the FK to be a required one instead.

Also, I've disabled the contract file from being sent to the client as we've talked last week. That way, PSF staff will be able to fully use admin to keep track of the contracts without worrying of automatically notifying the sponsors about it. Once the contract PDF gets a final confirmation and PSF staff gets used to the admin workflow, we can uncomment the line to enable the email sending.

And last, but not least, I realized that the final steps on contract review process were missing: to execute/nullify a contract. This is necessary so I can generate the CSV for #1806 knowing if the sponsorship was finalized (reviewed and with approved contract as well). I took care of creating log entries for such activities for PSF's internal control.

But, besides that, I think now it's safe to merge and deploy this PR to production. Let me know what do you think :+1:

PR #1711, opened 2020-12-23
Enable to rollback sponsorship applications to editing 

berinhard:
The discussion in #1710 introduced the idea of having an option to rollback the sponsorship to an editable state. Me and @ewdurbin talked briefly about this and agreed that the strategy is good because it prevents the code from unexpected changes in the application via admin.

This PR mirrors the same strategy as in Accept/Reject that's using a button + a simple page to confirm the rollback operation. This code doesn't know nothing about the statement of work yet. I'll update the PR #1702 with this branch and introduce SoW's required changes in this rollback workflow.


berinhard:
Screenshot from 2020-12-23 10-14-10

This is how the button is being displayed in the change form.

ewdurbin (approved):

Issue #1701, opened 2020-12-09
Calendar events should show the year for events which are not in the current year


malemburg:
See e.g. https://www.python.org/events/ the EuroPython conference entries.

![image](https://user-images.githubusercontent.com/6735498/101673932-1a6fce00-3a58-11eb-9ff5-84f0885139b2.png)

alvindera97:
I've sent in a PR: https://github.com/python/pythondotorg/pull/1714 to address this issue. Hopefully it gets reviewed soon as the maintainer of repository is available.

alvindera97:
#1714

PR #1713, opened 2020-12-25
Issue #1701: Calendar events now shows the year for events which are not in the current year

alvindera97:
Any calendar event that is dated beyond the current year now displays the year at which it has been scheduled. This happens for both start date and end date. 

I must say that I encountered serious difficulty with Django's `LiveServerTestCase` and at the end, I decided to just merge the 2 functional test method into one. Django's `LiveServerTestCase` for some reason **cannot** run multiple test methods [belonging to the same test class] at a stretch. This is because at the end of executing each test method, django sort of resets the data base leading to a lot of integrity errors and confusing test results. 

Please feel free to ask me any questions. This is my first time contributing to the open source community. 
Thanks in anticipated reply.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,8 @@ static/stylesheets/no-mq.css
 static/stylesheets/style.css
 __pycache__
 *.db
+
+
+# Ignore Microsoft Visual Studio Code's and Jetbrain's Pycharm Files
+.vscode/
+.idea/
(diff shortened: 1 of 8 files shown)


alvindera97:
On my local machine here, where I have selenium installed `pip install selenium`, all my tests work perfectly fine. However, I can see that the CI has spotted a failure in one of the tests that I wrote — directly related to `selenium`. I really don't know any other way to go about this. 

I am not aware if I can even attempt to change the configurations so that the chance to bring in functional testing. At least I would have set up a sequence for the CI to install selenium and add the required downloaded driver to the computer's path.

alvindera97:
I will now be submitting another PR with changes made to the tests so that they do not require selenium to run. 

**I will be basically replicating test concerns from the functional tests to unit tests** _[which don't require selenium to run_]

Issue #1729, opened 2021-02-09
Refresh of UserDetail view


ewdurbin:
When navigating to "Your account" after signing in to www.python.org, a very bare bones "dashboard" is visible to the user.

With #1728 adding additional information/views that will likely be accessed via this area, we should take a look at reorganizing this area of the site and potentially making other improvements to rendering and styling.

At the very least, we should update navigation to be more intuitive, currently the sections are just kind of plopped in there. We should probably organize them in some hierarchy to allow for future expansion.

With the addition of sponsorship, something like:

* Your Account
    * View/Edit Profile
    * Change Password
* Membership
    * View/Edit PSF Membership
    * Nominations
* Sponsorship (if applicable
    * View Sponsorships

Where possible, we should reuse other navigation components and styling.

aryanxk02:
Hey @ewdurbin I am new to Open Source Contribution and would like to contribute to the issues present in this repository. Could you please help me guide through this?? Thank You!

ewdurbin:
Hello @aryanxk02! @berinhard will be working on this. It's also not an _ideal_ candidate for a first issue. If you can identify something else that you might be interested in taking on, please let me know.

Issue #1728, opened 2021-02-09
Sponsors app: Create a view for Users associated with Sponsorships


ewdurbin:
Users associated with SponsorContacts should have a way to view the details of their Sponsorships that are in the 'APPLIED', 'APPROVED', and 'FINALIZED' state.

This may be accessible via the [`UserDetail`](https://github.com/python/pythondotorg/blob/2f5d5019ad89533cf863bd906515e35a21fa4bd1/users/views.py#L121) view and [associated template](https://github.com/python/pythondotorg/blob/master/templates/users/base.html) but doesn't necessarily need to be implemented in the `users` app.

Basic information that should be accessible for all associated `Sponsorship` objects:
  - Status (Applied, Approved, Finalized)
  - Effective and Expiration dates
  - Level and cost of sponsorship aggreement
  - Benefits
  - Associated contracts (Once #1702 is complete)

PR #1732, opened 2021-02-16
Refactor how logged user navbar is constructed

berinhard:
This PR opens path for us to work on #1728 and also makes #1729 work easier because not it comes down only to styling the nav bar.

@ewdurbin I decided to move the links construction to the backend via context processor, but I'm not sure how this will play with our cdn's cache. Let me know what you think about it.


ewdurbin (approved):
because the whole template is rendered together and is dependent on the content of the request (including cookies), constructing this via a context processor won't have any impact on our caching strategies.

I really like this approach as well, much cleaner than conditional logic in templates!

PR #1733, opened 2021-02-16
Sponsorship application detail view

### LINKED ISSUE #1728: Sponsors app: Create a view for Users associated with Sponsorships

### LINKED ISSUE #1729: Refresh of UserDetail view

berinhard:
Fixes #1728
Fixes #1729 

@ewdurbin here are the screen shots:

Screenshot from 2021-02-17 13-15-50

Screenshot from 2021-02-17 13-15-57


templates/sponsors/sponsorship_detail.html
```diff
@@ -0,0 +1,63 @@
...
+    {% if sponsorship.finalized_on %}
+    <li>
+      <b>Finalized date:</b> {{ sponsorship.finalized_on }}
+    </li>
+    {% endif %}
+    <li>
+      <b>Level:</b> {{ sponsorship.level_name }}
+    </li>
+    <li>
+      <b>Agreed sponsorship fee:</b> ${{ sponsorship.sponsorship_fee|intcomma }}
```
ewdurbin: After applying, this is currently showing the package price even if the benefits have been modified. We should show a similar "TBD" text as we do on the confirmation page.

static/sass/style.scss
```diff
@@ -2055,12 +2055,45 @@ $colors: $blue, $psf, $yellow, $green, $purple, $red;
...
+      padding: 0.25em 1em;
+
+      &:hover {
+          .section-links {
+              display: inline;
+          }
+      }
+
+      span {
+        @extend %button-style-yellow;
```
ewdurbin, line 2075: Screen Shot 2021-04-13 at 3 44 23 PM
This button styling doesn't seem to be working for me locally.

ewdurbin:
There are a couple of small things here that I think should be addressed, but the primary one is the styling of the navigation.

I am not sure why the styling shown in your screenshots isn't rendering for me, but if we could get something reliable and straight forward going, I think this is a good start.

berinhard:
@ewdurbin I updated this PR with 2 new commits addressing your comments. About the buttons style, I didn't commit a few lines in the final `style.css` file (e0ddfbf) and now it should work for you.

About the sponsorship fee information, I added a condition (d1c70dd) to check if the price should be displayed, but I'm wondering if the condition isn't to restrictive. Because it makes sense to me to don't display the price for new customized applications, but what about approved/finalized ones? I mean, at some point the `sponsorship_fee` will have the real and final agreed amount even for customized applications. So, what are these conditions?

ewdurbin:
> @ewdurbin I updated this PR with 2 new commits addressing your comments. About the buttons style, I didn't commit a few lines in the final `style.css` file ([e0ddfbf](https://github.com/python/pythondotorg/commit/e0ddfbf07bc15a89c00c78f3532171df38615dd8)) and now it should work for you.
> 
> About the sponsorship fee information, I added a condition ([d1c70dd](https://github.com/python/pythondotorg/commit/d1c70dd6046a347a653215d86c18de4263eb7bdd)) to check if the price should be displayed, but I'm wondering if the condition isn't to restrictive. Because it makes sense to me to don't display the price for new customized applications, but what about approved/finalized ones? I mean, at some point the `sponsorship_fee` will have the real and final agreed amount even for customized applications. So, what are these conditions?

I agree, that is a bit too restrictive. We basically shouldn't show a price for a modified package until approved.

ewdurbin (approved):

PR #1737, opened 2021-02-25
Bugfix: sponsorship package-only benefits disappearing 

berinhard:
Here are the steps listed by @ewdurbin to reproduce the error:

1. clear sponsorship_selected_benefits cookie
2. go to http://localhost:8000/sponsors/application/
3. Select ‘Supporting’
4. Click Submit
5. Click Back to select benefits
6. Note that Virtual Booth and Inclusion in a PSF blog about…. are in a weird mixed state (checked but disabled)
7. Click Submit
8. Note that benefits are missing
9. If you once again click “Back to select benefits” you will see the two benefits are now not checked at all.

Since the checkbox is disabled, the browser doesn't add it to the post body as part of the form data, even though it's being displayed as checked.

This PR fixes this by not disabling the checkbox if it's already checked and now, if you reproduce the steps above, you'll notice that both benefits are still checked but not disabled. 

I also added minor things to the sponsorship application admin. Both the new filter and the level name column helped me to investigate the database state faster, so I think it can help PSF staff as well.

### DIFF
--- a/sponsors/admin.py
+++ b/sponsors/admin.py
@@ -119,6 +119,21 @@ class SponsorBenefitInline(admin.TabularInline):
         return obj.open_for_editing
 
 
+class LevelNameFilter(admin.SimpleListFilter):
+    title = "level name"
+    parameter_name = "level"
+
+    def lookups(self, request, model_admin):
+        qs = SponsorshipPackage.objects.all()
+        return list(set([(program.name, program.name) for program in qs]))
+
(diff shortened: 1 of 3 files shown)


ewdurbin:
This seems to allow package only benefits to be selected from a blank form, and after clearing selections.

ewdurbin:
bleh, I was wrong, sorry :)

PR #1744, opened 2021-03-12
Implement new layout for the sponsorship application

berinhard:
@ewdurbin this PR implements the benefits form as specified in the Figma layout. I think the work is now solid enough to result in this PR since the page is completely functional for a desktop version.

I had to update a few `SponsorshipBenefit` objects in order to have them organized the same way as the layout. Also, to have a benefit as an add-on for any level, it must not have any Sponsorship Package associated to it. I added a few admin filters and improvements to help us to have the benefits somehow the same way the form lists them.

**Important** this PR does not contain the updated `style.css` file.


sponsors/forms.py
```diff
@@ -48,16 +49,23 @@ class SponsorshiptBenefitsForm(forms.Form):
...
+            "program"
+        )
+
         for program in SponsorshipProgram.objects.all():
             slug = slugify(program.name).replace("-", "_")
             self.fields[f"benefits_{slug}"] = PickSponsorshipBenefitsField(
                 queryset=benefits_qs.filter(program=program),
                 required=False,
-                label=_(f"{program.name} Sponsorship Benefits"),
+                label=_(f"{program.name} Benefits"),
```
merwok: gettext extraction will get the source string, but at runtime the f-string will be evaluated before it’s passed to gettext so this won’t be translated.
```suggestion
                label=_("{program_name} Benefits").format(program_name=program.name),
```

ewdurbin:
I think remaining fixes before we can ship this:

- [x] Ability to de-select benefits from a package
- [x] Render "Potential Add Ons" checkbox and deselected benefits as a "grey check" or "empty grey circle"
- [x] Stop package names from overlapping in the top bar in narrow desktop views

berinhard:
hey @ewdurbin! I just pushed new commits to this branch fixing the pending issues on the sponsorship benefit form. This page screams for a most up to date JS code strategy. To enable the user to customize the benefits, I had to introduce a lot of state control in the JS code. Also, since we're using images to control the input, I had to manually check the form's initial state to select which image to use and, also, which image to use next after a user click on the benefit.

berinhard:
hey @ewdurbin! I just pushed new commits to this branch fixing the pending issues on the sponsorship benefit form. This page screams for a most up to date JS code strategy. To enable the user to customize the benefits, I had to introduce a lot of state control in the JS code. Also, since we're using images to control the input, I had to manually check the form's initial state to select which image to use and, also, which image to use next after a user click on the benefit.

PR #1761, opened 2021-04-21
Topic/sponsorship form fixes

### LINKED ISSUE #1760: Tweaks to sponsorship's application form
To document a an issue, nice-to-have feedback from @ejodlowska about the sponsorship application form:

- Can we freeze the header row so when people scroll down they can see which column belongs to which package?
- The cost doesn’t change to custom pricing when changes are made to customize. Do we want to bring that back?

The first one is pretty straightforward as a CSS change, but the second touches the JS code. It'll have to check, whenever something changes, if the final list of selected benefits is the same as the select package's one. If the lists differs, the price should be replaced by a warning message.

berinhard:
Fixes #1760 

Here's a screen shot of both sticky header and custom fee message:

Screenshot from 2021-04-21 11-49-32

@ewdurbin I felt that adding a message was less disruptive with package header row organization than replacing the fee by a custom message. Even though, you can see that when the message appears, the header changes its heights and it annoys me a little bit. I didn't worry that much with the texting part as well, so review is needed.

### DIFF
--- a/templates/sponsors/sponsorship_benefits_form.html
+++ b/templates/sponsors/sponsorship_benefits_form.html
@@ -14,6 +14,7 @@
     <div id="select_sponsorship_benefits_container">
       <h1>Sponsor the PSF</h1>
       <p>Select a sponsorship package or customize the benefits that make the most sense for your business.</p>
+      <p><small>Obs: the listed fees are only valid if no changes were made to Sponsorship Package's default benefits.</small></p>
       <br/>
 
       <div class="row section-title">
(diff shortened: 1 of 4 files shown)


templates/sponsors/sponsorship_benefits_form.html
```diff
@@ -14,6 +14,7 @@
     <div id="select_sponsorship_benefits_container">
       <h1>Sponsor the PSF</h1>
       <p>Select a sponsorship package or customize the benefits that make the most sense for your business.</p>
+      <p><small>Obs: the listed fees are only valid if no changes were made to Sponsorship Package's default benefits.</small></p>
```
ewdurbin: Not sure what "Obs" means.

Copy can read "The listed sponsorship costs are subject to change if customization to selected benefits are made."

templates/sponsors/sponsorship_benefits_form.html
```diff
@@ -44,8 +45,9 @@ <h4 class="col">Select a Sponsorship Package</h4>
              {% for package in form.fields.package.queryset %}
              <div class="col col-items package-input" data-package-id="{{ package.id}}" onclick="updatePackageInput({{forloop.counter0}})">
                <h4>{{ package.name|upper }}</h4>
-               <span>${{ package.sponsorship_amount|intcomma }}</span>
+               <span class="package-price">${{ package.sponsorship_amount|intcomma }}</span>
```
ewdurbin: I wonder if we could add a hidden `*` that is also unhidden with the custom-fee message.

Copy for that message would then become `* Subject to change based on customization`

ewdurbin (changes requested):
In review, two small copy changes and a new request are made.

Additionally, as an aside, is it possible we could hide the radio buttons in the `package-input` elements?

templates/sponsors/sponsorship_benefits_form.html
berinhard: Ops, "Obs" is a common portuguese thing that I thought it did exist in english as well. It's like an abreviation for observations, amendments, disclaimers, etc.

berinhard:
@ewdurbin this PR is ready for a final review :+1:

ewdurbin (approved):

PR #1763, opened 2021-04-27
fix input-output order in interpreter example

sblondon:
The output of `input()` function should be printed immediately, not merged with the output of `print()` function:

```
>>> name = input('What is your name?\n')
What is your name?\nPython
>>> print('Hi, %s.' % name)
Hi, Python.
>>>
```

### DIFF
--- a/codesamples/factories.py
+++ b/codesamples/factories.py
@@ -33,7 +33,7 @@ def initial_data():
             >>> name = input('What is your name?\\n')
             <span class=\"output\">What is your name?
             Python</span>
-            >>> print('Hi, %s.' % name)
+            >>> print(f'Hi, {name}.')
             <span class=\"output\">Hi, Python.</span></code>
             </pre>
             """,
--- a/codesamples/factories.py
+++ b/codesamples/factories.py
@@ -31,10 +31,10 @@ def initial_data():
 
             <span class=\"comment\"># Input, assignment</span>
             >>> name = input('What is your name?\\n')
-            >>> print('Hi, %s.' % name)
             <span class=\"output\">What is your name?
-            Python
-            Hi, Python.</span></code>
+            Python</span>
+            >>> print('Hi, %s.' % name)
+            <span class=\"output\">Hi, Python.</span></code>
             </pre>
             """,
             """\


berinhard (approved):
Thanks for the catch.

sblondon:
Thanks @berinhard  for the review. Do we need to ping someone else for the merge?

berinhard:
Maybe a quick review from @ewdurbin or @Mariatta would be nice. Although I've been working on this code base for the last months, I'm not comfortable on merging PRs as the only reviewer.

codesamples/factories.py
```diff
@@ -31,10 +31,10 @@ def initial_data():
 
             <span class=\"comment\"># Input, assignment</span>
             >>> name = input('What is your name?\\n')
-            >>> print('Hi, %s.' % name)
             <span class=\"output\">What is your name?
-            Python
-            Hi, Python.</span></code>
+            Python</span>
+            >>> print('Hi, %s.' % name)
```
ewdurbin: I think we should highlight f-strings while we're here!

```suggestion
            >>> print(f'Hi, {name}.')
```

ewdurbin (changes requested):

codesamples/factories.py
ewdurbin: If there's no objection to this, I'm happy to commit the suggestion and merge the PR. Thank you @sblondon

sblondon: It's good for me.

There is a PR converting the string to f-string created by wkeithvan before this PR : https://github.com/python/pythondotorg/pull/1539 I guess we need to do something to it (like closing it with a link to this PR).

sblondon: I committed your suggestion and added wkeithvan and ewdurbin in the commit details.

ewdurbin (approved):

ewdurbin:
Thank you @sblondon

sblondon:
The fixed version is not visible on python.org. I hoped the release of python3.11 would embed it automatically but I was wrong.

Who should I request to push it publicly?

ewdurbin:
Didn't see the very bottom of the file where these load data into a model, thought they were generated from the factory code. Update has been applied.

PR #1813, opened 2021-07-13
Benefit Features db modeling

berinhard:
@ewdurbin this is still a WIP, but I'm opening it so you can review how I implemented the DB modeling we've talked about recently. Besides that, this PR also enables PSF staff from adding features to an existing benefit vai Django's stack inline admin. 

The process is pretty much intuitive:

- Click on a benefit to edit;
- Click on "Add another Benefit Feature";
- Select which feature you want to add on the menu that should open;
- Configure the feature and save the benefit;

The benefit feature can be afterwards edited and/or deleted, of course. Here are print screens to display the menu to pick the type of feature and an example of a sponsor placement at PSF's sponsor page:

**menu**

Screenshot from 2021-07-13 11-47-43

**logo at sponsors page**

Screenshot from 2021-07-13 11-50-18

I'm stuck to the "feature" naming because I couldn't thing about anything better. Help with naming or help texts is always an amazing support for non-english speakers =)


sponsors/enums.py
```diff
@@ -0,0 +1,18 @@
...
+    FOOTER = "footer"
+    DOCS = "docs"
+    DOWNLOAD_PAGE = "download"
+    DEV_GUIDE = "devguide"
+
+class PublisherChoices(Enum):
+    FOUNDATION = "psf"
+    PYCON = "pycon"
+    PYPI = "pypi"
+    CORE_DEV = "core"
```
berinhard, line 18: We have other choice fields in the sponsors app, such as sponsorship's status or contract's status. For the sake of integrity, I'd like to refactor them to also use `Enum` too if you think it's a good idea.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 -r base-requirements.txt
 -r prod-requirements.txt
+django-polymorphic==3.0.0
```
ewdurbin: move to base-requirements before merge.

sponsors/enums.py
ewdurbin: yes! i think this is a wonderful idea!

sponsors/models.py
```diff
@@ -819,3 +839,33 @@ def nullify(self, commit=True):
         if commit:
             self.sponsorship.save()
             self.save()
+
+
+class BenefitFeature(PolymorphicModel):
```
ewdurbin: This seems like a suitable base name honestly. "Features" will include:

- Logo Placements
- Requirements
- Deadlines
- Notifications

So I think this is the best generic term we can really come up with!

sponsors/models.py
```diff
@@ -819,3 +839,33 @@ def nullify(self, commit=True):
...
+    benefit = models.ForeignKey(SponsorshipBenefit, on_delete=models.CASCADE)
```
ewdurbin: Thinking now, we will likely need a set of "matching models" for relation to "SponsorBenefit" objects that mirror the base... similar to how we mirror "SponsorshipBenefit" data into instances of "SponsorBenefit".

Specifically so that values can be customized after implementation. I need to think about this, but for now I think we are on the right course for Sponsor Logo placements.

It may make sense to have a "BenefitFeatureConfiguration" that is associated with the SponsorshipBenefit... and then individual "BenefitFeature" instances that exist associated with individual SponsorBenefits.

ewdurbin (commented):
Thanks for making this concrete. In review I caught the concern highlighted in the comment regarding "BenefitFeature" model that we should probably clarify before moving forward.

Everything else looks great.

sponsors/models.py
berinhard: Hm, I see your point here! What do you think about having a [JSON field](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/fields/#jsonfield) on the sponsor benefit listing all the features that were applied when it was created. Something like this:

```python
sponsor_benefit.feature_configuration = [
  {
    "feature_model": "sponsors.models.SponsorLogoPlacement",
    "feature_kwargs": {"publisher": "psf", "logo_place": "sponsors"},
  }
]
```

That way we'll be able to always initialize all the features related to the sponsors considering the values that were defined at creation time. This is pretty much the same idea you suggested with the`BenefitFeatureConfiguration` model, but without having an extra model.

My concern on duplicating models is that we'll have to duplicate the one-level tree of specific benefit features as well. For example, for the logo placement we'll have to have a `LogoPlacementeConfiguration` (FK to `SponsorshipBenefit`) and a `LogoPlacement` (FK to `SponsorBenefit`). 

This duality also brings a new questions when it comes to specifics feature behavior implementation. For example, we'll probably have to implement a method under the logo placement model to define things such as the placement type (image only or image + text) or run validations. Where we should place such behavior if we have 2 very similar classes? It's highly probable that the right place would be the `SponsorBenefit` related class, but I don't have general understanding of upcoming benefits features to assert that this will always be true.

The downside I see when it comes to this JSON strategy is that refactoring, such as renames, on the features models will come with a high risky of breaking how existing `SponsorBeneift` configuration defines objects initialization. 

I'm good with the 2 models solutions since it's already what we're doing, but I just want to bring this other possibility to the table to exercise our thinking on this. Let me know what do you think about it :+1:

berinhard: @ewdurbin please, ignore this previous comment. I was able to reduce duplication using django abstract models as you can see in 8f8ac5b

berinhard:
@ewdurbin I think the DB modeling part of this work is done but I guess we have a problem with the build. It's failing due to this error:

```
ERROR: django-polymorphic 3.0.0 has requirement Django>=2.1, but you'll have django 2.0.13 which is incompatible.
```

This error wasn't happening in my local env because I've installed the project with `pip install django-polymorphic` and, thus, the installation proceeded with success. After reinstalling the package with `pip install -r dev-requirements`, as CI build does, I'm seeing the same error =S

Since we talked this week about upgrading Django version to 2.2 LTS, I think this build error is a new evidence suggesting this. If you're ok with it, I can open a new PR addressing this. If you want to take a look, here are the backward incompatibilities list:

- [2.1 backward incompatibilities](https://docs.djangoproject.com/en/3.2/releases/2.1/#backwards-incompatible-changes-in-2-1)
- [2.2 backward incompatibilities](https://docs.djangoproject.com/en/3.2/releases/2.2/#backwards-incompatible-2-2)

Django 2.2 will be supported until April 2022, which give us enough room to safely schedule a Django 3.0 upgrade too.

\* can we also rename the `master` branch to `main`?

berinhard:
@ewdurbin as we've spoken, I've created #1819 to address Django 2.2 LTS upgrade and changed django-polymorphic version to one which doesn't conflict with Django 2.0. Feel free to give this PR a final review and, if you feel everything is on place, merge it.  Thanks =)

sponsors/models.py
```diff
@@ -513,6 +530,13 @@ def new_copy(cls, benefit, **kwargs):
             **kwargs,
         )
 
+        # generate benefit features from benefit features configurations
+        for feature_config in benefit.features_config.all():
+            feature = feature_config.get_benefit_feature(sponsor_benefit=sponsor_benefit)
```
ewdurbin, line 535: I really appreciate that this is a method that can be defined! Will be super helpful in setting up things like "Deadlines" which may be dependent on the date of approval!

sponsors/models.py
```diff
@@ -819,3 +856,105 @@ def nullify(self, commit=True):
...
+
+    @property
+    def benefit_feature_class(self):
+        """
+        Return a subclass of BenefitFeature related to this configuration.
+        Every configuration subclass must implement this property
+        """
+        raise NotImplementedError
+
+    def get_benefit_feature(self, **kwargs):
```
ewdurbin, line 901: For now this is 💯, but it may make sense to leave this as unimplemented and allow for each Configuration to make decisions, although of course they can be overridden in inheritance :)

ewdurbin (approved):

PR #1826, opened 2021-07-26
Implement "tiered quantities" for sponsorship benefits and sponsor benefits

ewdurbin:
@berinhard this is my very initial bit at modeling "tiered quantities" for a given `SponsorshipBenefit`.

I'm not _positive_ if I'm using the polymorphic library just right...

TODO:
  - [x] sanity check with Bernardo
  - [x] implement the feature on the resulting `SponsorBenefit`
  - [x] style the benefits form
  - [x] render quantities along side benefits
  - [x] tests


ewdurbin:
The initial result here is with configuration such as:

Screen Shot 2021-07-26 at 3 41 58 PM
Screen Shot 2021-07-26 at 3 41 48 PM

We render:

Screen Shot 2021-07-26 at 3 28 59 PM

on the initial benefits selection form

berinhard:
@ewdurbin the model definitions are good so far and polymorphic usage is correct! I'm only concerned about db performance in the sponsorship page because of the `has_tiers` check for each listed benefit. But I think this is something we can address with a thin db-level caching layer.

sponsors/models.py
```diff
@@ -969,6 +973,18 @@ class Meta(BaseTieredQuantity.Meta, BenefitFeatureConfiguration.Meta):
         verbose_name = "Tiered Benefit Configuration"
         verbose_name_plural = "Tiered Benefit Configurations"
 
+    @property
+    def benefit_feature_class(self):
+        return TieredQuantity
+
+    def get_benefit_feature(self, **kwargs):
+        if kwargs["sponsor_benefit"].sponsorship.level_name == self.package.name:
```
ewdurbin, line 1010: thoughts on how to handle the fact that we don't keep a pointer to the source `SponsorshipPackage` around, @berinhard? This seems _OK_ since the package will almost certainly exist with the same name when the application is submitted... but still don't feel great about making it a string comparison.

ewdurbin:
@berinhard I'd love to get another review of my approach before implementing tests. I think there are some considerations we may need to take into account given how awkward some of the sponsorship package "handling" situations are.

sponsors/models.py
berinhard: I felt the same after coding the `list_sponsors` template tag. I coded a [similar package name checking](https://github.com/python/pythondotorg/blob/main/sponsors/templatetags/sponsors.py#L38) and I wanted to talk about a refactoring as well.

I'm not sure why we made the decision of not pointing to the `SponsorshipPackage` instance initially but this condition exposes that we need to have it. I can work on a data migration and refactoring to add this to the `Sponsorship` model.

sponsors/models.py
```diff
@@ -904,6 +940,9 @@ def benefit_feature_class(self):
         raise NotImplementedError
 
     def get_benefit_feature(self, **kwargs):
+        return self.build_benefit_feature(**kwargs)
+
+    def build_benefit_feature(self, **kwargs):
```
berinhard: This can be confusing for newcomers to this model because this can look like as an unnecessary indirection. I'd suggest to go on with a more Django CBV-like strategy to enable full customization on how to build a benefit feature. We could add a new method such as `get_benefit_feature_kwargs`.

This new method would build the default `kwargs` dict and return it **OR** return an empty dict if no benefit should be created. That way, we could still have the generic logic under the `get_benefit_feature` method as well.

berinhard (approved):
The code looks good to me, although I left some refactoring considerations. I don't think any of them are deal breakers, but they could be nice to have. 

Let me know if you want any help to code the tests.

sponsors/models.py
```diff
@@ -222,9 +229,20 @@ def __str__(self):
     def _short_name(self):
         return truncatechars(self.name, 42)
 
+    def name_for_display(self, package=None):
+        name = self.name
+        for feature in self.features_config.all():
+            if isinstance(feature, TieredQuantityConfiguration) and feature.package == package:
+                name = feature.display_modifier(name)
```
berinhard: I think that maybe we can refactor this code as well. We could refactor the `BenefitFeatureConfig.display_modifier` to accept  `**kwargs` alongside with the `name` parameter. That way we could move any type of conditional checking to the specific configuration instance. I think we could end up with something like:


```python
def name_for_display(self, package=None):
    name = self.name
    for feature in self.features_config.all():
        name = feature.display_modifier(name, package=package)
    return name
```

And, then, for the `TieredQuantityConfiguration.display_modifier`

```python
def display_modifier(name, **kwargs):
    package = kwargs.get("package")
    if package == self.package:
        return f"{name} ({self.quantity})"
    else:
       return name
```

This will be easier for us to unit test as well.

berinhard:
Hey @ewdurbin I pushed a few commits introducing refactoring, unit tests and performance upgrades. They don't address the package checking situation which, I agree with you, is sup-optimal. My suggestion is to address this issue in a separate and new PR since this one is already big enough and it doesn't require anything else to be merged in. I created issue #1840 to address the package thing specifically.

sponsors/models.py
ewdurbin: This approach makes the most sense to me 👍🏼

ewdurbin:
Thank you for pushing this over the line while I was away, @berinhard!

PR #1516, opened 2019-10-13
Rename to macOS

hugovk:
For #1254.

This changes the "static" content of things like "Macintosh", "OSX", "Mac OSX", "Mac OS X" to "macOS".

This doesn't change the `/downloads/mac-osx/` URL because I couldn't get Vagrant to work reliably to test it locally.

PR #1827, opened 2021-07-27
Python 3.9 upgrade

ewdurbin:
Minimal work to reach passing tests on Python 3.9.

Removes unmaintained Vagrant environment.

PR #1829, opened 2021-07-29
s/Mac OS X/macOS/

### LINKED ISSUE #1254: Change "Mac OS X" and "OS X" references to "macOS"
https://en.wikipedia.org/wiki/MacOS_version_history

I think there are at least two parts to this.  The easier one is to catch and change all the references in "static" content on pydotorg.  The second, and more complicated part, would be to change references in the CMS-managed content, in particular, the entries for the numerous Python releases: that content is managed by the cPython release managers.  I'm not so concerned about all that.

In any case, none of this is high-priority.

ambv:
The database on python.org is already updated, including supernavs and other boxes. There's also a dynamic redirect from the old downloads page to the new one which now is:

https://www.python.org/downloads/macos/

Supersedes #1516. Closes #1254.


coderanger:
Screen Shot 2021-07-29 at 3 32 43 PM

A redirect is in place using the dynamic system so the old URL still works too.

coderanger:
Merging this as it seems safe and the data side has already been done.

coderanger:
@ewdurbin Going to leave this in staging since I don't want to promote #1827 without your watchful eye :) It will be there when you get to it.

PR #1830, opened 2021-08-01
Upgrade to Django 2.2.

### LINKED ISSUE #1819: Upgrade to Django 2.2 LTS version
Currently python.org is running on top of Django 2.0.13 to which the support was dropped on April, 2019. Besides of don't getting security or other updates from other LTS versions, it also makes harder for the project to use most recent versions third-party apps.

To upgrade Django to 2.2 since it's a LTS one and will [have support until April 2022](https://www.djangoproject.com/download/#supported-versions).

Even though we already have Django 3.2 LT version, I think upgrading to Django 2.2 is a good initial step on upgrading directions. Also, we can only think about upgrading to Django 3.2 LTS after going through the 2.* upgrade pipeline. 

Also, I think this process will help us to understand the consequences of such upgrade with a smaller step and, thus, better prepare ourselves for the desired Django 3.2 upgrade.

**Additional context**
During this process, we must consider eventual backward incompatibilities introduced by Django 2.1 and 2.2 and address them. Here's the link to both lists:

- [2.1 backward incompatibilities](https://docs.djangoproject.com/en/3.2/releases/2.1/#backwards-incompatible-changes-in-2-1)
- [2.2 backward incompatibilities](https://docs.djangoproject.com/en/3.2/releases/2.2/#backwards-incompatible-2-2)

From what I've already seen, upgrading Django itself won't be a problem. The biggest chunk of work will be to update the Django third-party libraries used by Python.org. They can be found on `base-requirements.txt` and `dev-requirements.txt`.

If you're new to the project, please read our [install instructions](https://pythondotorg.readthedocs.io/install.html) to run it locally on your computer. You can tag me (@berinhard) or @ewdurbin for any doubts or questions on this process.

luzfcb:
Fix #1819

- Replaced the os.environ usage by [python-decouple](https://github.com/henriquebastos/python-decouple)
- Added the `env_sample` file with the environment variables that is available to use. If a file named `.env` exists on the project root path, the [python-decouple](https://github.com/henriquebastos/python-decouple) will read. Note: Environment variables have precedence over the python-decouple config files.
- Upgraded to Django 2.2.
- Updated dependencies for django 2.2 compatibility
- Synced django manage.py implementation with upstream
- Added the `apps.py` files for all modules that have migrations, template tags, or management commands
- Use `path()` or `re_path()` instead of `url()`
- Use `gettext_lazy` instead of `ugettext_lazy`
- Renamed `filter_fields` to `filterset_fields` and `filter_class` to `filterset_class` to be compatible with the new version of django-filter
- Removed `W605 invalid escape sequence \`  in  events/tests/test_importer.py (related to https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior)
- Fix factory_boy DjangoModelFactory imports
- Fix `url_name` function shadowing on pydotorg/context_processors.py 
- Added `FORM_RENDERER = 'django.forms.renderers.DjangoTemplates'` to `pydotorg/settings/base.py `
- Added the `obj=None` argument on `sponsors.admin.SponsorBenefitInline.has_add_permission` method
- Renamed `base_name` to `basename` on `pydotorg/urls_api.py` to be compatible with the new version of django-restframework
- Added a missing migration to the `added_by_user` field of `SponsorBenefit` model that that Django <=2.0 cannot detect. This migration only includes `blank=True` in the migration and does not touch the database, which means do NOT generate SQL.
- Load `static` templatetag instead of `admin_static`.
- Load `static` templatetag instead of `staticfiles`.
- Replace `NullBooleanField()` by `BooleanField(null=True)`. Note: The migration will not generate an SQL. [Django 2.1 release notes recommend this change](https://docs.djangoproject.com/en/2.1/releases/2.1/#models) but only in Django 3.1 `NullBooleanField` is effectively marked as deprecated.
- Replaced django-easy-pdf by django-easy-pdf3 (use `BytesIO` from `io` and not from `six`)

### TODO: 
- Upgrade django-allauth
- Migrate to sentry-sdk (raven is deprecated)
- Check if it is possible to generate the OpenAPI schema and after using it as a source for [schemathesis](https://github.com/schemathesis/schemathesis) generate and run tests and thus ensure that there are no update errors (at least in the underlying code that can be executed via API)

### Django 3.2 TODO:
Note: The code as it is in this pull-request is already capable of running the test suite with Django 3.2

- Add `DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'` to `pydotorg/settings/base.py` to avoid unwanted migrations
- Replace `from django.contrib.postgres.fields import JSONField` by `from django.db.models import JSONField` on `community/models.py` and deal with the migrations.
- Remove `default_app_config` variable from all `__init__.py` files of all apps.

--
### Pull-request sponsored by

[image](https://labcodes.com.br/)
[Need a team for your project? Let's talk](https://labcodes.com.br/)


luzfcb:
@berinhard @ewdurbin I have no idea the cause of this error. https://app.travis-ci.com/github/python/pythondotorg/builds/234219521#L1327-L1429
Any suggestion? Any help is welcome.

.gitignore
```diff
@@ -17,4 +17,9 @@ static/stylesheets/no-mq.css
 static/stylesheets/style.css
 __pycache__
 *.db
+*.py[co]
```
luzfcb, line 20: While I also think a global .gitignore is good, I don't see it as a good thing for the maintainer side. The world is wild out there. Providing a .gitignore with things we don't want to get into the codebase for me is the right way to go. Also, this facilitates open source contribution.
(Check the comments at the top of the file.)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
+repos:
```
luzfcb: I remember seeing a comment that keeping pre-commit up to date can be difficult. Well, we can automate that part too https://github.com/pydanny/cookiecutter-django/blob/master/.github/workflows/pre-commit-autoupdate.yml

Yes, I know the pre-commit is out of the scope of this pull-request.

luzfcb: This is the setting I use in one of my projects.

```yaml
  - repo: https://github.com/browniebroke/django-codemod
    # django-codemod is a tool to automatically
    # fix Django deprecations.
    # Any tag/version (>=1.5.8):
    # https://github.com/browniebroke/django-codemod/tags
    #
    # Automatic execution of this hook has been disabled
    # because it is a bit slow, but its configuration is
    # kept here, so it can be easily run:
    #
    # pre-commit run --hook-stage manual djcodemod
    # or
    # pre-commit run -a --hook-stage manual djcodemod
    rev: v1.5.8
    hooks:
      - id: djcodemod
        stages: [manual]
        args: [
            'run',
            '--deprecated-in', '1.9',
            '--deprecated-in', '1.10',
            '--deprecated-in', '1.11',
            '--deprecated-in', '2.0',
            '--deprecated-in', '2.1',
            '--deprecated-in', '2.2',
#            '--deprecated-in', '3.0',
#            '--deprecated-in', '3.1',
        ]
  - repo: https://github.com/asottile/pyupgrade
    rev: v2.24.0
    hooks:
      - id: pyupgrade
        args: ['--py3-plus', '--py39-plus']
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.0.1
    hooks:
    -   id: fix-byte-order-marker
    -   id: check-case-conflict
    -   id: check-merge-conflict
    -   id: check-symlinks
    -   id: debug-statements
    -   id: detect-private-key
    -   id: double-quote-string-fixer
    -   id: mixed-line-ending
    -   id: no-commit-to-branch
        args: [--branch, main, --branch, master]
  - repo: https://github.com/PyCQA/flake8
    rev: 3.9.2
    hooks:
    -   id: flake8
```

base-requirements.txt
```diff
@@ -1,46 +1,47 @@
...
-django-ordered-model==3.4.1
+djangorestframework==3.12.2
+django-filter==2.4.0
+django-ordered-model==3.4.3
 django-widget-tweaks==1.4.8
-django-countries==6.1.3
+django-countries==7.2.1
 xhtml2pdf==0.2.5
-django-easy-pdf==0.1.1
+django-easy-pdf@https://github.com/labcodes/django-easy-pdf/archive/60f2a70a056655d271a215ceded57f97ffaa4fea.zip
```
berinhard: @luzfcb why to use a custom fork instead of a regular version?

.pre-commit-config.yaml
berinhard: I'm still not sure about `pre-commit` =/

base-requirements.txt
```diff
@@ -1,46 +1,47 @@
 dj-database-url==0.5.0
-django-pipeline==1.6.14
-django-sitetree==1.10.0
-Django==2.0.13
+django-pipeline==2.0.6
+django-sitetree@https://github.com/idlesign/django-sitetree/archive/a2d77e9c282fc4be60698a8421c42a494393e794.zip
```
berinhard: Same thing here. Can't we use a more specific tag?

env_sample
```diff
@@ -0,0 +1,30 @@
...
+AWS_SECRET_ACCESS_KEY=
+AWS_STORAGE_BUCKET_NAME=
+
+
+# production optional
+#SPONSORSHIP_NOTIFICATION_FROM_EMAIL=
+#SPONSORSHIP_NOTIFICATION_TO_EMAIL=
+#AWS_DEFAULT_ACL=
+#AWS_S3_REGION_NAME=
+#AWS_S3_ENDPOINT_URL=
```
berinhard, line 30: Thanks for coming up with this file. I'm a fan of `.env` files to configure django apps as well.

events/tests/test_importer.py
```diff
@@ -45,7 +44,7 @@ def test_modified_event(self):
 DESCRIPTION:<a href="https://www.barcamptools.eu/pycamp201604">PythonCamp C
  ologne 2016</a>
 LAST-MODIFIED:20160401T210533Z
-LOCATION:GFU Cyrus AG\, Am Grauen Stein 27\, 51105 Köln\, Germany
+LOCATION:GFU Cyrus AG, Am Grauen Stein 27, 51105 Köln, Germany
```
berinhard, line 47: Why don't we need the escape char anymore?

users/models.py
```diff
@@ -86,8 +86,8 @@ class Membership(models.Model):
     postal_code = models.CharField(max_length=20, blank=True)
 
     # PSF fields
-    psf_code_of_conduct = models.NullBooleanField('I agree to the PSF Code of Conduct', blank=True)
-    psf_announcements = models.NullBooleanField('I would like to receive occasional PSF email announcements', blank=True)
+    psf_code_of_conduct = models.BooleanField('I agree to the PSF Code of Conduct', blank=True, null=True)
+    psf_announcements = models.BooleanField('I would like to receive occasional PSF email announcements', blank=True, null=True)
```
berinhard, line 98: Since this was deprecated only since 3.1, I think it's better to keep the `NullBooleanField` instead. That way, we'll reduce the volume of database migrations within this PR and leave more complex django 3 changes, such as these, to a subsequent PR.

users/migrations/0014_auto_20210801_2332.py
```diff
@@ -0,0 +1,28 @@
...
+        migrations.AlterField(
+            model_name='membership',
+            name='psf_code_of_conduct',
+            field=models.BooleanField(blank=True, null=True, verbose_name='I agree to the PSF Code of Conduct'),
+        ),
+        migrations.AlterField(
+            model_name='user',
+            name='first_name',
+            field=models.CharField(blank=True, max_length=150, verbose_name='first name'),
+        ),
```
berinhard: Remove this migration file if we roll back to use `NullBooleanField`.

sponsors/migrations/0031_auto_20210801_2038.py
```diff
@@ -0,0 +1,18 @@
...
+        ('sponsors', '0030_auto_20210715_2023'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='sponsorbenefit',
+            name='added_by_user',
+            field=models.BooleanField(blank=True, default=False, verbose_name='Added by user?'),
+        ),
+    ]
```
berinhard: Why does this migration exist within this PR? I mean, I don't see changes on `sponsors/models.py` file to generate new migrations.

berinhard (commented):
@luzfcb thanks again for opening such an important PR. I noticed that there are 3 tests failing but I didn't have the opportunity to run then locally to investigate the problem. As soon as I finish with other priorities, I'll investigate them. If you want to pair on them, I'll be happy to do so. I'm usually available during morning time.

base-requirements.txt
luzfcb: Is only this change:
https://github.com/labcodes/django-easy-pdf/commit/60f2a70a056655d271a215ceded57f97ffaa4fea

Was required to replace 
`from django.utils.six import BytesIO`
by 
`from io import BytesIO `

because the `django.utils.six` is deprecated.

Anyway, django-easy-pdf seems unmaintained project

base-requirements.txt
luzfcb: This points to the last commit on the master as there were some fixes not yet shipped in a regular release, however, 4 days ago the v1.17.0 release was made, so I will move to the correct version number.

events/tests/test_importer.py
luzfcb: Pycharm warned me about the error.

image

https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

Triple double-quote string is already a raw string, so I don't think it would make sense to put an `r` on the string definition. (please correct me if my conclusion is wrong)

users/models.py
luzfcb: I consider this change simple because it doesn't generate SQL.

```python
$ python manage.py sqlmigrate users 0014
BEGIN;
--
-- Alter field psf_announcements on membership
--
--
-- Alter field psf_code_of_conduct on membership
--
--
-- Alter field first_name on user
--
ALTER TABLE "users_user" ALTER COLUMN "first_name" TYPE varchar(150);
COMMIT;

```

Should I keep the NullBooleanField and create a new pull-request?

Changing the `first_name` is only necessary for when upgrading to Django 3.1, I will remove it

sponsors/migrations/0031_auto_20210801_2038.py
luzfcb: The #1682 includes the [`added_by_user` field with `blank=True`](https://github.com/python/pythondotorg/pull/1682/files#diff-07dd10655f36432bf9245868f1b9b19a10cad6abf5b7807169bc334975275b1cR377), [but the migration](https://github.com/python/pythondotorg/pull/1682/files#diff-82a26abccbb6e7cf641e0fec6f37e9f4edde489ef3f8a0070162ee5176b5bfb2R16) sent with the pull-request does not include `blank=True`.
The django migration framework makes it mandatory to create a migration whenever any changes to the Model, model field, model fields attributes and Meta are made even if this change (`blank=True`) does not generate SQL

luzfcb: I investigated a little more, and django 2.0.13 cannot detect the divergence between the last `added_by_user` migration and the `added_by_user` model field. As of django 2.1, something has changed in the django migration framework (I couldn't find anything to explain this in the release notes), and the divergence is finally detected, and `python manage.py makemigrations` generates a new migration.

luzfcb: the simplest way to reproduce the problem is on the main branch, update as minimum as possible so as not to have errors in the migration framework, and finally run `makemigrations`

```python
pip install "django<2.2,>2.0.13" django-tastypie django-sitetree django-filter==2.0.0 -U
python manage.py makemigrations

```

base-requirements.txt
luzfcb: fixed

users/models.py
luzfcb: Django 2.1 release notes also recommend this change https://docs.djangoproject.com/en/3.2/releases/2.1/#models

luzfcb: I removed the change made to first_name

luzfcb:
@berinhard I updated the description, made some requested changes, and rebased with the main branch. Can review again?

luzfcb:
I added a test case to check for missing migrations.
    If this test fails, something like this will be shown:
    
    ```python
    FAIL: test_missing_migrations (pydotorg.tests.test_migrations.MissingMigrationTestCase)
    Tests if there are any changes left in the Django models for which the migration script was not generated.
    ----------------------------------------------------------------------
      File "/home/travis/build/python/pythondotorg/pydotorg/tests/test_migrations.py", line 24, in test_missing_migrations
        self.assertFalse(
    AssertionError: {'sponsors': [<Migration sponsors.0031_auto_20210811_0901>]} is not false : Missing migrations were detected on the following apps: sponsors. Run `python3 manage.py makemigrations` to generate new ones.
    ```
    
    https://app.travis-ci.com/github/python/pythondotorg/builds/235045671#L1655
    
    yes, I know there is `python manage.py makemigrations --dry-run --check`, but that way, we ensure there will never be a missing migration even if we forget to run the `makemigrations --dry-run --check`
    
    
I created a separate pull-request for this issue https://github.com/python/pythondotorg/pull/1839

base-requirements.txt
berinhard: Thanks, it makes sense to me. @ewdurbin how to handle this? Should we add a fork Labcode's fork with @python organization so we have control of such type of customization.

sponsors/migrations/0031_auto_20210801_2038.py
berinhard: Thanks for the explanation! I'm good to keep this migration given the overall context.

users/models.py
berinhard: Got it! So I guess we can keep this since there's no SQL being ran. Thanks for removing the `first_name` migration too.

.pre-commit-config.yaml
luzfcb: Pre-commit settings have been removed.

templates/components/navigation-widget.html
```diff
@@ -14,6 +14,6 @@
 
 {% if URL_NAMESPACE %}
     {% sitetree_tree from URL_NAMESPACE template "sitetree/sidebar_menu_root.html" %}
-{% endif %}
-
+{% elif URL_NAME %}
     {% sitetree_tree from URL_NAME template "sitetree/sidebar_menu_root.html" %}
+{% endif %}
```
luzfcb, line 19: In theory, this makes the 3 tests start to pass
https://app.travis-ci.com/github/python/pythondotorg/builds/235762782#L1343-L1648

I'm not sure this is the correct way to fix this.

Anyway, thanks @lucianoratamero for helping me debug this error.

luzfcb:
@berinhard @ewdurbin this pull request is ready for review.

templates/components/navigation-widget.html
berinhard: From what I could understand, [this context processor](https://github.com/python/pythondotorg/blob/main/pydotorg/context_processors.py#L9), for some reason, wasn't raising the `Resolver404` exception for db-routes such as is happening with tests related to `Page` or `Release` objects. 

I also checked the `PageView` implementation to confirm if it wasn't doing something unexpected but I didn't see anything out of place.

I tested 2 type of dynamic pages to check what would break if the path wasn't configured correctly. If I access an URL (`/lorem-ipsum/`) mapped to a `Page` object with the URL not being listed under the sitetree, I can see the content but without the navigation links as you can see here:

Screenshot from 2021-08-24 12-03-50

If update this same object to have a path which is listed under the site tree, such as `/about/gettingstarted/`, the user sees both the page content and the navigation links as expected. Here's the print screen:

Screenshot from 2021-08-24 12-04-27

So, I think this fix is a good one :+1:

berinhard (approved):
@luzfcb the PR #1848 replaced travis by github actions and this is resulting in conflicts. Can you update `.github/workflows/ci.yml` file instead?

@ewdurbin I tested it locally and everything seems to be good so far. But since you have a more "trained" pair of eyes on how the site should look like, it would be nice to have your review and also your comments on the still open discussions.

luzfcb:
@berinhard 
> @luzfcb the PR #1848 replaced travis by github actions and this is resulting in conflicts. Can you update .github/workflows/ci.yml file instead?

Fixed.

luzfcb:
@berinhard I pushed a fix for PostgreSQL running in GitHub Actions

luzfcb:
@berinhard @ewdurbin  this pull-request is ready for review and merge.

base-requirements.txt
ewdurbin: that'd be preferred, unless the fork here is acceptable: https://pypi.org/project/django-easy-pdf3/ it seems that the only diff is _effectively_ the same change

sponsors/migrations/0037_add_added_by_user_sponsorbenefit.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.24 on 2021-08-19 13:16
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sponsors', '0036_auto_20210826_1930'),
```
ewdurbin: Some additional database migrations were introduced in #1852, so I think once we update this migration this is good to go! Sorry for the delay in review.

ewdurbin (changes requested):
Spun this up and smoke tested it with a copy of the prod db and all seems well! There's one migration in the sponsors app that needs to be updated and then we can ship it!

base-requirements.txt
luzfcb: I switched to django-easy-pdf3

sponsors/migrations/0037_add_added_by_user_sponsorbenefit.py
luzfcb: I fixed the migrations in https://github.com/python/pythondotorg/pull/1830/commits/6b218e546e50fceb1ade902bf18c0bcc95257833

luzfcb:
@ewdurbin I made the requested changes. It's ready for review. As it's the first time I've contributed to this project, Github Action workflow needs manual approval again to run

ewdurbin (approved):
Thank you for this effort!

ewdurbin:
Deploying now! I'll be keeping a close eye on the application through the rest of the day <3

ewdurbin:
Thanks so much @luzfcb! This has been rolling smoothly now for days!

ameenkhan003 (commented):
awesome

PR #1839, opened 2021-08-10
Configure github-actions to ensure not have migrations to be created

luzfcb:
This pull-request adds `python manage.py makemigrations --dry-run --check` to the github actions configuration to avoid forget to create a migration file when modifying a django model or update a dependency.

--
#### Pull-request sponsored by

[image](https://labcodes.com.br/)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,11 @@ jobs:
         run: |
           pip install -U pip setuptools wheel
           pip install -r dev-requirements.txt
+      - name: Check for missing migrations
+        run: |
+          python manage.py makemigrations --dry-run --check
+        env:
+          DATABASE_URL: postgres://postgres:@127.0.0.1:5432/python.org
       - name: Run Tests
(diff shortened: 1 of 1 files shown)


berinhard (changes requested):
@luzfcb I think failing tests because migrations are missing can be something annoying under certain scenarios (configuring blank for admin forms, for example), even though having this check is definitely a nice to have. What do you think about, instead of having these tests, adding a call for `python manage.py makemigrations --dry-run --check` to our CI file? 

In my opinion this sounds better because the contributor will still have a warning about the missing migration without having a direct affect on their development workflow.

berinhard (approved):

luzfcb:
@berinhard I updated pr to use Github Actions instead of Travis-ci

berinhard (approved):

raghavthind2005 (approved):

ewdurbin:
Closed by https://github.com/python/pythondotorg/pull/2248

Issue #1806, opened 2021-06-21
Replace hardcoded Python Software Foundation sponsor logo placements with EthicalAds


ewdurbin:
**Is your feature request related to a problem? Please describe.**

Certain tiers of Python Software Foundation sponsors receive recognition of their sponsorship on python.org with a logo, examples include:

  - https://www.python.org/downloads/
  - https://www.python.org/jobs/

Currently managing these logo placements is manual, performed via the [boxes](https://github.com/python/pythondotorg/tree/master/boxes) application.

Downsides are that it requires us to format HTML manually. We also receive no analytics on actions taken from these logo placements.

**Describe the solution you'd like**

Ideally, using the data modeled for sponsors in our existing [sponsors]() application:

  - Create necessary placements/flights via the EthicalAds API
  - Render the necessary flights on python.org based on the specific placement (downloads, jobs, etc...)

**Additional context**

We would like to have the links render in such a way that rather than going directly to the sponsor's landing page URL, they are taken to the [PSF sponsors page](https://www.python.org/psf/sponsorship/sponsors/). This will provide us with metrics on actions taken from sponsor logo recognitions, without turning them into advertisements.

Ideally the rendering of the sponsor recognitions will be similar to the current side bar recognitions on PyPI.org:

<img width="263" alt="Screen Shot 2021-06-21 at 10 57 28 AM" src="https://user-images.githubusercontent.com/1200832/122783665-c1db8200-d27f-11eb-83d6-e0fd37ee7468.png">

PR #1849, opened 2021-08-24
Introduce logo placement API endpoint

berinhard:
This PR is complementary to #1806 work. It introduces a new endpoint `/api/v2/sponsors/logo-placement/` to list all of the sponsor logo placement information. This endpoint replaces the previous `gen_sponsors_csv` management command we've created to start with the integrations with the logo placement server.

@ewdurbin, although this PR has all the code needed by the endpoint the CSV file still have invalid URLs for the logo images, such as https://pypi.org/static/images/sponsors/color/google.png. Feel free to update the CSV file with the new you'll generate after you go through the process of updating the uploaded files.


sponsors/api.py
```diff
@@ -0,0 +1,59 @@
...
+    logo = serializers.URLField()
+    start_date = serializers.DateField()
+    end_date = serializers.DateField()
+    sponsor_url = serializers.URLField()
+
+
+# TODO Currently this endpoint only lists sponsors from pypi sponsors CSV.
+# Once we have all sponsorship data input into pydotorg, we should be
+# able to change this view to fetch data from the database instead.
+class LogoPlacementeAPIList(APIView):
```
berinhard, line 38: @ewdurbin I noticed other API endpoints also define authentication and permission with:

```python
authentication_classes = (TokenAuthentication,)
permission_classes = (IsStaffOrReadOnly,)
```

I'm not sure if we need this same config for this endpoint. I mean, the permission class shouldn't be required since this is already a read-only endpoint. But what about the token authentication? Should we generate a token and send to Eric? Let me know what you think about that.

ewdurbin: Hmmm, perhaps the only notable thing worth guarding is the 'start/end' dates... so yes. Perhaps we should authenticate this endpoint and create a specific permission for it such as `SponsorPublisher` or similar?

I am not _positive_ if we have that kind of granular control.

ewdurbin (changes requested):

sponsors/api.py
berinhard: I guess we'll have to have this control anyway. If we enable the `TokenAuthentication` to work, the API will always require a token which is linked to an user. So, we'll always have to send tokens for clients of this endpoint and, thus, define users for them. When defining these users, which are specials by definition, we can also configure the specific permission. 

I agree with you that think defining a custom permission is the way to because:

- We won't leave it open just because it's a read-only endpoint;
- We won't have to create staff users to enable clients to access it;
- We don't make it visible to existing API users;

I'm updating this PR to add both authentication and permission

berinhard:
@ewdurbin this PR is ready for review. You can notice I added a missing migration with cb7ceb5. To make it easier to merged/deployed, I cherry-picked the commit on a separate branch and opened the PR #1851.

ewdurbin (approved):

PR #1858, opened 2021-09-07
Purge FTP index listing all Python versions from fastly cache

ewjoachim:
I could be wrong but I believe this is the reason why at the time of writing, https://www.python.org/ftp/python/ doesn't list the 2 new Python releases (3.6.15 & 3.7.12).

A manual cache may also be triggered for this page because this PR only addresses future Python releases.

Screenshot from 2021-09-07 07-49-12

Thank you :heart:


ewdurbin (approved):

ewdurbin (approved):

ewjoachim:
Yay thanks :)

PR #1865, opened 2021-09-08
New boolean to SponsorshipPackage

berinhard:
This PR introduces a new boolean flag `advertise` to the `SponsorshipPackage` model. By default it's false for new packages, but this PR includes a data migration to make sure all the existing ones get update. Also, the form accessible at `/sponsors/application/` only list packages with the flag enabled.


ewdurbin:
Thank you @berinhard!

PR #1892, opened 2021-10-15
Sponsorship dashboard

berinhard:
This PR updates the user account sections to improve how sponsorships are being managed.

It changes the current sponsorships detail page to also display sponsor information. The page now also display a link to a new form where the user can edit their sponsor information associated with the sponsorship. It also adds a sponsorship dashboard to list all the sponsorships. First the active ones (finalized and with valid start/end dates) and, them, any other one grouped by status.


sponsors/forms.py
```diff
@@ -463,3 +463,55 @@ def get_notification(self):
...
+        if self.data:
+            self.contacts_formset = factory(self.data, **formset_kwargs)
+        else:
+            self.contacts_formset = factory(**formset_kwargs)
+        # display fields as read-only
+        for disabled in self.READONLY_FIELDS:
+            self.fields[disabled].widget.attrs['readonly'] = True
+
+    class Meta:
+        exclude = ["created", "updated", "creator", "last_modified_by", "web_logo", "print_logo"]
```
ewdurbin: I think we can include web_logo and print_logo. But should not allow them to be cleared, only replaced.

ewdurbin (approved):
I don't think the note on logos is a blocker to merging this. If it is quick let's change it, but otherwise address it in another revision?

berinhard:
@ewdurbin Thanks for your review. I rolled back the inputs and remove the clear option from them. Once CI passes, I'll move on and merge this branch :+1:

berinhard:
Here's a print of how the fields are being displayed now:

Screenshot from 2021-10-19 13-06-57

PR #1893, opened 2021-10-15
added extra quatation mark

ghost:
just done some simple thing


ewdurbin:
this style comment uses three quotation marks by definition

PR #1900, opened 2021-10-22
Enable PSF staff to force a required image from a sponsorship via benefit configuration

berinhard:
!!! THIS IS WIP !!! 

This PR has a few commits from #1897, but once it gets merged, only the commits to address the custom assets will be displayed. 

I'm opening this PR to get feedbacks from @ewdurbin on the DB modeling. Here's how I though about how to support generic assets under the sponsorship app:

1. Define a pair of benefit feature/configuration models so admin users can configure the benefits;
2. Every feature/configuration must define an unique `internal_name` and a `related_to` choice field (sponsors / sponsorship). Both fields will be controlled to know to which object the asset should relate to and also to query over Generic FKs using the `internal_name` to check if the sponsor and/or sponsorship has the required asset;
3. The dashboard can iterate over all of the sponsorship's benefits and check the ones which require assets. Then, we'll use the `internal_name` and `related_to` object to fetch for the stored data in our DB via a more generic model;

I was thinking on following a very similar strategy with the documents assets (docx, pdf, ppts etc). We could define a new `RequiredDocumentAsset` configuration/benefit following the same strategy and using a new `GenericFileAsset` class to store the uploaded document.


sponsors/models/assets.py
```diff
@@ -0,0 +1,45 @@
...
+    ext = "".join(Path(filename).suffixes)
+    name = f"{instance.internal_name} - {instance.content_type} - {instance.object_id}{ext}"
+    return f"{directory}{name}{ext}"
+
+
+# Stores image assets
+class GenericImgAsset(models.Model):
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    content_object = GenericForeignKey('content_type', 'object_id')
```
ewdurbin, line 34: Is this what is handling the relationship to either a `Sponsor` or `Sponsorship` object?

sponsors/models/assets.py
```diff
@@ -0,0 +1,45 @@
...
+from django.contrib.contenttypes.models import ContentType
+
+
+def generic_asset_path(instance, filename):
+    """
+    Uses internal name + content type + obj id to avoid name collisions
+    """
+    directory = "sponsors-assets"
+    ext = "".join(Path(filename).suffixes)
+    name = f"{instance.internal_name} - {instance.content_type} - {instance.object_id}{ext}"
```
ewdurbin: I wonder if having a UUID on `GenericImgAsset` and using that `f{instance.internal_name}-{instance.id}{ext}` would be more clear? If my reading is correct though, this encodes the actual object type and id the asset is related to, which is kind of nice, but could potentially lead to people enumerating over ids to look for files if we accidentally mixed up access permissions in the object store.

ewdurbin (commented):
The modeling looks sound, assuming that the actual relations to `Sponsor`/`Sponsorship` objects will be handled in views/controllers.

sponsors/models/assets.py
ewdurbin: If so, let's add some comments to these fields to note that this is what's used to relate these to `Sponsor`/`Sponsorship` objects.

sponsors/models/assets.py
berinhard: Totally agree with adding an UUID to this model and use it to format the file path.

ewdurbin (approved):

PR #1928, opened 2021-11-25
docs: enhance sponsorship uc documentation

berinhard:
This PR tries to document better the sponsor app to make it easier for other developers to onboard it.

It mostly updates the use cases with docstrings explaining their logic. It also adds an image file with the database schema to the readthedocs page.

To generate the DB schema, I used [django-extensions graph model](https://django-extensions.readthedocs.io/en/latest/graph_models.html) command line interface.


docs/source/administration.rst
```diff
@@ -106,6 +106,13 @@ they represent:
 :contract.py: The `Contract` model which is used to generate the final contract document and other
               support models;
 
+.. image:: _images/sponsors-db.png
```
JacobCoffee, line 109: Not sure if the image (and the dependency that generates it) is necessary here as it is quite complex and I feel more lost with it 😓

adityaraute:
Does this PR still need a change?

PR #1941, opened 2021-12-16
Bugfix to don't ignore add-ons

berinhard:
This PR addresses the following feedback about the sponsorships module:

```
Add-on items did not work when being added to a package for Meta

Not confirmed, but report of add-on items being selected and displayed while
applying and confirming, but not associated with the resulting sponsorship application.
```

The bug has happening due to a missing logic to `SponsorshiptBenefitsForm.get_benefits` method. This method is used to get all the benefits that should be used during the new sponsorship creation. But it wasn't returning the add ons benefits, even though the data was present in `form.cleaned_data`. 

This PR fixes this logic and also updates the views tests to be more accurate about the benefits creation.

### DIFF
--- a/sponsors/forms.py
+++ b/sponsors/forms.py
@@ -89,11 +89,15 @@ class SponsorshiptBenefitsForm(forms.Form):
                 conflicts[benefit.id] = list(benefits_conflicts)
         return conflicts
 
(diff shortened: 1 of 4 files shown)


ewdurbin (approved):

ewdurbin:
Thanks @berinhard.

PR #1942, opened 2021-12-16
Enable admin staff to add benefits to Sponsorship via admin

berinhard:
This PR addresses the following feedback from sponsorship app:  

    - I can delete benefits from a sponsorship application, but I can't add them.

    Once application is complete, being able to add a sponsor benefit to an 
    application before approving/generating contract/finalizing will require logic to be added to 
    admin routes to correctly handle creation of sponsorship benefit objects.

It does so by not disabling the link add new entries by `SponsorshipAdmin`. It also updates the form used to inline display the sponsors to make sure the benefit features are updated as expected as well.


ewdurbin (approved):
Thanks Bernardo, I added one little missing bit with the `internal_value`, but otherwise this is working splendidly!

PR #1946, opened 2021-12-24
Enable to have sponsorship applications only with a la carte benefits

berinhard:
This PR implements the necessary support to have:

    A la Carte benefit only sponsorships. Example: Job Fair Table and Open Source/Community Booth
    on the production application form, these can be purchased completely independently of a sponsorship package.
    
The overall logic here does:

- Add a new `a_la_carte` boolean to both `SponsorshipBenefit` and `SponsorBenefit`;
- Proper admin validation to manage a la carte benefits;
- Sponsorship application form now lists both add-ons and a la carte benefits;
  - The difference between both is that an user can apply an application containing **only a la carte** benefits, but with no package. The add-ons, on the other hand, **always require a package** to validate the application.
- Defines a default `A La Carte Only` package that is created dynamically;
- CSS/HTML updates to properly display the form

**Screenshots**

Here's how the form is being displayed with **both add-ons and a la carte** benefits:

Screenshot from 2021-12-24 11-32-24

Here's it **only with a la carte**

Screenshot from 2021-12-24 11-33-05

As you can see, the steps counter gets updated dynamically by checking if there are or not both a la carte or add-ons. **IMPORTANT:** I've manually created the image for the title number 5  using Gimp. It's pretty much similar to the existing title images, but it's still different. I don't know which font was used to create the images so there's a mismatch between them. I thought about generating all images once again using the same font, but I wanted to check before if this is needed.

Last, but not least, here's a screen shot of the **summary of an a la carte only application**:

Screenshot from 2021-12-24 11-29-45

And here is the **same application in admin** after being finalized:

Screenshot from 2021-12-24 11-43-49

Screenshot from 2021-12-24 11-43-58


ewdurbin (approved):

ewdurbin:
Thanks @berinhard in review, the only note I had was that we _could_ probably set the sponsorship fee to the sum of A La Carte Only benefits if it's A La Carte Only... but we'll see what feedback we get from sponsorship team.

PR #1948, opened 2021-12-27
Improves response time of sponsors page

berinhard:
This PR adds an internal cache to not rely only on CDN caching. 

The sponsors list page takes time to finish the request for the first time because it performs a lot of database queries. It has to:

- List all packages;
- List all finalized sponsorships with logo placement for that page (here the Django's polymorphic models play an important role);
- Per image, run `solr-thumbnail` queries to locate the expected image thumbnail path;

This PR does reduce the initial load time, but there's still some room for improvements. Here's the difference between a cold-cache from a warm one:

**Cold-cache**

Screenshot from 2021-12-27 10-44-23

**Warm-cache**

Screenshot from 2021-12-27 10-44-32

**Improvements**

The high volume of DB queries (79 on cold-cache) is mostly `solr-thumbnail`'s responsibility. Under the hood, it uses a [Key Value Store](https://sorl-thumbnail.readthedocs.io/en/latest/requirements.html#key-value-store) to store the path of a specific thumbnail generated from an image. If we don't customize anything, it uses a **cached database** and this results in what we're seeing: multiple I/O operations in the database being called. 

To reduce I/O and, thus, CPU usage, we could switch the key value store to use memcached and, thus, have faster in-memory I/O operations. [From their docs](https://sorl-thumbnail.readthedocs.io/en/latest/requirements.html#cached-db), if we set up Django Cache to use memcached, it should be enough to improve the performance. 

Since this touches infra configuration as well, I didn't move on with that because I'd like to sync with @ewdurbin first to check if it makes sense to enable Django's cache on Pythondotorg's current stack.

### DIFF
--- a/templates/sponsors/partials/sponsors-list.html
+++ b/templates/sponsors/partials/sponsors-list.html
@@ -1,4 +1,8 @@
 {% load thumbnail %}
+{% load cache %}
+
+{% comment %}30-days cache{% endcomment %}
+{% cache 2592000 CACHED_SPONSORS_LIST %}
 
 {% if logo_place == "download" %}
   <h2 class="widget-title">Sponsors</h2>
@@ -51,3 +55,4 @@
   {% endif %}
   {% endfor %}
 {% endif %}
+{% endcache CACHED_SPONSORS_LIST %}
--- a/templates/sponsors/partials/sponsors-list.html
+++ b/templates/sponsors/partials/sponsors-list.html
@@ -1,8 +1,8 @@
 {% load thumbnail %}
 {% load cache %}
 
-{% comment %}30-days cache{% endcomment %}
-{% cache 2592000 CACHED_SPONSORS_LIST %}
+{% comment %}cache for 1 day{% endcomment %}
+{% cache 86400 CACHED_SPONSORS_LIST %}
 
 {% if logo_place == "download" %}
   <h2 class="widget-title">Sponsors</h2>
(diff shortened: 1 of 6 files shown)


berinhard:
@ewdurbin I updated this PR to draft just to prevent the merge before a more detailed discussion on how to configure Django's cache. Let me know what are your thoughts on that.

templates/sponsors/partials/sponsors-list.html
```diff
@@ -1,4 +1,8 @@
 {% load thumbnail %}
+{% load cache %}
+
+{% comment %}30-days cache{% endcomment %}
+{% cache 2592000 CACHED_SPONSORS_LIST %}
```
ewdurbin:
```suggestion
{% cache 86400 CACHED_SPONSORS_LIST %}
```

templates/sponsors/partials/sponsors-list.html
```diff
@@ -1,4 +1,8 @@
 {% load thumbnail %}
+{% load cache %}
+
+{% comment %}30-days cache{% endcomment %}
```
ewdurbin:
```suggestion
{% comment %}cache for 1 day{% endcomment %}
```

ewdurbin:
Was hoping to use Redis as a cache, but that isn't available until Django 4.0. I'm hoping that this cache is fast enough  to improve the solr-thumbnail situation... for now let's roll it out and asses.

Otherwise this is a great start!

PR #1953, opened 2022-01-04
New admin fieldset section to display user customizations on packages benefits

berinhard:
This PR updates the `SponsorshipAdmin` edit page to have a new read-only section listing user customization on sponsorships. The new section has 2 fields, one to list added benefits and another for the removed ones. By default the section is collapsed, but it appears as a regular one if there's any customization to list.

Screenshot from 2022-01-04 15-05-07

It also gives more importance to the `For Modified Package` flag by making it the first field displayed in the page.

Screenshot from 2022-01-04 15-00-15

### DIFF
--- a/sponsors/admin.py
+++ b/sponsors/admin.py
@@ -289,6 +289,16 @@ class SponsorshipAdmin(admin.ModelAdmin):
         ),
     ]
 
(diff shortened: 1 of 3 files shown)


ewdurbin (approved):
this is _fantastic_

PR #1954, opened 2022-01-05
update salt url to https://saltproject.io/

thbde:
The old url forwards to a page on vmware who bought salt. That page is just some kind of a press release statement and not helpful for the target audience of the information presented here (people that want to learn about great python projects).


thbde:
It may be nice if the list of urls is structured per line to support git diff and blame. But I do not want to make such a big change here.

ewdurbin (approved):

PR #1956, opened 2022-01-05
Improvements to sponsorship form layout

berinhard:
This PR introduces updates to the benefits form from the sponsorship application process. The changes are on the package buttons and on benefits sections as well.

The buttons were updated to look more like as clickable and also the radio input is hidden. Here's how they are looking:

**No selection**
Screenshot from 2022-01-05 15-09-34

**Selected**
Screenshot from 2022-01-05 15-09-47

**On mobile**
Screenshot from 2022-01-05 15-16-39

The other change is a new behavior in the form. Since it is a big form, this PR now introduces the option to collapse/expand sections. Here's how they're looking like now:

Screenshot from 2022-01-05 16-12-31

### DIFF
--- a/static/js/sponsors/applicationForm.js
+++ b/static/js/sponsors/applicationForm.js
@@ -1,3 +1,5 @@
+const DESKTOP_WIDTH_LIMIT = 1500;
+
 $(document).ready(function(){
   const SELECTORS = {
     packageInput:  function() { return $("input[name=package]"); },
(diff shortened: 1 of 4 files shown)


ewdurbin:
Screen Shot 2022-01-07 at 1 55 51 PM

Looks like there are some minor overflows on some sizing.

Also I noted that when selecting a new package, it auto expands all sections that were previously collapsed. If there were some way to keep expand/collapse state when switching packages that might be less jarring for users.

ewdurbin:
@berinhard I pushed a commit to revert the transition point back to 1200px, please take a look. it slightly decreases the font size of the package name as it approaches the breakpoint in order to retain the rather reasonable 1200px width. I fear 1500px might be a little wide.

My sass-compiler was spitting out a huge swath of changes, so I have not included compiled css with the push.

If you're OK with the proposed changes, I think we're all set here!

berinhard:
@ewdurbin yep, the font-size you propose works smoothly. I added 2 new commits, one for a missing line on CSS and the other to roll back to 1200px on the final CSS media queries.

ewdurbin:
Thanks @berinhard!

PR #1959, opened 2022-01-10
Sponsor benefit provided assets

ewdurbin:
WIP, for initial review by @berinhard. Creates the ability to configure "Provided Assets" associated with a sponsorship that will be fulfilled by the PSF.

The initial "ProvidedTextAsset" is intended to be used for PyCon US 2022 voucher codes, which will be unique to each voucher "type" and sponsorship level.

Additionally, we will likely include "ProvidedFileAsset" that will be used for all sponsorships with Expo Hall benefits to share with them a common "Exhibitor Packet"


berinhard (commented):
@ewdurbin so far the code seems fine to me. I tested it locally and admin seems to be working as expected. I tested the process locally and the user sponsorship dashboard got updated with the provided assets. But the form are the next steps, right?

PR #1978, opened 2022-02-09
Fix bug that prevents users from deleting sponsorship applications

### LINKED ISSUE #1977: Sponsorship Admin: Unable to delete Sponsorship objects.
**Describe the bug**
PSF Staff members are unable to delete duplicate Sponsorship objects created due to double submit situations.

Initially reported due to permissions error:

```
Deleting the selected sponsorship would result in deleting related objects, but your account doesn't have permission to delete the following types of objects:
  * Asset
```

When deleting in Admin.

That seems to be due to the `has_delete_permission` method on the GenericAssetModelAdmin object, which is resolved via:

```diff
diff --git a/sponsors/admin.py b/sponsors/admin.py
index cb36f24..c2b471e 100644
--- a/sponsors/admin.py
+++ b/sponsors/admin.py
@@ -831,8 +831,11 @@ class GenericAssetModelAdmin(PolymorphicParentModelAdmin):
         classes = self.get_child_models(*args, **kwargs)
         return self.model.objects.select_related("content_type").instance_of(*classes)
 
-    def has_delete_permission(self, *args, **kwargs):
-        return False
+    def get_actions(self, request):
+        actions = super().get_actions(request)
+        if 'delete_selected' in actions:
+            del actions['delete_selected']
+        return actions
 
     def has_add_permission(self, *args, **kwargs):
         return False
```

Which just removes the Delete button in admin.

However after this, for sponsorships that have a benefit that includes logo placement, the failure is:

```
IntegrityError at /admin/sponsors/sponsorship/
update or delete on table "sponsors_benefitfeature" violates foreign key constraint "sponsors_providedfil_benefitfeature_ptr_i_abb6a2b3_fk_sponsors_" on table "sponsors_providedfileasset"
DETAIL:  Key (id)=(1188) is still referenced from table "sponsors_providedfileasset".
```

**To Reproduce**
1. Configure a sponsor benefit with various benefit configurations, including a logo placement
2. Create a Sponsorship that includes this benefit
3. From any state (applied/approved/finalized/rejected) attempt to delete the sponsorship object.

berinhard:
Fixes #1977

The cascade deletion was being disabled by [known Django bug](https://code.djangoproject.com/ticket/23076) combined with django-polymorphic. The introduced fix changes the default `base_manager_class` to use a non-polymorphic model manager which ensures cascade deletion as well. This manager is called by the delete process and, thus, now it's possible to delete sponsorships.

Here are screen shots to proof that the deletion operation is working:

Screenshot from 2022-02-09 20-08-40

Screenshot from 2022-02-09 20-08-50

### DIFF
--- a/sponsors/admin.py
+++ b/sponsors/admin.py
@@ -831,8 +831,11 @@ class GenericAssetModelAdmin(PolymorphicParentModelAdmin):
         classes = self.get_child_models(*args, **kwargs)
         return self.model.objects.select_related("content_type").instance_of(*classes)
 
-    def has_delete_permission(self, *args, **kwargs):
-        return False
+    def get_actions(self, request):
+        actions = super().get_actions(request)
+        if 'delete_selected' in actions:
+            del actions['delete_selected']
+        return actions
 
     def has_add_permission(self, *args, **kwargs):
         return False
--- a/sponsors/models/assets.py
+++ b/sponsors/models/assets.py
@@ -10,6 +10,7 @@ from django.db import models
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
 from django.db.models.fields.files import ImageFieldFile, FileField
+from polymorphic.managers import PolymorphicManager
 from polymorphic.models import PolymorphicModel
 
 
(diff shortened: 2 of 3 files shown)


ewdurbin:
Thank you @berinhard!

PR #1981, opened 2022-02-10
Use EthicalAds to place logos under Sponsors page

berinhard:
This PR replaces the current thumbnail solution to place sponsors logo by EthicalAds integration. 

Here's a screen shot of the page working in my local environment.

sponsors_page

One thing to notice is that the logo sizes will differ from production and my local environment. This is because the `ideal_dimension` template filter depends on reading the image's width and height. And, even though I have a local of the sponsors app db, the same is not true for the logos. Thus, there was no way to calculate the ideal dimensions. I added a fallback for such scenario and it already improved the logo sizes, but I know that they will differ when this PR gets deployed.

@ewdurbin another minor issue is the missing O'Reilly logo. This is because the logo file is formatted as EPS and probably EthicalAds had trouble processing such file format. I think it's better for us to convert the image to PNG and update our database instead of delegate the fix. Maybe this is also indicate we're missing some validation on the sponsorship application form to only accept JPG or PNG files as logos. What do you think about this?


ewdurbin (approved):
Thanks @berinhard!

PR #1998, opened 2022-02-25
API endpoint to filter assets by internal name

berinhard:
This PR introduces a new endpoint in the sponsors API to filter assets by their `internal_name`. The endpoint URL is `/api/v2/sponsors/sponsorship-assets/` and it always require a `internal_name` to be passed as a query string parameter. If `internal_name` is missing, the response is a bad request.

By default, the endpoint only lists assets which have a value (text or image) associated to it, but it's possible to list every asset by passing `list_empty=true` within the query string.

Here's a response payload example I got from my local database:

```json
[{
  "internal_name": "pycon_logo",
  "uuid": "55f69c46-8d94-4196-9157-76ed6219b120",
  "value": "/media/sponsors-app-assets/55f69c46-8d94-4196-9157-76ed6219b120.jpeg",
  "content_type": "Sponsor",
  "sponsor": "Globe"
}, {
  "internal_name": "pycon_logo",
  "uuid": "2f05d02c-e78a-4976-8b77-5a665612c776",
  "value": "/media/sponsors-app-assets/2f05d02c-e78a-4976-8b77-5a665612c776.jpeg",
  "content_type": "Sponsor",
  "sponsor": "BerinTech"
}]
```

### DIFF
--- a/sponsors/serializers.py
+++ b/sponsors/serializers.py
@@ -21,10 +21,11 @@ class LogoPlacementSerializer(serializers.Serializer):
 class AssetSerializer(serializers.ModelSerializer):
     content_type = serializers.SerializerMethodField()
     value = serializers.SerializerMethodField()
+    sponsor = serializers.SerializerMethodField()
 
     class Meta:
         model = GenericAsset
-        fields = ["internal_name", "uuid", "value", "content_type"]
+        fields = ["internal_name", "uuid", "value", "content_type", "sponsor"]
 
     def get_content_type(self, asset):
         return asset.content_type.name.title()
@@ -33,6 +34,12 @@ class AssetSerializer(serializers.ModelSerializer):
         if not asset.has_value:
             return ""
         return asset.value if not asset.is_file else asset.value.url
+
+    def get_sponsor(self, asset):
+        if asset.from_sponsorship:
+            return asset.content_object.sponsor.name
+        else:
+            return asset.content_object.name
 
 
 class FilterLogoPlacementsSerializer(serializers.Serializer):
--- a/sponsors/serializers.py
+++ b/sponsors/serializers.py
@@ -1,0 +1,43 @@
+
+from rest_framework import serializers
+from sponsors.models.enums import PublisherChoices, LogoPlacementChoices
+
+class LogoPlacementSerializer(serializers.Serializer):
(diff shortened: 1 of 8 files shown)


berinhard:
@ewdurbin I'm not sure why this PR is breaking on CI. The tests aren't eve running due to an import error. The error is: `ImportError: cannot import name 'getStringIO' from 'reportlab.lib.utils' (/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/reportlab/lib/utils.py)` but even when I run the same command locally, they run just fine. I've already reset my virtualenv to reinstall the dependencies, but they still pass.

Here's a proof that I'm being able to import this function:

```bash
$ pip freeze | grep reportlab
reportlab==3.6.1
$ python -c "from reportlab.lib.utils import getStringIO; print(getStringIO)"
<function getStringIO at 0x7ff979753488>
```

Any ideas on what can be causing the import error?

berinhard:
@ewdurbin the CI should be fixed now. The issue was happening due to the new [reportlab 3.6.7 release](https://pypi.org/project/reportlab/3.6.7/) this last February 18. For some reason, the import only started to break after it. Under versions 3.6.6, I was able to run the tests as expected.

The 3.6.7 version was being installed because the one of the `easy_pdf` dependencies depends on it and was declaring it with `reportlab>=3.3.0`. Thus, I added ab7da79 to force a limit to the version we want to use.

sponsors/serializers.py
```diff
@@ -0,0 +1,82 @@
...
+    end_date = serializers.DateField()
+    sponsor_url = serializers.URLField()
+    level_name = serializers.CharField()
+    level_order = serializers.IntegerField()
+
+
+class AssetSerializer(serializers.ModelSerializer):
+    content_type = serializers.SerializerMethodField()
+    value = serializers.SerializerMethodField()
+    sponsor = serializers.SerializerMethodField()
```
ewdurbin, line 24: can we add the sponsor slug to this as well? I think we're all set then!

ewdurbin (changes requested):

sponsors/serializers.py
berinhard: Done in ebc2227e

ewdurbin (approved):

PR #2006, opened 2022-03-09
phase one of pep 676 finalization, temporary redirects to peps.python.org

ewdurbin:
See: https://peps.python.org/pep-0676/#specification

### DIFF
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -268,6 +268,18 @@ http {
                   return 302 https://www.python.org/dev/peps/;
                 }
 
+                location ~ ^/dev/peps/?(.*)$ {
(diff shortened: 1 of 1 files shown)


config/nginx.conf.erb
```diff
@@ -268,6 +268,18 @@ http {
                   return 302 https://www.python.org/dev/peps/;
                 }
```
merwok, line 270: The redirects above could have been updated too

ewdurbin: Just getting the minimum in place before we transition to permanent redirects. Once this has been out in the wild for a bit, I'll simplify the whole setup.

Also plan to add a Surrogate-Key so we can purge all the redirects being dispatched from Nginx.

PR #2055, opened 2022-06-02
Point issue tracker links to GitHub (closes #2054)

### LINKED ISSUE #2054: python.org points to bugs.python.org (instead of GitHub) for issues references
**Describe the bug**
References to the issue tracker on python.org still point to the old tracker (bugs.python.org), e.g. [`Contributing` in the site footer](https://github.com/python/pythondotorg/blob/main/fixtures/sitetree_menus.json#L2344) and the [`weneedyou` widget](https://github.com/python/pythondotorg/blob/211edf68a3f70002be2dd009c7a6f7d5df62e91c/fixtures/boxes.json#L153) on the `Documentation` page.

I know that there is a notice at the top of the old tracker pointing users to the GitHub tracker, but it seems like an unwanted layer of indirection now that the old tracker is read-only and all the issues are migrated to GH anyway.

Will open a PR to change these references, but please close with prejudice if there is some plan to eventually point `bugs.python.org` to the current tracker :sweat_smile:

SnoopJ:


Mariatta (approved):
Thank you.

zware (approved):
There are still several other references to `bugs.python.org` throughout the repository, but this covers the most user-visible ones.

PR #2080, opened 2022-07-21
add waffle flag to hide sponsorship menu and application

ewdurbin:
During periods where we are modifying the benefits for a given year, we don't want to display interim states.

This flag will allow us to hide the menu and application form during those periods, while still allowing staff to view/preview that page.

Also addresses the white text in the submit button noted by @loren-c


ewdurbin:
@JackieAugustine @loren-c This functionality will be controllable via https://www.python.org/admin/waffle/flag/2/change/

ewdurbin:
Mock up of when the flag is set to "off":
Untitled

PR #2087, opened 2022-07-30
Keep track of current sponsorship year

### LINKED ISSUE #2084: Add some way of tracking "Sponsorship Year" to Sponsorship objects.
The sponsorship team has asked for some way to organize Sponsorships and benefits by the sponsorship year, which is independent of calendar year.

This will allow for filtering views and differentiating renewals.

Considerations:

- Create a singleton object with “Current Application Year” for self-service in Django admin
- Compose the configured Packages, Benefits, and Application Year when a new application is created, copy Year into the sponsorship object
- Access “Year” from related objects via the Sponsorship
- Add Admin filters on views for Sponsorships, Benefits, etc on this value.
- Should probably be a real life “Year” field not just an integer or string?

berinhard:
Fixes #2084

This PR introduces a new singleton model in the sponsors app called `SponsorshipCurrentYear`. This model is used to keep rack of the current application year. 

I wanted to not use Django to manage the singleton because I felt it would make the Python code more complex. Instead, I left this responsibility to Postgres. I did so by adding an unique index on a constant value in this new table. By doing that, the only changes required in Django were to disable deletion and addition for this new model. Even though the singleton is implemented in Postgres, this PR adds Django unit tests to ensure it's working. This PR also introduces an initial data migration to populate the singleton and to set the year to 2023.

Then, a new `year` indexed column was added to the following models: `Sponsorship`, `SponsorshipPackage` and `SponsorshipBenefit`. The first one represents the sponsorship application itself, while the others are used as configuration models used by application to properly populate its related objects. The year of the sponsorship gets populated when it gets created using the value from `SponsorshipCurrentYear.get_year`. The year for the configuration models can be edit via Django admin, but this PR introduces an initial migration to populate them with 2023.

The queries to list packages, their benefits, add-ons and a la carte benefits in the sponsorship application form were updated to only return configured objects from the current year. That way, we can customize the whole sponsorship application form by year.

It's worthwhile to mention that the `SponsorshipCurrentYear.get_year` has a cache policy to avoid repeated and unnecessary db hits to fetch for the year value. The cached value, when populated, is set to never expire. But, when the year gets updated in the database, the `save` method invalidates the existing cache. So, that way, the cache will get populated again with the new value once `SponsorshipCurrentYear.get_year` is called again.


berinhard:
@ewdurbin I didn't manage to test this PR via the web interface. I still couldn't load the data for the sponsors app using the dump JSON files you sent me, but we can talk about this topic next Monday.

ewdurbin:
I like that the packages/benefit configurations are now "versioned" by year! I think this is a solid move.

This will allow us theoretically setup a new sponsorship year before setting it to "Live" by updating the singleton.

I think the flow might look something like:

  - Sponsorship coordinator initiates a "clone" of current sponsorship year into a new named sponsorship year from Django Admin (i.e. "Clone 2022 sponsorship year packages, benefits, and benefit configurations into a 2023 sponsorship year")
  - Application creates new Packages, Levels, Benefits, and BenefitConfigurations etc as needed and associates the new sponsorship year.
  - At a later date, Sponsorship coordinator returns and sets the singleton to the new sponsorship year (i.e. 2023). This "launches" the new program on the form and all new sponsorship applications and associated objects are created with that year.


Thoughts:
  - Sponsorship coordinators should be able to view the new packages/benefits and such via the form _before_ launch. This should be on a route that requires superuser or some specific permission.

ewdurbin:
Fixtures added including `user` and `admin` users for logging in (with password `password`).

berinhard:
@ewdurbin I guess now we have the whole workflow implemented. I managed to test the whole process of cloning an sponsorship application, updating the new year's benefits info, previewing it and everything work smoothly. I'm adding a few screen shots here to guide you through you own QA:

## 1 - Clone by year

You can access the following page via admin going to Sponsors > Active Yer and then clicking on the button "Clone Application Configuraton". 

Screenshot from 2022-08-04 17-55-47

After submiting the form, the same page will be displayed for you with the links to the benefits and packages from the new configured year in the list of already configured years:

Screenshot from 2022-08-04 17-55-43

## 2 - Customizing the configuration for the year

You can change benefits and package information for that year. You can access them directly from the filter URLs displayed in the Active Year as the screen shot bellow shows:


Screenshot from 2022-08-04 20-00-23

## 3 - Previewing the form from a specific year

As you can see, the previous page also lists a link to preview the application form for those years. I'm using the same URL as the current one, but with an extra querystring parameter called `config_year` that only gets evaluated if the user is a staff one. Also, as you can see bellow, I added a warning message so the staff user can now it's a preview mode:

Screenshot from 2022-08-04 19-35-33

ewdurbin:
This is stellar work @berinhard I pushed one minor tweak to make the preview a little more obvious.

Otherwise, I was able to test this with a full prod dataset for sponsorship and it is 💯

PR #2111, opened 2022-08-08
Add some assitive ordering and reprs for objects related to Sponsorship years in Django Admin.

ewdurbin:
Noted while walking through the new features with @loren-c.


ewdurbin:
cc @berinhard.

PR #2112, opened 2022-08-09
Sponsorship: make year a dropdown for sponsorship admin forms that need it

ewdurbin:


ewdurbin:
cc @berinhard, this isn't _ideal_ lets discuss next call if there's something we can do to improve.

PR #2113, opened 2022-08-09
Add support for Sigstore verification materials

di:
This PR adds support for including the files that comprise Sigstore verification materials with CPython downloads. This is one of the last steps in the [plan documented here](https://docs.google.com/document/d/1_y0O7mbv4TVdw3rx9tUCrIZItA9GGOA-RNH_qQW1JS8/edit?resourcekey=0-vsbpfJONHGpYf3AfytbD8g).

~This also fixes some small issues I ran into during the development process.~ moved to #2114 

(cc @ewdurbin @pablogsal @Yhg1s)


templates/humans.txt
```diff
@@ -59,5 +59,5 @@
 	Core: Python 3 and Django 1.7
 	Components: Modernizr, jQuery, Susy (susy.oddbird.net)
 	Software: SASS and Compass, Coda, Sublime Text, Terminal, Adobe CS, Made on Macs
-	Hardware Stack: Ubuntu 14.04, Postgresql 9.x, Nginx, uwsgi
+	Hardware Stack: Ubuntu 14.04, Postgresql 10.x, Nginx, uwsgi
```
ewdurbin: nit: we're also on heroku now and use gunicorn 🙃

sponsors/migrations/0086_auto_20220809_1655.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.2.24 on 2022-08-09 16:55
```
ewdurbin: exclude this file for now please, I'll add a followup PR to include this dangling migration.

templates/downloads/release_detail.html
```diff
@@ -49,6 +56,9 @@ <h1 class="page-title">Files</h1>
               <th>MD5 Sum</th>
               <th>File Size</th>
               <th>GPG</th>
+              {% if has_sigstore_materials %}
+              <th colspan="2">Sigstore</th>
```
ewdurbin: Would it be a bad idea to link from this to some reference on what sigstore is?

templates/downloads/release_detail.html
```diff
@@ -40,6 +40,13 @@ <h1 class="page-title">{{ release.name }}</h1>
             <h1 class="page-title">Files</h1>
         </header>
 
+        {% set has_sigstore_materials = false %}
```
ewdurbin: I don't think this is valid for Django templates:

```
django.template.exceptions.TemplateSyntaxError: Invalid block tag on line 43: 'set', expected 'endblock'. Did you forget to register or load this tag?
```

base-requirements.txt
```diff
@@ -6,7 +6,7 @@ docutils==0.12
 Markdown==3.3.4
 cmarkgfm==0.6.0
 Pillow==8.3.1
-psycopg2==2.8.6
+psycopg2-binary==2.8.6
```
ewdurbin: is this strictly necessary? could it be addressed in a separate hygiene PR?

.python-version
```diff
@@ -0,0 +1 @@
+3.9.6
```
ewdurbin: is this strictly necessary? could it be addressed in a separate hygiene PR?

templates/downloads/release_detail.html
di: That's the plan! I'll get that page up before we merge this, and add the link here.

base-requirements.txt
ewdurbin: Just a note on why: I'm not fully certain how this affects deployment pipeline and don't want it to impact the ability to safely merge this.

.python-version
di: Moved this to https://github.com/python/pythondotorg/pull/2114

base-requirements.txt
di: Moves this to https://github.com/python/pythondotorg/pull/2114

sponsors/migrations/0086_auto_20220809_1655.py
di: Moved this to https://github.com/python/pythondotorg/pull/2114.

ewdurbin (approved):

ewdurbin:
cc @python/python-release-managers, take a look if you're able before I merge this tomorrow.

pablogsal:
Will this need changes to our upload script (https://github.com/python/release-tools/blob/master/add-to-pydotorg.py) ?

di:
@pablogsal Looks like yes, I was unaware of that script, will make a PR to update it tomorrow.

di:
(shouldn't block this PR though I think)

di:
@pablogsal PR here: https://github.com/python/release-tools/pull/17

ewdurbin:
Was it determined that this won't break the existing release script? If so I'm good to merge!

di:
Yes, the changes to the release script are to include the verification materials if they're present. As is, it just won't include them, but won't break.

pablogsal:
@di Seems that we need some extra changes because after signing 3.10.7 with sigstore we ran into some problems:

https://www.python.org/downloads/release/python-3107/

If you click the sigstore cert it shows 403 Forbidden.

Also, unfortunately there was a NameError in the original changes for the release scripts:

https://github.com/python/release-tools/pull/19

Also, seems that the automation is broken now because sigstore blocks until you paste some code from the auth and therefore this code here:

https://github.com/python/release-tools/blob/401afdec47eadf4b5f17a68f5bb87f4ca04a86d1/run_release.py#L720-L744

Doesn't work anymore.  Sadly I will need to comment out the signing in the `add_to_python_org.py` script until that is fixed.

ewdurbin:
Are the 403ing files on disk on the downloads backend host? Seems they must be being created with strict permissions?

pablogsal:
Yeah, they are on disk. Seems that the changes in the release tools so far are not enough.  The release script changes the permissions for the artifacts here:

https://github.com/python/release-tools/blob/401afdec47eadf4b5f17a68f5bb87f4ca04a86d1/run_release.py#L574-L578

So we need changes to the `add_to_python_dot_org.py` script to correctly change the permissions as well.

pablogsal:
Given that this is impacting releases, I am reverting the sigstore changes in the repository until they are ready.

pablogsal:
I also fixed manually the permissions for 3.10.7

di:
Thanks, I filed https://github.com/python/release-tools/issues/21 to capture what (I think) needs done to fix this.

PR #2120, opened 2022-08-13
Update benefits validation rules

### LINKED ISSUE #2117: Sponsorship Application Form: Enforce new rules regarding non-package benefits
Depends on #2116 for implementation and terminology used below.

- a la carte benefits should only be able to be applied for if a package is selected
- applications for standalone benefits should be mutually exclusive with applications for packages
- combining the two above rules, this means that applications for standalone benefits should not be able to include a la carte benefits

Additionally, packages should be configurable as to wether they allow a la carte benefits from section 2 of the form. This should also control as to wether the "potential add-on" in the benefits grid is shown.

These rules should be enforced in form validation on submission as well as represented in the UI.

Example: Associate package for the current sponsorship application will not be allowed to include any benefits not configured as part of the package in the backend.

Thus, the following selection should not be shown or enabled:

grid

Additionally if the Associate package is selected the following two sections should not be enabled for selection, and any existing selections should be cleared.

after-sell

berinhard:
Fixes https://github.com/python/pythondotorg/issues/2117

This PR updates both the form validation and the frontend UX to enforce the new rules on non-package benefits.


ewdurbin (changes requested):
@berinhard I added a few small tweaks to help clarify and ease the UX a bit.

Only outstanding issues I have before this is ready to merge:

1) I think we need a "clear selection" button that discards all state, once I've selected a package I am not able to get back to state that allows me to change my mind and apply for a stand alone only
2) When initially loading the state of the page, the a la carte and standalone boxes are enabled, even if the package my session currently has selected doesn't allow them. this lets me select add-ons or stand alone and hit submit (obviously getting a backend validation). we should run the appropriate code to disable/enable these selections based on the package selection stored in session.

otherwise, clean and tidy. feel free to discard or rework the 3 commits I added.

berinhard:
@ewdurbin I pushed the fix for the item number 2 from you comment and the styling part of the clear form button. I'll finish with the remaining JS part by tomorrow.

berinhard:
@ewdurbin the JS part to clean the form is ready and is working considering the scenarios I tested on. Feel free to change anything on the behavior or let me know if other improvements are necessary. But, as far as we talked, this PR is ready for a final review and merge.

ewdurbin (approved):
Thanks @berinhard! This new form functionality is going to be very helpful for the sponsorship team!

PR #2122, opened 2022-08-16
Sponsorship form: Do not display benefits marked as unavailable in admin

ewdurbin:
Sponsorship team was trying to use this to hide benefits that are not to be available at launch.

Turns out we didn't actually do anything with that field. This uses it in the managers to filter out unavailable benefits for display.


berinhard (commented):
The PR looks good :+1:  Thanks for the quick fix!

PR #2130, opened 2022-08-24
Dockerizing the pythondotorg codebase:

cegerhardson:
This commit will add the appropriate Dockerfile and docker-compose.yml file to begin the process of dockerizing the pythondotorg codebase.

Next steps include completeing a Makefile.


docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+      POSTGRES_PASSWORD: pythondotorg
+      POSTGRES_DB: pythondotorg
+      POSTGRES_HOST_AUTH_METHOD: trust  # never do this in production!
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "pythondotorg", "-d", "pythondotorg"]
+      interval: 1s
+
+  web:
+    build: .
+    command: sh dev/wait-for.sh postgres:5432 -- dev/docker-entrypoint.sh python manage.py runserver 0.0.0.0:8000
```
ewdurbin: Ah! So the `wait-for.sh` and `entrypoint.sh` scripts won't be necessary for this! We're replacing `wait-for.sh` with the health check configured for Postgres and the depends_on configuration for web.

Additionally we'll discuss how we're replacing the entrypoint  with Makefile targets as well.

The command we run for the web container can probably just be `python manage.py runserver 0.0.0.0:8000`.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+	@$(MAKE) -pRrq -f $(lastword $(MAKEFILE_LIST)) : 2>/dev/null\
+        | awk -v RS= -F: '/^# File/,/^# Finished Make data base/ {if ($$1 !~ "^[#.]") {print $$1}}'\
+        | sort\
+        | egrep -v -e '^[^[:alnum:]]' -e '^$@$$'
+	@echo
+	@exit 1
+
+.state/docker-build-web: Dockerfile dev-requirements.txt base-requirements.txt
+	#Build web container for this project 
+	docker-compose build --build-arg IPYTHON=$(IPYTHON) --force-rm web
```
ewdurbin: I think the `--build-arg IPYTHON=$(IPYTHON)` here is extraneous as we don't have any build args for our Dockerfile.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+	docker-compose run --rm web ./manage.py migrate 
+	docker-compose run --rm web ./manage.py loaddata fixtures/*.json
+
+	#Mark the state so we don't rebuild this needlessly.
+	mkdir -p .state
+	touch .state/db-initialized
+
+
+serve: .state/db-initialized
+	docker-compose up -d
```
ewdurbin: `-d` here starts the containers in the background, which can mean its hard to keep track of what's going on. I think we should probably start them in the foreground by removing that argument.

ewdurbin: I'd suggest the `--remove-orphans` flag here to ensure that only containers defined in the service remain up. This is helpful if people start running arbitrary containers to test or modify the environment.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+
+clean:
+	docker container rm $$(docker ps -aq) -f
```
ewdurbin: This will work, but will also impact _all_ docker containers on the host. Using `docker-compose down -v` will ensure that we only take down/clean up the containers relevant to this codebase.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+	rm -f .state/*
```
ewdurbin: We might want to remove each state file independently in the event that the `.state` directory gets used for anything else.

ewdurbin (changes requested):
Overall this is working. Suggesting some refinements that will improve the overall experience and keep this from over-impacting the host environment.

ewdurbin (commented):
This is working splendidly! Thank you!

Three things I think we should add to wrap this up, that represent common developer patterns:

- `make migrations` which should call the Django `makemigrations` management command.
- `make migrate` which should call the Django `migrate` command independently of the db initialization, we should refactor the initialization target to run this instead of duplicating the call to `manage.py migrate` between them.
- `make shell` which drops a developer into a Django shell via the `shell` management command.

As a stretch goal, a `make manage` target that accepts arbitrary arguments after the manage and runs `manage.py` inside the container with those arguments would be very helpful!

Makefile
```diff
@@ -0,0 +1,52 @@
...
+	@echo
+	@$(MAKE) -pRrq -f $(lastword $(MAKEFILE_LIST)) : 2>/dev/null\
+        | awk -v RS= -F: '/^# File/,/^# Finished Make data base/ {if ($$1 !~ "^[#.]") {print $$1}}'\
+        | sort\
+        | egrep -v -e '^[^[:alnum:]]' -e '^$@$$'
+	@echo
+	@exit 1
+
+.state/docker-build-web: Dockerfile dev-requirements.txt base-requirements.txt
+	#Build web container for this project
```
ewdurbin: Small nit: Convention for comments is to have one trailing space after the `#` like `# This is a comment`. Makes it a little easier to read.

Makefile
```diff
@@ -0,0 +1,52 @@
...
+	#Load all fixtures 
+	docker-compose run --rm web ./manage.py loaddata fixtures/*.json
+
+	#Mark the state so we don't rebuild this needlessly.
+	mkdir -p .state && touch .state/db-initialized
+
+serve: .state/db-initialized
+	docker-compose up --remove-orphans
+
+migrations: .state/docker-build-web
```
ewdurbin: We want an initialized database before being able to create new migrations, so this should have `.state/db-initialized` as a requisite, not just a built docker image.

Makefile
```diff
@@ -0,0 +1,52 @@
...
+	
+migrate: .state/docker-build-web
+	#Run Django migrate 
+	docker-compose run --rm web ./manage.py migrate 
+
+manage: .state/db-initialized
+	#Run Django manage to accept arbitrary arguments
+	docker-compose run --rm web ./manage.py $(filter-out $@,$(MAKECMDGOALS))
+
+shell: .state/docker-build-web
```
ewdurbin: It's probably sufficient to have a built image to open a shell, but I think a fully initialized database is probably better here, so this should have `.state/db-initialized` as a requisite, not just a built docker image.

ewdurbin (approved):
This is great! Thanks @cegerhardson!

ewdurbin:
Let's address the changes noted above and this is good from a technical perspective. We probably want to add some docs to https://github.com/python/pythondotorg/blob/main/docs/source/install.rst with how to get setup using docker/docker-compose.

docs/source/install.md
```diff
@@ -1,6 +1,69 @@
...
+
+Install [Docker Engine](https://docs.docker.com/engine/install/) 
+
+```{note}
+The best experience for building Pythondotorg on Windows is to use the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/)(WSL) in combination with both [Docker for Windows](https://docs.docker.com/desktop/install/windows-install/) and [Docker for Linux](https://docs.docker.com/engine/install/).
+```
+
+Verify that the Docker installation is successful by running: `docker -v`
+
+Building the Pythondotorg Container
```
ewdurbin: I think this could be renamed "Running pythondotorg locally".

docs/source/install.md
```diff
@@ -1,6 +1,69 @@
...
+-----------------------------------
+Once you have Docker and Docker Compose installed, run:
+
+```
+make serve
+```
+
+This will pull down all the required docker containers, build Pythondotorg and establish all of the neccessary services.
```
ewdurbin: It might be good to show the expected output ala:

```
web_1       | Starting development server at http://0.0.0.0:8000/
web_1       | Quit the server with CONTROL-C.
```

And also instruct people to view the result in their local web browser at `http://localhost:8000`

docs/source/install.md
```diff
@@ -1,6 +1,69 @@
...
+
+This will pull down all the required docker containers, build Pythondotorg and establish all of the neccessary services. 
+
+To reset your local environment, run:
+
+```
+make clean
+```
+
+To create or apply new migrations, run:
```
ewdurbin: This will only apply migrations, we need a separate note for creating new ones `make migrations`.

docs/source/install.md
```diff
@@ -0,0 +1,230 @@
...
+
+Running pythondotorg locally 
+----------------------------
+Once you have Docker and Docker Compose installed, run:
+
+```
+make serve
+```
+
+This will pull down all the required docker containers, build Pythondotorg and establish all of the neccessary services.
```
ewdurbin:
```suggestion
This will pull down all the required docker containers, build the environment for pythondotorg, run migrations, load development fixtures, and start all of the necessary services. 
```

docs/source/install.md
```diff
@@ -0,0 +1,230 @@
...
+----------------------------
+Once you have Docker and Docker Compose installed, run:
+
+```
+make serve
+```
+
+This will pull down all the required docker containers, build Pythondotorg and establish all of the neccessary services. 
+
+```
```
ewdurbin, line 53:
````suggestion
Once complete, you will see the following in your terminal output:

```
````

File: docs/source/install.md [hunk 14]
```diff
@@ -0,0 +1,230 @@
...
+
+You can view these results in your local web browser at: `http://localhost:8000`
+
+To reset your local environment, run:
+
+```
+make clean
+```
+
+To apply new migrations, run:
```
ewdurbin:
```suggestion
To apply migrations, run: 
```

docs/source/install.md
```diff
@@ -0,0 +1,230 @@
...
+make clean
+```
+
+To apply new migrations, run: 
+
+``` 
+make migrate
+```
+
+To create new migrations, run:
```
ewdurbin:
```suggestion
To generate new migrations, run: 
```

docs/source/install.md
```diff
@@ -0,0 +1,230 @@
...
+``` 
+make migrate
+```
+
+To create new migrations, run: 
+
+```
+make migrations
+```
+
```
ewdurbin, line 77:
````suggestion

You can also run arbitrary Django management commands via:

```
make manage <NAME_OF_COMMAND>
```

This is a simple wrapper around running `python manage.py` in the container, all arguments passed to `make manage` will be passed through.

````

ewdurbin (approved):
All good! Some final tweaks added via suggestion.

ewdurbin:
Thank you @cegerhardson!

PR #2137, opened 2022-09-07
Downloads: Enforce constraint that only one ReleaseFile per OS per Release has "Download button" enabled.

ewdurbin:
Our most favorite (and common) release failure!

This enforces the constraint on the DB level and raises a ValidationError if attempting to make a change that would breach it.

Before, RMs might accidentally mark multiple ReleaseFile objects as the "Download" for a given OS on a Release. This breaks the rendering of the downloads page for that release (and the main downloads page if it's the current release!) due to the [Django "get" query here](https://github.com/python/pythondotorg/blob/a886cf8a25ac735b2aebc62659aed74226dcaa65/downloads/models.py#L107-L112).


ewdurbin:
Please take a look @python/python-release-managers

ewdurbin:
More specifically @python/release-managers-in-development-maintenance-and-security-mode

ned-deily:
I took a look at it and, while I'm not prepared to try to test it ahead of time, it looks good to me in principle. I'd recommend that you deploy it soon and let us know so we can test it live and then make any adjustments prior to the next set of releases. Thanks for tackling this, @ewdurbin! We've stumbled across this once too often.

ewdurbin:
This constraint is now live @python/release-managers-in-development-maintenance-and-security-mode please check your tooling. In particular it appears that the recent Windows release for `Python 3.11.0rc2` had both the 64bit and ARM installers marked as downloadable.

PR #2182, opened 2022-11-02
configure cmarkgfm to make cms pages more flexible

### LINKED ISSUE #2181: Add a "Markdown with HTML" option to the CMS pages
The current CMS uses https://pypi.org/project/cmarkgfm/ for rendering Markdown syntax when selecting "Markdown" in the CMS.

This defaults to safe rendering of the given Markdown, which is fine. However, in some cases, it may be useful to e.g. make use of the available JQuery UI support for showing more advanced widgets such as accordions.

For the PSF Trademark FAQ page, it would be useful to have an option to add the needed raw HTML directly to the Markdown content (unless there's some other way to get this working): https://www.python.org/psf/trademarks-faq/ The page is becoming rather long and an accordion would help make navigation easier.

Adding support should be easy: simply add another tuple to https://github.com/python/pythondotorg/blob/main/pages/models.py#L51 using a lamdba to enable the `cmarkgfm.cmark.CMARK_OPT_UNSAFE` option. I can give this a try, if there's a chance to get this accepted.

ewdurbin:
closes #2181

### DIFF
--- a/pages/models.py
+++ b/pages/models.py
@@ -21,6 +21,7 @@ from markupfield.fields import MarkupField
 from markupfield.markup import DEFAULT_MARKUP_TYPES
 
 import cmarkgfm
+from cmarkgfm.cmark import Options as cmarkgfmOptions
 
 from cms.models import ContentManageable
 from fastly.utils import purge_url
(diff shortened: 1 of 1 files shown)


malemburg:
Looks good, but this will enable unsafe markdown on all pages. Do you think that's a good idea ?

My idea was to only add this as an extra option in the CMS.

ewdurbin:
Fair point. Updated.

malemburg (approved):
LGTM. Editors of CMS pages will know what the risks are, I suppose, so perhaps a new option is not needed.

malemburg (approved):
Ok, this is even better. Thanks :-)

malemburg:
Thanks for implementing this, Ee.

How long will we have to wait before this goes live ?

ewdurbin:
looks like you were able to test it in staging based on issue discussion, it won't go to prod until its merged into the `release` branch, but seems you may have more to look into.

PR #2183, opened 2022-11-02
Use a custom version of the Github MD formatter

### LINKED ISSUE #2181: Add a "Markdown with HTML" option to the CMS pages

malemburg:
This doesn't filter away the script (and some other) HTML tags.

Fixes #2181.

### DIFF
--- a/pages/models.py
+++ b/pages/models.py
@@ -32,7 +32,8 @@ DEFAULT_MARKUP_TYPE = getattr(settings, 'DEFAULT_MARKUP_TYPE', 'restructuredtext
 
 # Set options for cmarkgfm for "unsafe" renderer, see https://github.com/theacodes/cmarkgfm#advanced-usage
 CMARKGFM_UNSAFE_OPTIONS = (
-    cmarkgfmOptions.CMARK_OPT_UNSAFE
+    cmarkgfmOptions.CMARK_OPT_UNSAFE |
+    cmarkgfmOptions.CMARK_OPT_GITHUB_PRE_LANG
 )
 
 PAGE_PATH_RE = re.compile(r"""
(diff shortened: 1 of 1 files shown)


pages/models.py
```diff
@@ -32,7 +32,8 @@
 
 # Set options for cmarkgfm for "unsafe" renderer, see https://github.com/theacodes/cmarkgfm#advanced-usage
 CMARKGFM_UNSAFE_OPTIONS = (
```
ewdurbin: Can we just move these options into the `unsafe_markdown_to_html` function now that they're not being shoveled into a lambda? I only made them a constant to make them more clear in the lambda definition.

pages/models.py
```diff
@@ -65,10 +66,29 @@
...
+        extensions=[
+            'table', 'autolink', # 'tagfilter',
+            'strikethrough', 'tasklist'
+        ])
+
 RENDERERS.append(
     (
         "markdown_unsafe",
-        lambda markdown_text: cmarkgfm.github_flavored_markdown_to_html(markdown_text, options=CMARKGFM_UNSAFE_OPTIONS),
+        lambda markdown_text: unsafe_markdown_to_html(markdown_text),
```
ewdurbin: this no longer needs to be a lambda.

ewdurbin (changes requested):

malemburg:
Implemented both change requests.

ewdurbin (approved):

Issue #2190, opened 2022-11-10
Website home page, in the code example 5(Quick & Easy to Learn) has SyntaxError in first line at https://www.python.org/ .


ozgurturkiye:
**Describe the bug**
Website home page, in the code example 5(Quick & Easy to Learn) has SyntaxError in first line at https://www.python.org/ .
```python
# Simple output (with Unicode)
>>> print(\"Hello, I'm Python!\")

print(\"Hello, I'm Python!\")
           ^
SyntaxError: unexpected character after line continuation character
```

**Screenshots**
<img width="543" alt="Screenshot 2022-11-10 at 1 59 00 PM" src="https://user-images.githubusercontent.com/8600407/201183516-d3e35b58-a58c-46f0-9f44-a5483cadce8f.png">

hugovk:
This was fixed by #2191:

<img width="488" alt="image" src="https://github.com/user-attachments/assets/4dec14bf-6a3d-45d4-9f73-2f6afb4d234f">

```pycon
Python 3.13.0rc2 (v3.13.0rc2:ec610069637, Sep  6 2024, 17:57:29) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> print("Hello, I'm Python!")
Hello, I'm Python!
>>>
```

PR #2191, opened 2022-11-11
Fixed Issue#2190 | Homepage code syntax error

eshaanmandal:
Fixed issue #2190. Website home page, in the code example 5(Quick & Easy to Learn) has SyntaxError in first line at https://www.python.org/

### DIFF
--- a/codesamples/factories.py
+++ b/codesamples/factories.py
@@ -31,7 +31,7 @@ def initial_data():
             <span class=\"output\">Hello, I'm Python!</span>
 
             <span class=\"comment\"># Input, assignment</span>
-            >>> name = input('What is your name?\\n')
+            >>> name = input('What is your name?\n')
             <span class=\"output\">What is your name?
             Python</span>
             >>> print(f'Hi, {name}.')
--- a/codesamples/factories.py
+++ b/codesamples/factories.py
@@ -27,7 +27,7 @@ def initial_data():
         (
             """\
             <pre><code><span class=\"comment\"># Simple output (with Unicode)</span>
-            >>> print(\"Hello, I'm Python!\")
+            >>> print("Hello, I'm Python!")
             <span class=\"output\">Hello, I'm Python!</span>
 
             <span class=\"comment\"># Input, assignment</span>


eshaanmandal:
Hi,
I think I have fixed all the issues with my last pull request. Please review my changes as a first-time contributor.
Thank you.

ozgurturkiye (approved):

jbw3:
FYI: I noticed that the next line of code in the example also has an extra backslash.
Instead of this...
```py
name = input('What is your name?\\n')
```
...it should be this...
```py
name = input('What is your name?\n')
```

ozgurturkiye (approved):

codesamples/factories.py
```diff
@@ -27,11 +27,11 @@ def initial_data():
         (
             """\
             <pre><code><span class=\"comment\"># Simple output (with Unicode)</span>
-            >>> print(\"Hello, I'm Python!\")
+            >>> print("Hello, I'm Python!")
             <span class=\"output\">Hello, I'm Python!</span>
```
jayaddison, line 31: Shall we put a unicode snake into the code (and HTML)?

```suggestion
            >>> print("Hello, I'm Python \N{SNAKE}")
            <span class=\"output\">Hello, I'm Python! &#x1f40d;</span>
```

image

jayaddison:
@ewdurbin apologies to ping directly, but I thought that this might be worth a quick fixup.  It relates to a syntax error in one of the `python.org` homepage code examples.

@eshaanmandal's fix should work as-is.  I've added a suggestion that we could additionally use unicode in the example code (to match an existing comment), but I could move my suggestion into a separate follow-up pull request.

(also a quick deployment note/reminder: as you [discovered before](https://github.com/python/pythondotorg/pull/1763#issuecomment-1303777674), these examples are fixture data that are loaded into the db to take effect)

jayaddison:
:tada: nice - problem solved :)

image

(confirmed within a Python REPL session too)

jayaddison:
@ewdurbin ek, sorry - a nitpick about this: although the code syntax for this sample has been fixed on the homepage, some of HTML escaping may have gone a bit wonky, and that results in all of the text appearing in green (instead of a mixture of syntax highlighting colours, as with the other samples)

## Expected syntax highlighting in HTML
image

#### HTML snippet
```html
<span class="comment"># Python 3: Simple arithmetic</span>
```

## Appearance of code sample five
image

#### HTML snippet
```html
<span class="\&quot;comment\&quot;"># Simple output (with Unicode)</span>
```

(some kind of HTML-escaping issue?)

jayaddison:
That looks much better now - thank you!

image

Issue #2102, opened 2022-08-02
Sponsorship Admin Page - Send Notifications/Emails: Add archive option for email templates


oliviasauls:
**Location of change**
python.org/admin/sponsors/sponsoremailnotificationtemplate/ - Admin/Sponsorships/Sponsor Email Notification Templates

**Describe the solution you'd like**
  * Add archive option for email templates
  * Add “hidden” or “archived” field to object, add filter with default on django admin view
     
**Reason for change**
  * Helps sort notifications sent in past years/no longer needed 
  
**Priority**
3. Low

**Date Needed**
Sponsorship Launch 

**Additional context**
<img width="1452" alt="Screen Shot 2022-08-02 at 2 07 02 PM" src="https://user-images.githubusercontent.com/89146495/182454309-f2b83e0f-2291-45f8-9ab0-58a3fd4a1c53.png">

HigorMonteiro:
I going work in this issue

augdiebold:
Working on that one!

Issue #103, opened 2014-02-20
Peculiar layout on options bar at small screen width


linssen:
![screen shot 2014-02-20 at 12 38 23](https://f.cloud.github.com/assets/67624/2218294/18466d74-9a2c-11e3-83ac-b5433909ddfa.png)

adamn:
This is no longer an issue and can be closed.

linssen:
I don't think it's been resolved has it? Screenshot below from today
<img width="496" alt="Screenshot 2020-01-21 at 19 00 05" src="https://user-images.githubusercontent.com/67624/72834255-505d6580-3c80-11ea-856e-8d3bb7a02fc6.png">

PR #2218, opened 2022-12-29
feat(#2102):  Add archive sponsorship email templates option

augdiebold:
Related ISSUE: [#2102](https://github.com/python/pythondotorg/issues/2102) 

### Description
Create a possibility to archive old email templates that won't be useful anymore instead deleting them.

- Created a **boolean** field named 'active' on BaseEmailTemplate model, default=True.
- Added an active filter filter on django sponsor email templates admin page. Default lookup set to "yes".
- Created an admin action to archive multiple templates (set them all active=False)

### Screenshots

image
image
image


JacobCoffee (changes requested):
This is old, sorry about that, but the migrations have advanced quite a bit since this was made (up to #0103 now).

This will need to be rebased on the latest `main` and then the local migration re-ran.

JacobCoffee:
I understand it has been some time, so I'm going to close this for now. However, this is still wanted/needed I imagine so feel free to open it right back up if you still have the desire to contribute and rebase it on `main`

With an updated PR, I will 100% review it and get it merged in if it's ready.

augdiebold:
No problem @JacobCoffee , thanks for your response I'll rebase and solve it again!

PR #2270, opened 2023-04-27
Show membership type on user profile page

### LINKED ISSUE #398: Membership Class Designation and its visibility
This is related to becoming a member, but I believe it is different from others reported so I opened a new issue.
1. How does one specify what type of member they are? In the FAQ there are specific links for supporting members or sponsors, but not others. As a basic member, can I simply declare myself to be a contributing member or does someone need to approve that?  
2. There should be some type of distinction (visual clue?) between various types of members noted.  First of all, so I can see what membership class the PSF considers me to be in. And second, so I can see the designation class of other members. This second point might be especially useful as the organization grows so people have some perspective about who it is they are interacting with. Is it a brand new basic member or someone who has been heavily involved the system?

ojii:
Instead of only showing whether the user is a PSF Member, show what type of member they are on their user profile page.

This (at least partially) resolves https://github.com/python/pythondotorg/issues/398 

I would appreciate this feature because I've self-certified as a contributing member, but I don't know if that self-certification actually applied.

### DIFF
--- a/templates/users/user_detail.html
+++ b/templates/users/user_detail.html
@@ -21,7 +21,7 @@
                 <p><span class="profile-label">Name:</span> {% firstof object.get_full_name object %}</p>
             </div>
             <div class="user-profile-membership">
-                <p><span class="profile-label"><span class="icon-python" aria-hidden="true"></span> PSF Member?</span> {{ object.has_membership|yesno|capfirst }}</p>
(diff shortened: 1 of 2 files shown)


darjeeling:
@ewdurbin I think this PR is helpful to users to get easily current type of membership.
can you check once?

ewdurbin:
Hi, python.org does not currently act as or sync with the source of truth for membership. #2271 attempts to clarify until such a time as its possible to surface membership type beyond Basic on python.org.

PR #2271, opened 2023-05-12
add notes regarding basic membership

ewdurbin:
Updates the form:

form

And user detail page:

dash

With notes regarding Basic membership and how to verify other membership types.


ojii:
❤️ thank you Ee. I wasn't aware of that profile page on psfmember.


EDIT: is the psfmember website also open source? I noticed that even though I'm logged in, on the home page, there's no link to my user profile or anything, only a logout button.

Issue #2265, opened 2023-03-26
unused import in sourcecode


vignesh7056:
jobs/migrations/0013_auto_20170810_1625.py

`from django.db import models, migrations`

this is an unused import can be removed.

Saimanohar9999:
Is the above issue resolved????

JacobCoffee:
There are lots of these but they can be cleaned up by #2540 later! Thanks anyways :) Check out https://github.com/python/pythondotorg/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+label%3Ahelp-wanted for issues you can contribute to if interested

PR #2276, opened 2023-05-30
unused import in sourcecode #2265

siddhukamble45:
Remove unused import from jobs/migrations/0013_auto_20170810_1625.py


Mariatta:
Thanks for the PR. However, since this is a django migration file, we may want to leave it as is.

PR #2284, opened 2023-06-16
don't denote old versions as 'the newest major release of Python'

erlend-aasland:
See also python/cpython#105844

### DIFF
--- a/docs/source/administration.rst
+++ b/docs/source/administration.rst
@@ -65,7 +65,7 @@ Downloads and Releases
 The ``downloads`` app stores all of the structured data regarding Python releases.  Each ``Release`` object has associated ``ReleaseFile`` objects that contain information on the various download formats Python.org
 supports.
 
-If the version you are creating should be considered the "latest" release for the major version in question (Python 2.x.x, 3.x.x, etc)
+If the version you are creating should be considered the "latest" release for the minor version in question (Python 2.x.x, 3.x.x, etc)
 then check the 'Is this the latest release' checkbox.  When the ``Release`` is saved, the previous version will be automatically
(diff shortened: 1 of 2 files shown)


erlend-aasland:
I was unable to find out where the _"is the newest major release of the Python programming language"_ string for 3.11.x was located (but I did not devote much time for the investigation either ...):

https://www.python.org/downloads/release/python-3114/

For 3.11.4, it is true that it's the newest major release of Python, but that's not true for 3.11.3 and older. Perhaps "the newest stable release of Python" is more accurate.

erlend-aasland:
(I reverted 252ddbdd81fcae970dacbab7a7a9218996ceb1e3 as it broke CI, and I have no time to investigate that)

jamestwebber:
> I was unable to find out where the _"is the newest major release of the Python programming language"_ string for 3.11.x was located (but I did not devote much time for the investigation either ...):

It seems like these are only present in the `fixtures/downloads.json` file, which I think means it is stored in the database as markdown and just gets rendered in its entirety? So possibly someone (the release manager?) is writing this by hand (or has a template they use for it)

erlend-aasland:
Yep, makes sense.

ewdurbin:
The changes here are for the fixtures. The contents itself is managed in the CMS in the live database.

Happy to merge, but it will have no impact on the live pages at python.org. Someone from @python/python-release-managers will need to update the existing CMS contents.

erlend-aasland:
If it has no impact, there's no need to merge :) Thanks for the info!

jamestwebber:
The change to `docs/source/administration.rst` is still relevant, yes? It's a tiny point but perhaps would help nudge future announcements to use precise wording.

PR #2293, opened 2023-07-18
Updated init.py file

anirudh-hegde:
Screenshot (134)


ewdurbin:
Hi and thank you. Can you describe the reason for this change?

anirudh-hegde:
> Hi and thank you. Can you describe the reason for this change?

__init__.py file is the first thing that runs when you import your Python package.
This is essential file so I made this change for newbies to understand and get to know about python.
Thank You.

Issue #2020, opened 2022-04-04
Bug: meta element `og:description` property needs escaping for job board posts


joncle:
Example: https://www.python.org/jobs/6405/

The content for this is:

```
<meta property="og:description" content="<p>BENEFITS:</p>
--
  | <ul class="simple">
  | <li>Structured and well-defined book of work</li>
  | <li>Good infrastructure</li>
  | <li>Tailored work calendar</li>
  | <li>Smart and hard-working colleagues</li>
  | <li>Bonu…">
```

This is causing some of the `<ul>` element to appear above the site header. 

I'd be happy for escaping the " to &quot; for now - it's the first time I've notice this happen so any form of quick fix is fine.

Jon Clements.
Co-Chair of the PSF Job Board Working Group.

joncle:
Ooops... of course I meant `"` to `&quot;`... ironic given the issue itself is about escaping...

SDprogramer:
Is this issue still open or it is fixed by other members please tell. If not then i want to fix it. I will tell the approach clearly.

karnrautela:
is this issue resolved if no then i want to resolve it

genevievebrooks:
@JacobCoffee It looks like this issue has been fixed. I am not seeing any issues on any of the job pages that I checked. For example: [https://www.python.org/jobs/7963/](https://www.python.org/jobs/7963/)

hugovk:
Fixed by https://github.com/python/pythondotorg/pull/2316.

PR #2298, opened 2023-07-26
Update job_detail.html

joncle:
Explicity escape the job description for the meta og:description where the description contains html which can throw off the rendering of the page...

See https://www.python.org/jobs/7314/ for example.

### DIFF
--- a/templates/jobs/job_detail.html
+++ b/templates/jobs/job_detail.html
@@ -8,7 +8,7 @@
 {% block content_attributes %}with-right-sidebar{% endblock %}
 
 {% block og_title %}Job: {{ object.job_title }} at {{ object.company_name }}{% endblock %}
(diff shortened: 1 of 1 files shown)


joncle:
This should close #2020 for now

joncle:
Alternatively - it _might_ be better to escape after truncating `{{ object.description|truncatechars:200|escape }}` at the risk of going over 200 characters but at least then it wouldn't run the risk of truncating escape sequences although I think that's less of a potential issue than having the meta tag bleed html into the page... happy with either.

ewdurbin (approved):

JacobCoffee:
This seems it did not do the thing needed to fix as examples in #2307 and #2020 remain.
- https://www.python.org/jobs/6405/
- https://www.python.org/jobs/7329/

PR #2311, opened 2023-09-16
Prioritise 64-bit downloads over 32-bit

### LINKED ISSUE #2194: List 64-bit Windows Installer Above 32-bit Windows embedded

Many new comers to programming start with Python. It is not safe to assume that these newcomers will understand the difference between 32-bit and 64-bit. If Python already promotes 64-bit over 32-bit on the main download, then for users upgrading within the same minor version (3.10.X -> 3.10.Y), this promotion should remain. The Windows 64-bit installer is most likely the most downloaded binary yet it is at the bottom of the list.

https://www.python.org/downloads/release/python-3108/
https://www.python.org/downloads/windows/

I'd like the Windows downloads to be sorted as follows:

[Windows installer (64-bit)](https://www.python.org/ftp/python/3.10.8/python-3.10.8-amd64.exe)
[Windows installer (32-bit)](https://www.python.org/ftp/python/3.10.8/python-3.10.8.exe)
[Windows help file](https://www.python.org/ftp/python/3.10.8/python3108.chm)
[Windows embeddable package (64-bit)](https://www.python.org/ftp/python/3.10.8/python-3.10.8-embed-amd64.zip)
[Windows embeddable package (32-bit)](https://www.python.org/ftp/python/3.10.8/python-3.10.8-embed-win32.zip)

Status quo

**Additional context**
image
image

hugovk:
Fixes https://github.com/python/pythondotorg/issues/2194.

# Before

image

http://0.0.0.0:8000/downloads/release/python-3107/

image

http://0.0.0.0:8000/downloads/windows/

# After

image

http://0.0.0.0:8000/downloads/release/python-3107/

image

http://0.0.0.0:8000/downloads/windows/


gpshead:
Could the "embeddable package" ones also sort below the "installer" ones?

That way the first Windows thing people see in list is our Recommended one.

hugovk:
Updated to put those matching `file.name.startswith('Windows embeddable package')` last:

image

image

Older ones (see right column) have a different filename structure, so it doesn't match them:

image

We could match something like `'embeddable' in file.name`, but maybe we don't need to worry about the older ones?

gpshead:
I wonder how many people download the windows help file (whatever that is).  it could be sorted lower, probably below the installers before embedded.

regardless, this PR is good.  lets ship it and iterate further later.

hugovk:
Updated with an explicit preferred sort order for Windows files, how does this look?

image

image

gpshead:
nice!

elibroftw:
@ewdurbin Please merge this.

downloads/templatetags/download_tags.py
```diff
@@ -14,3 +14,37 @@ def has_sigstore_materials(files):
...
+def sort_windows(files):
+    if not files:
+        return files
+
+    # Put Windows files in preferred order
+    files = list(files)
+    windows_files = []
+    other_files = []
+    for preferred in (
+        'Windows installer (64-bit)',
```
ewdurbin, line 34: Are these strings generated by code elsewhere in the codebase or manually input by Release Managers?

If it is manual input, this feels brittle and likely to cause confusion.

ewdurbin: They appear to be strings stored on the [ReleaseFile model via the NameSlugModel mixin](https://github.com/python/pythondotorg/blob/c4ee749942227ca75c8e670546afe67232d647b2/downloads/models.py#L305). I'm pretty iffy on handling this this way honestly.

gpshead: @Yh1gs for RM input

Yhg1s: Yes, they're freeform strings, set by the script adding the ReleaseFile objects: https://github.com/python/release-tools/blob/master/add-to-pydotorg.py#L90.

hugovk: Before checking for the exact strings, I had something that checked for substrings like "(64-bit)", "(32-bit)" and "Windows embeddable package", see the old version at: [`dadd924` (#2311)](https://github.com/python/pythondotorg/pull/2311/commits/dadd9249a3910fd28f992580ad0b92f92495ff75)

Would you prefer something like that? Or some other approach?

merwok: IMHO having the strings is fine, if comments were added to reference this file from the other and vice-versa

ofek:
I'm willing to help with whatever needs doing here, just let me know.

ofek (approved):

PR #2344, opened 2023-12-18
Sponsorship - adding renewal option for contract generation

jessiebelle:
- allow 'renewal' box to be ticked on review of a sponsorship application
- new docx template for a renewal contract (currently the same as the existing new contract)
- logic to select this new contract when a renewal is selected
- associated tests


sponsors/tests/test_use_cases.py
```diff
@@ -118,6 +118,24 @@ def test_update_sponsorship_as_approved_and_create_contract(self):
         self.assertEqual(self.sponsorship.sponsorship_fee, 100)
         self.assertEqual(self.sponsorship.package, self.package)
         self.assertEqual(self.sponsorship.level_name, self.package.name)
+        self.assertFalse(self.sponsorship.renewal)
+
+
+    def test_update_renewal_sponsorship_as_approved_and_create_contract(self):
```
ewdurbin, line 124: This test looks great for the use-case. But I'm unable to get this to work locally without manually marking the sponsorship as a renewal in admin after checking the box in the approval flow. It looks like the "renewal" value is not making it onto the sponsorship object when the approval use case is executed.

ewdurbin: Resolved via https://github.com/python/pythondotorg/pull/2345/commits/3593c83073c981821b62aed57ac87822c8cb6473

ewdurbin (approved):

Issue #2339, opened 2023-12-07
Add Software Bill-of-Materials (SBOM) documents and user guide to python.org/downloads


sethmlarson:
As a part of https://github.com/python/cpython/issues/112302, the Software Bill-of-Materials documents should be downloadable per-artifact on python.org/downloads

- One format for now, we can add the other if someone asks. Scanners should all support both formats.
- Create a new column for each artifact
  - Naming according to OpenSSF guide on SBOM naming (ie `<artifact-name>.spdx.json` or `<artifact-name>.cdx.json`)
- User documentation on how to get SBOM documents for their corresponding Python release.
  - Third-party distributions should provide their own SBOMs, potentially using ours as a base.
- User documentation on how to use a scanner with our SBOM to detect vulnerabilities in their version of Python (ie, with our OSV vuln data and CVE vuln data)
- User documentation on how to use VEX to avoid false-positives and get up-to-date vulnerability remediation information.
- We only need one VEX document per Python release since we can reference dependencies within different SBOMs from a single VEX document (although we'll need to duplicate statements to do this? But I don't see another way right now). The VEX document can live in the PSF Advisory Database.

madpah:
@sethmlarson - I see you're working on SBOM stuff for `python.org` - that's cool. As one of the authors of the CycloneDX Python implementations, here to help if we can.

sethmlarson:
Looks like SBOM documents are being shown properly on the release page and the user documentation is published. Closing this issue!

PR #2359, opened 2024-01-12
Add support for hosting SPDX-2 SBOMs alongside release artifacts

sethmlarson:
Part 1 of https://github.com/python/pythondotorg/issues/2339

I modeled this PR off of the Sigstore bundle PR since it's a very similar approach.

There may be more than one SBOM per artifact eventually (for multiple reasons, either users demanding CycloneDX or a new major version of SPDX) so accounting for that possibility by using `spdx2` as the prefix for the database field.

~~I don't know what the "admin" interface needs to look like for release managers to be able to submit SBOM files for new releases, this might already be enough if Sigstore bundles are doable today.~~


hugovk:
Updating from `main` should fix the Read the Docs build.

sethmlarson:
@hugovk Thanks Hugo! I've rebased on `main`.

sethmlarson:
Okay I believe that by adding the `sbom_spdx2_file` field to `ReleaseFileResource` is enough to add support for uploading via API. This will require a change in https://github.com/python/release-tools/pull/82 once it's available here.

ewdurbin (approved):

PR #2362, opened 2024-01-26
Add DearPyGui to home page

### LINKED ISSUE #2360: Add DearPyGui on home page under "Use Python for…" section
[DearPyGui](https://pypi.org/project/dearpygui/) is an excellent GUI framework and I think it deserves to be listed in "Use Python for…" section of https://www.python.org/ (see below).

image

I am not sure how to do a PR for this.

cfsok:
Closes #2360


ewdurbin:
@ambv did you update the actual data in the Boxes admin? merging this doesn't actually change the content.

ambv:
I saw the change appear locally:
Screenshot 2024-02-21 at 15 34 03

If our prod DB needs an update, I'll do that after I fix main tests.

ewdurbin:
yes, locally the fixtures are loaded. in prod they were only ever used once to initialize the content.

ambv:
OK, thanks for pointing this out. Added in admin, confirmed this was the only difference in the live version of the box. It's visible on Python.org.

PR #2395, opened 2024-03-08
add celery

ewdurbin:
preparing for cabotage, which does not have Heroku Scheduler, so we'll run our own periodic tasks with Celery Beat


ewdurbin:
Current schedule
Screenshot 2024-03-08 at 7 42 00 AM

PR #2401, opened 2024-03-11
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
Generates pull requests like mymarilyn/clickhouse-driver#423 to fix warnings like at the bottom right of
https://github.com/python/pythondotorg/actions/runs/8235999706

* https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

Fixes #2424

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,4 +26,4 @@ updates:
       patterns:
         - "*"  # Group all Actions updates into a single larger pull request
   schedule:
-    interval: monthly
+    interval: daily
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,3 +19,11 @@ updates:
   - dependency-name: lxml
     versions:
     - 4.6.2
+- package-ecosystem: github-actions
+  directory: /
+  groups:
+    github-actions:
+      patterns:
+        - "*"  # Group all Actions updates into a single larger pull request
+  schedule:
+    interval: weekly
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,10 +20,10 @@ updates:
     versions:
     - 4.6.2
 - package-ecosystem: github-actions
-  directory: /
+  directory: "/"
   groups:
     github-actions:
       patterns:
         - "*"  # Group all Actions updates into a single larger pull request
   schedule:
-    interval: weekly
+    interval: monthly


.github/dependabot.yml
```diff
@@ -19,3 +19,11 @@ updates:
   - dependency-name: lxml
     versions:
     - 4.6.2
+- package-ecosystem: github-actions
+  directory: /
```
hugovk: Match the one above:
```suggestion
  directory: "/"
```

.github/dependabot.yml
```diff
@@ -19,3 +19,11 @@ updates:
...
+  groups:
+    github-actions:
+      patterns:
+        - "*"  # Group all Actions updates into a single larger pull request
+  schedule:
+    interval: weekly
```
hugovk: Monthly is less noisy:
```suggestion
    interval: monthly
```

cclauss: It has been three weeks...

GitHub Actions have very infrequent major version changes.  For instance, `setup-python` has only had four major upgrades in its lifetime.  Also, when they are upgraded, it often happens in batches.  The `pattern: *` will consolidate all GHA updates into a single pull request to further reduce chattiness.

There is a tradeoff between supply chain security and chattiness.  Given that we have a few GHAs that are updated rarely and usually in batches and we are using `pattern: *` to ensure that there will only ever be a single GHA upgrade PR at a time, my recommendation would have been to kick off this process in weekly mode.  It is trivial to switch to monthly mode if we sense it becomes chatty.  I doubt that it will.

cclauss: Perhaps the `pattern: *` approach should be used on the `pip` dependencies which are far more chatty than the GHA updates could ever be.  https://github.com/python/pythondotorg/pulls/app%2Fdependabot

ezio-melotti (approved):

hugovk (approved):

.github/dependabot.yml
```diff
@@ -19,3 +19,11 @@ updates:
...
     versions:
     - 4.6.2
+- package-ecosystem: github-actions
+  directory: "/"
+  groups:
+    github-actions:
+      patterns:
+        - "*"  # Group all Actions updates into a single larger pull request
+  schedule:
+    interval: monthly
```
ewdurbin:
```suggestion
    interval: daily
```

no worries on noisiness since dependabot handles superseding

PR #2413, opened 2024-03-19
feat: migration from 2.2 -> 3.2 -> 4.2

### LINKED ISSUE #1860: Upgrade to Django 3.2 LTS version
The PR #1830 upgrades the project up to 2.2 LTS version. But, as described in #1819, the support time lasts until next April. We're still under a maintained version until there, but it's something to have in mind. 

An upgrade to Django 3.2 (3.* LTS version) which shall be maintained until April's 2024.

**Additional context**
The PR by @luzfcb #1830 has a lot of documentation on things that have already been addressed and further discussions are in #1819

JacobCoffee:
## What

- Attempts to migration from the current 2.2 dependency set to 3.2. Then migrates from 4.2 after fixing any 3.2 issues.
- After 4.2 migration, fixes any remaining stale dependencies by finding the first available, noting it in its respective `*-requirements.txt`, and bumping to the latest available non-breaking version to give "breathing room"
- At this point, able to get a live webserver, we can apply migrations and get the loading webpage

## Status

- Tests passing
- This does require a production database migration from PostgreSQL 10 -> 12. I'm not sure what that will entail for the infra team

## Todo

- ~~Does the scope of this include cleaning up the BigAutoField warnings?~~
```shell
Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.

HINT: Configure the DEFAULT_AUTO_FIELD setting or the $SOMEMODEL.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
```
  - these dont happen in current version so im going to say yes
  - resolved via [e1ecf70](https://github.com/python/pythondotorg/pull/2413/commits/e1ecf709d2e9f96b51a214f46b5a77e020ac2dfa)
  - Verify this is an okay or not change and discuss path forward if not
---

- Does the scope of this include cleaning up the naive datetime warning?
```shell
RuntimeWarning: DateTimeField $THING received a naive datetime (2013-08-24 00:00:00) while time zone support is active.
```
  - these happen in current version so im going to say no (?)
  - Opened issue https://github.com/python/pythondotorg/issues/2415
  
## Closes

- Fixes #1860


pydotorg/settings/base.py
```diff
@@ -163,6 +163,7 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'pages.middleware.PageFallbackMiddleware',
     'django.contrib.redirects.middleware.RedirectFallbackMiddleware',
+    'allauth.account.middleware.AccountMiddleware',
```
JacobCoffee, line 172: right place?

docker-compose.yml
```diff
@@ -1,8 +1,21 @@
...
+#       POSTGRES_PASSWORD: pythondotorg
+#       POSTGRES_DB: pythondotorg
+#       POSTGRES_HOST_AUTH_METHOD: trust  # never do this in production!
+#     healthcheck:
+#       test: ["CMD", "pg_isready", "-U", "pythondotorg", "-d", "pythondotorg"]
+#       interval: 1s
+
   postgres:
-    image: postgres:10-bullseye
+    image: postgres:12-bullseye  # required for 4.2 upgrades
```
JacobCoffee: bullseye vs bookworm? would give us more runway on EOL (bullseye has until 2026, though)

base-requirements.txt
```diff
@@ -22,32 +22,32 @@ chardet==4.0.0
...
 
-django-honeypot==1.0.1
-django-markupfield==2.0.0
-django-markupfield-helpers==0.1.1
+django-honeypot==1.0.4  # 1.0.4 is first version that supports Django 4.2
+django-markupfield==2.0.1
+# django-markupfield-helpers==0.1.1  # this hasnt been moved out of alpha and is last published 2018, we do import it
 
-django-allauth==0.41.0
+django-allauth==0.61.1  # 0.55.0 is first version that supports Django 4.2
```
JacobCoffee, line 39: bumped higher than least because of CVE (i think)

base-requirements.txt
```diff
@@ -22,32 +22,32 @@ chardet==4.0.0
...
 python-dateutil==2.8.2
 
 requests[security]>=2.26.0
 
-django-honeypot==1.0.1
-django-markupfield==2.0.0
-django-markupfield-helpers==0.1.1
+django-honeypot==1.0.4  # 1.0.4 is first version that supports Django 4.2
+django-markupfield==2.0.1
+# django-markupfield-helpers==0.1.1  # this hasnt been moved out of alpha and is last published 2018, we do import it
```
JacobCoffee: i cant find where we use this, but it seems we dont. anyway, it doesnt support either 3.2 or 4.2 iirc

base-requirements.txt
```diff
@@ -1,10 +1,10 @@
 dj-database-url==0.5.0
-django-pipeline==2.0.6
-django-sitetree==1.17.0
+django-pipeline==3.0.0  # 3.0.0 is first version that supports Django 4.2
+django-sitetree==1.18.0  # >=1.17.1 is (?) first version that supports Django 4.2
```
JacobCoffee, line 3: unclear changelog on when exactly 4.2 support was introduced

JacobCoffee:
Somewhere along the way the email duplication validation from allauth became wonky causing https://github.com/python/pythondotorg/blob/9b4deddbd43cf6e5518a72ddcbc4c94c7b6d487f/users/tests/test_views.py#L233 to fail.
It seems to be my last failing test, barring i didnt do anything 'bad' by 'fixing' the failing tests post-migration 😅 

---

edit: this was a change in allauth 0.52.0, but disabling it would be a negative since i believe it was added for increased security.

I can add
```py
ACCOUNT_PREVENT_ENUMERATION = False
"""Added to ``django-allauth==0.52.0``. Set during the Django 2.2 -> 4.2 migration."""
```
into settings.py
or (preferably) add it *only* to test which seems safer.
```py
def test_user_duplicate_username_email(self):
    """Test that a user can't be created with a duplicate username or email or both."""
    settings.ACCOUNT_PREVENT_ENUMERATION = False
```
it prevents people from bruteforcing to see which emails are registered
(read more at https://docs.allauth.org/en/latest/account/configuration.html -> search for `ACCOUNT_PREVENT_ENUMERATION`)

JacobCoffee:
I think this is ready for someone to make fun of me now :rofl:, all tests are passing

JacobCoffee:
[38f1eff](https://github.com/python/pythondotorg/pull/2413/commits/38f1eff6b8c0e09483ae657f17de5831c2ecb99a) may require `override_settings` from the test suite? im not familiar enough with this to know the potential spillage...

users/tests/test_views.py
```diff
@@ -231,6 +229,7 @@ def test_user_new_account(self):
         })
 
     def test_user_duplicate_username_email(self):
+        settings.ACCOUNT_PREVENT_ENUMERATION = False
```
JacobCoffee: this may require `override_settings` from the test suite? im not familiar enough with this to know the potential spillage...

pydotorg/settings/base.py
```diff
@@ -31,6 +31,12 @@
     )
 }
 
+DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
+"""The default primary key field type for Django models.
+
+Required during the Django 2.2 -> 4.2 migration.
+"""
```
browniebroke, line 38: I'm not sure about this one, I noticed it generated a bunch of migrations. Should we set it to `django.db.models.AutoField` while we upgrade and change it after? 

Reason I'm asking is that I noticed that [the docs](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-DEFAULT_AUTO_FIELD) mention that not all cases are covered by the migration framework:

> Unfortunately, the primary keys of existing auto-created through tables cannot currently be updated by the migrations framework.
> 
> This means that if you switch the value of DEFAULT_AUTO_FIELD and then generate migrations, the primary keys of the related models will be updated, as will the foreign keys from the through table, but the primary key of the auto-created through table will not be migrated.
> 
> In order to address this, you should add a [RunSQL](https://docs.djangoproject.com/en/5.0/ref/migration-operations/#django.db.migrations.operations.RunSQL) operation to your migrations to perform the required ALTER TABLE step. You can check the existing table name through sqlmigrate, dbshell, or with the field’s remote_field.through._meta.db_table property.

I don't know the codebase too much, so might not be an issue, but it feels like this change in primary key doesn't have to be part of the upgrade (which is quite big already).

docker-compose.yml
```diff
@@ -1,8 +1,21 @@
...
+#       - "5433:5432"
+#     environment:
+#       POSTGRES_USER: pythondotorg
+#       POSTGRES_PASSWORD: pythondotorg
+#       POSTGRES_DB: pythondotorg
+#       POSTGRES_HOST_AUTH_METHOD: trust  # never do this in production!
+#     healthcheck:
+#       test: ["CMD", "pg_isready", "-U", "pythondotorg", "-d", "pythondotorg"]
+#       interval: 1s
+
```
browniebroke: I guess we should remove this before merging, right?

```suggestion
```

events/tests/test_importer.py
```diff
@@ -2,6 +2,7 @@
 
 from django.test import TestCase
 from django.utils.timezone import datetime, make_aware
+from django.utils.html import strip_tags
```
browniebroke: Is that needed?

prod-requirements.txt
```diff
@@ -3,6 +3,6 @@ gunicorn==19.9.0
 raven==6.10.0
```
browniebroke, line 3: Oh that's an old one 😄 Not sure it supports Django 4.2, but maybe it's fine... Would be nice to see if we can switch to its successor, sentry-sdk.

users/tests/test_forms.py
```diff
@@ -1,5 +1,5 @@
 from django.contrib.auth import get_user_model
-from django.test import TestCase
+from django.test import TestCase, RequestFactory
```
browniebroke, line 2: Are you using this new import?

users/tests/test_forms.py
```diff
@@ -50,16 +50,22 @@ def test_duplicate_username(self):
         self.assertIn('username', form.errors)
 
     def test_duplicate_email(self):
+        from django.http import HttpRequest
```
browniebroke: Did you mean to use the `RequestFactory` above?

users/tests/test_views.py
browniebroke: Better to use `override_settings` as it will undo the setting change at the end of the test, resetting it to its original value. Setting it directly like this might leak to the other tests (I think)

browniebroke (commented):
Solid work 💪🏻 I noted a few things. FWIW, I only went though everything you changed (I haven't looked at what you may have missed). Also note that I'm not too familiar with the codebase, but I went through a few Django upgrades, I hope these coments are of use...

docker-compose.yml
JacobCoffee: ahh yes thank you, I wasn't sure on the production deployment side what all it might entail for them to migrate.

users/tests/test_forms.py
JacobCoffee: yep, thank you 😬

codesamples/tests.py
```diff
@@ -16,9 +16,13 @@ def setUp(self):
             is_published=False)
 
     def test_published(self):
-        self.assertQuerysetEqual(CodeSample.objects.published(),
-                                 ['<CodeSample: Copy One>'])
+        expected = ['<CodeSample: Copy One>']
+        published_qs = CodeSample.objects.published()
+        actual = [f'<CodeSample: {str(obj)}>' for obj in published_qs]
+        self.assertEqual(actual, expected)
```
browniebroke, line 22: Perhaps this would simpler using `assertCountEqual()`, like this:

```suggestion
        self.assertCountEqual(CodeSample.objects.published(),[self.sample2])
```

I find the name of the method slightly misleading, but it would work nicely here, I think:

> Test that sequence first contains the same elements as second, regardless of their order

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

The only difference I see is that we would no longer cover the `__str__` of the model. That could be separate test, if it's deemed important.

ephes:
Cool, this really looks good! I didn't see your PR yesterday and started working a bit on the Django [update by myself](https://github.com/ephes/pythondotorg/commits/django-update/). I also didn't notice that `makemigrations` would create a lot of migrations, so I got the tests running again without adding those migrations. I honestly don't know whether they are necessary. It's probably a good idea not to fall too much behind, so I think they should be added.

Another thing I did different was trying to update all packages in [base and dev-requirements.txt](https://github.com/ephes/pythondotorg/commit/df1a19f3be5e2459ae7dcea3eeaa6b6c45f1ea32). And except `tablib` and `urllib3` it seems to be possible to upgrade all dependencies.

pydotorg/settings/base.py
ephes: Seems fine to me.

docker-compose.yml
ephes: I would prefer bookworm. And postgres:12 is also kind of old - I have postgres:16 on my local machine and it worked without problems.

ephes:
The test fails because allauth wont tell you whether an email already exists or not to mitigate against email enumeration. In 0.52.0 there was this setting added, but since 0.55.0 it has no longer an effect because it didn't really work from the start.

But even if no error message is shown to the user, an error message is sent via email to the original user of the email address. What I did was splitting up the [test into two](https://github.com/ephes/pythondotorg/commit/12cdf5c846d8c5ac763408f8fc89351a69fa71af) to make sure the error message is sent.

ephes:
> Somewhere along the way the email duplication validation from allauth became wonky causing
> 
> https://github.com/python/pythondotorg/blob/9b4deddbd43cf6e5518a72ddcbc4c94c7b6d487f/users/tests/test_views.py#L233
> 
> to fail.
> It seems to be my last failing test, barring i didnt do anything 'bad' by 'fixing' the failing tests post-migration 😅
> edit: this was a change in allauth 0.52.0, but disabling it would be a negative since i believe it was added for increased security.
> 
> I can add
> 
> ```python
> ACCOUNT_PREVENT_ENUMERATION = False
> """Added to ``django-allauth==0.52.0``. Set during the Django 2.2 -> 4.2 migration."""
> ```
> 
> into settings.py or (preferably) add it _only_ to test which seems safer.
> 
> ```python
> def test_user_duplicate_username_email(self):
>     """Test that a user can't be created with a duplicate username or email or both."""
>     settings.ACCOUNT_PREVENT_ENUMERATION = False
> ```
> 
> it prevents people from bruteforcing to see which emails are registered (read more at https://docs.allauth.org/en/latest/account/configuration.html -> search for `ACCOUNT_PREVENT_ENUMERATION`)

The test fails because allauth wont tell you whether an email already exists or not to mitigate against email enumeration. In 0.52.0 there was this setting added, but since 0.55.0 it has no longer an effect because it didn't really work from the start.

But even if no error message is shown to the user, an error message is sent via email to the original user of the email address. What I did was splitting up the [test into two](https://github.com/ephes/pythondotorg/commit/12cdf5c846d8c5ac763408f8fc89351a69fa71af) to make sure the error message is sent.

codesamples/tests.py
ephes: It's possible to rewrite [those tests](https://github.com/ephes/pythondotorg/commit/3d3458cd25c8287f31e5c77b9366f1321b5061ce) like this:
```python
    def test_published(self):
        self.assertQuerySetEqual(CodeSample.objects.published(),
                                ['<CodeSample: Copy One>'], transform=repr)
```

cclauss:
Solid work @JacobCoffee !!

A few things you might consider...
1. Land a smaller PR on this repo so that `1 workflow awaiting approval` disappears and GitHub Actions are run on commits.
2. Perhaps upgrade to 3.2 and let the site run for a few weeks before making a separate upgrade to 4.2.
3. Add `Django` to the title: `feat: migration from Django 2.2 -> 3.2 -> 4.2`.

@ambv Your review, please.

ewdurbin:
Current PG version running python.org's db: 15.3

JacobCoffee:
@ewdurbin is this looking okay so far? If so I can try to finish it up - but I wanted to make sure we wanted to move forward here before continuing anymore work.

ewdurbin:
@JacobCoffee Yes, overall this is looking very promising. I was able to get things stood up in a test environment with a clone of the production DB without any obvious errors.

pydotorg/settings/base.py
ewdurbin: I agree, reverting this and setting `DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'` seems preferable (and doesn't lead to the missing migration in tastypie that is causing failing CI)

base-requirements.txt
ezio-melotti: This was used in `sponsors/pdf.py` and `sponsors/tests/test_pdf.py` which no longer exist.  I created a PR to remove the dependency:
* https://github.com/python/pythondotorg/pull/2474

You can either wait for my PR to be approved/merged and then merge `main` into this branch, or just remove it directly from this branch:
```suggestion
```

JacobCoffee:
In favor of #2520

PR #2491, opened 2024-07-25
fix(frontend): Improve low constrast of code blocks and links

### LINKED ISSUE #2446: A11y: Contrast of code formatting too low
**Describe the bug**

**To Reproduce**
1. Go to https://www.python.org/downloads/release/python-3130a6/
2. View the list of removed "dead batteries" with code formatting

**Actual behavior**

The contrast of the green text on a grey background is too low, making it hard to read and affecting accessibility.

It's something like 2.88:1.

https://webaim.org/resources/contrastchecker/?fcolor=11A611&bcolor=F1F2F3

**Expected behavior**

The ratio should be at least 4.5:1.

> WCAG 2.0 level AA requires a contrast ratio of at least 4.5:1 for normal text and 3:1 for large text. ... WCAG Level AAA requires a contrast ratio of at least 7:1 for normal text and 4.5:1 for large text.

AA is a good target to aim for.

**Screenshots**

image

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version 124

**Additional context**

image

The contrast of the blue links in the table with different grey backgrounds is also a little low, would be good to bump these to 4.5:1+ as well, but the green above is more important.

* 4.24:1 https://webaim.org/resources/contrastchecker/?fcolor=3776AB&bcolor=F0F0F0

* 4.47:1 https://webaim.org/resources/contrastchecker/?fcolor=3776AB&bcolor=F6F6F6

tomasr8:
Closes #2446 

I changed the code block color for the closest darker green that surpasses `4.5` - Firefox is now reporting it as `4.57`.

I didn't want to change the link color for all links, so I just added a new rule to target links in tables only. Firefox now reports the new value as `4.56`.

Let me know if it's ok like this!

### DIFF
--- a/static/sass/_base.scss
+++ b/static/sass/_base.scss
@@ -85,7 +85,7 @@ $darkerblue: #1e2933;
 $link-color: $blue;
 $link-hover: $darkerblue;
 
-$code-green: #11a611; 
+$code-green: #0c7d0c;
 $default-border-color: $grey-lighter; /* Set a generic border color here to keep them consistent */
 
 
--- a/static/sass/_base.scss
+++ b/static/sass/_base.scss
@@ -85,7 +85,7 @@ $darkerblue: #1e2933;
 $link-color: $blue;
 $link-hover: $darkerblue;
 
-$code-green: #0c7d0c;
+$code-green: #0d870d;
 $default-border-color: $grey-lighter; /* Set a generic border color here to keep them consistent */
 
 
--- a/static/sass/mq.css
+++ b/static/sass/mq.css
@@ -1651,24 +1651,24 @@ html[xmlns] .slides { display: block; }
     padding-top: 1em; }
 
   .about-banner {
-    background: 120% 0 no-repeat url('../img/landing-about.png?1646853871') transparent;
+    background: 120% 0 no-repeat url('../img/landing-about.png?1767349471') transparent;
     min-height: 345px;
     padding-bottom: 3.5em;
     margin-bottom: -2.5em; }
 
   .download-for-current-os {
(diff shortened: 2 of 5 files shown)


hugovk:
I only checked the green so far, looks better.

Before:

image


After:

image

You know, I might be tempted to drop that rounded grey background box altogether, we don't have that sort of thing at https://docs.python.org/3/whatsnew/3.12.html, or keep the grey background but drop the shadow and rounder border radius.

image

image

(Possibly also ditching the green?)

image

image

tomasr8:
Could also decrease the padding.. It looks pretty big

JacobCoffee:
I think I like this the best out of the options:

image

but the current "After:" state is good, too.

JacobCoffee (approved):
thanks for your first contribution, @tomasr8 :)

hugovk:
Looks like we forgot about this PR :)

> I think I like this the best out of the options:
> 
> image

Yeah, let's ditch all padding/margin/background/box stuff.

@tomasr8 Please could you update to remove that too?

tomasr8:
> Please could you update to remove that too?

I missed the ping, sorry, will update it hopefully tomorrow :)

tomasr8:
So I updated the PR:
- Removed background, padding and margins for the `code` element
- Adjusted the green shade since it's against a white background

I also committed the generated files. I figured since they're not gitignored, they should be added as well, but let me know if that's not the case!

Here's a different example for reference:
image

JacobCoffee:
thanks @tomasr8 & @hugovk !

PR #1714, opened 2020-12-27
Issue #1701: Events page displays year for events scheduled to start or end at a future year

alvindera97:
This is the PR replacing https://github.com/python/pythondotorg/pull/1713. The functional tests have been removed and logical/functional concerns have been replicated with unit tests that don't require selenium to run. 

_This is basically because of the CI Pipeline which makes no provision for installing selenium's web driver required by selenium to run the [headless] functional tests._ 

**There is something that I would like to note:**
— When I was running the unit tests, I found out that Django was somehow caching the results of the tests and so the subsequent test results I was getting were rather untrustworthy. 

• I was able to get around this by adding a `tearDown()` method to the relevant unit test class. This `tearDown()` method basically clears the cache at the end of each test (`cache.clear()`) suggested [here](https://www.reddit.com/r/django/comments/6fx18f/psa_django_core_caching_pollutes_test_results/?utm_source=share&utm_medium=web2x&context=3)

Please feel free to ask any questions. This is basically my first shot at contributing to open source software.

PR #2500, opened 2024-08-08
fix(#1701): Events page displays year for events scheduled to start or end at a future year

### LINKED ISSUE #1701: Calendar events should show the year for events which are not in the current year

alvindera97:
## What
- Addition of event year text to events page (for events scheduled with any date **not** in the current year.
- Replaces https://github.com/python/pythondotorg/pull/1714.

## Closes
- Fixes https://github.com/python/pythondotorg/issues/1701.


alvindera97:
Good day everyone, I hope you're all doing well.

I initially worked on this issue (#1701) some time ago, and it was opened in 2020. The problem involved events with start or end dates that are not in the current year (i.e. not matching `datetime.datetime.now().year` on the server). These events were displayed without indicating the year, which could cause confusion for users.

While event names often include the year, displaying the year directly with the event dates on the frontend provides additional clarity and improves user experience by avoiding any ambiguity.

If possible, I'd appreciate a review of my pull request. I’m happy to answer any questions you may have and will do my best to clarify any details.

Thank you for your time, and I look forward to your feedback.

Best regards,  
Godwin

alvindera97:
After reviewing some other pull requests on the repo, I decided that adding screenshots of before and after would better give a visual summary of what has been done

## Before: 
Screenshot from 2024-09-15 17-01-50



## After: 
Screenshot from 2024-09-15 16-57-00

events/tests/test_models.py
```diff
@@ -186,3 +185,25 @@ def test_event_previous_event(self):
         # 'Event.previous_event' can return None if there is no
         # OccurringRule or RecurringRule found.
         self.assertIsNone(self.event.previous_event)
+
+    def test_scheduled_to_start_this_year_method(self):
+        now = seconds_resolution(timezone.now())
+
+        occurring_time_dtstart = now + datetime.timedelta(days=3)
```
JacobCoffee: Wouldnt this fail if it were the last 2 days of the year?

events/tests/test_models.py
```diff
@@ -186,3 +185,25 @@ def test_event_previous_event(self):
...
+            event=self.event,
+            dt_start=occurring_time_dtstart,
+            dt_end=occurring_time_dtstart
+        )
+        self.assertTrue(self.event.is_scheduled_to_start_this_year())
+
+    def test_scheduled_to_end_this_year_method(self):
+        now = seconds_resolution(timezone.now())
+
+        occurring_time_dtstart = now + datetime.timedelta(days=3)
```
JacobCoffee: same

events/tests/test_models.py
alvindera97: Hmmmm... Thinking about this test again. On the off chance that tests are run between the 29th of December and the 31st of December, this test will fail since the next year will be set to 3 days ahead...

So if tests are run on the 29th of December, `occurring_time_dtstart` will be set to 3 days from then which would be the 1st of January (the following year).

Since the date comparison is made against `datetime.datetime.now().year` at `Event.is_scheduled_to_start_this_year` and `Event.is_scheduled_to_end_this_year`, (https://github.com/alvindera97/pythondotorg/blob/fa6e0ced4827a1652aa2eef0a80db22b1a2c4fe2/events/models.py#L184-L189) having a timedelta of 3 days makes running tests within the last 3 days of the year flaky.

events/tests/test_models.py
alvindera97: Same comment here concerning 
```
occurring_time_dtstart = now + datetime.timedelta(days = 3)
``` 
with previous comment at https://github.com/python/pythondotorg/pull/2500/files#r1764230987

**If it's fine with everyone, I'll await feedback first before making any more changes to the pull request so I'd attempt resolving all concerns with as minimal back-and-forth as possible.**

Thank you for your time and best regards everyone.

events/tests/test_models.py
hugovk: You can use something like https://github.com/spulec/freezegun to freeze time so the tests run reliably.

alvindera97: > You can use something like https://github.com/spulec/freezegun to freeze time so the tests run reliably.

Thanks, @hugovk, for suggesting [freezegun](https://github.com/spulec/freezegun) to handle the time-dependent tests. I have tested the updates locally, and I’m ready to push the changes to the current pull request. However, before proceeding, I wanted to get clarity on one remaining concern regarding the introduction of external dependencies into the project.

Specifically, I’m unsure of the current team policy regarding the addition of new dependencies. My concern revolves around potential risks that come with relying on third-party libraries, such as maintenance issues, security vulnerabilities, and compatibility with future versions of Python/Django.

In this case, while freezegun is a robust library for freezing time, I wonder if it’s necessary to introduce an entire dependency solely to address the time-delta requirements of the Event object in the tests. I'm considering whether there might be an alternative approach that avoids the need for an external library, such as directly mocking time or using built-in Python features to clamp dates.

I would appreciate your thoughts on whether the use of freezegun is acceptable for this situation, or if the team would prefer exploring another solution.

However, as an example, I thought of reducing the time-deltas. Here's a snippet **without freezegun**: 
```
    def test_scheduled_to_start_this_year_method(self):
        now = seconds_resolution(timezone.now())

        occurring_time_dtstart = now + datetime.timedelta(seconds=10)
        OccurringRule.objects.create(
            event=self.event,
            dt_start=occurring_time_dtstart,
            dt_end=occurring_time_dtstart + datetime.timedelta(days=3)
        )
        self.assertTrue(self.event.is_scheduled_to_start_this_year())

        OccurringRule.objects.get(event=self.event).delete()

        event_not_scheduled_to_start_this_year_occurring_time_dtstart = now + datetime.timedelta(days=365)
        OccurringRule.objects.create(
            event=self.event,
            dt_start=event_not_scheduled_to_start_this_year_occurring_time_dtstart,
            dt_end=event_not_scheduled_to_start_this_year_occurring_time_dtstart + datetime.timedelta(days=3)
        )

        self.assertFalse(self.event.is_scheduled_to_start_this_year())

    def test_scheduled_to_end_this_year_method(self):
        now = seconds_resolution(timezone.now())
        occurring_time_dtstart = now + datetime.timedelta(seconds=10)

        OccurringRule.objects.create(
            event=self.event,
            dt_start=occurring_time_dtstart,
            dt_end=occurring_time_dtstart
        )

        self.assertTrue(self.event.is_scheduled_to_end_this_year())

        OccurringRule.objects.get(event=self.event).delete()

        event_not_scheduled_to_end_this_year_occurring_time_dt_end = now + datetime.timedelta(days=365)

        OccurringRule.objects.create(
            event=self.event,
            dt_start=now,
            dt_end=event_not_scheduled_to_end_this_year_occurring_time_dt_end
        )

        self.assertFalse(self.event.is_scheduled_to_end_this_year())

```

Looking forward to your feedback 🙂

hugovk: Thanks for your question, it's always good to keep these things in mind. My feeling is that it would be okay, as there are already a number of third-party dependencies, but I'll leave it to the maintainers to answer more concretely.

An alternative would be to use something like `unitest.mock` to fake the date returned by `datetime.timedelta`.

JacobCoffee: Freezgun is fine, it should go into here https://github.com/python/pythondotorg/blob/7928f4fb6eb77f94623534a8caf854d0dbe61b67/dev-requirements.txt#L3 @alvindera97

alvindera97: Good day @hugovk and all,

I've opted to mock timezone (https://github.com/alvindera97/pythondotorg/blame/c4db8d42591f359eb899a5940ea25b71aebcf8a3/events/tests/test_models.py#L191-L247). While Django's import path for timezone seems stable, any future changes would require updating the mock path. Currently, I'm not aware of any deprecations or changes affecting timezone.

I'm open to further feedback, but in the absence of any, I'd like to thank you all again for your valuable input and prompt responses.

JacobCoffee (approved):
image
lgtm

JacobCoffee:
Thanks for your continued contributions, @alvindera97! Very much appreciated :)

PR #2515, opened 2024-08-26
feat: testing the waf

JacobCoffee:
## What

- Pulls our python.org infra into IaC
- Applys Fastlys NGWAF
- Manages DNS entires on route53

Creates test service to test all of this under test.python.org


JacobCoffee:
```
fastly_service_vcl.test_python_org: Destroying... [id=KYF1wSNuqOtKdujNmlAXu3]
fastly_service_vcl.test_python_org: Destruction complete after 0s
fastly_service_vcl.test_python_org: Creating...
╷
│ Error: 400 - Bad Request:
│ 
│     Title:  Bad request
│     Detail: either an IAM role or an access key and secret key pair must be provided
│     RateLimitRemaining: 901
│     RateLimitReset:     1724713200
│ 
│   with fastly_service_vcl.test_python_org,
│   on fastly.tf line 1, in resource "fastly_service_vcl" "test_python_org":
│    1: resource "fastly_service_vcl" "test_python_org" {
│ 
╵
Operation failed: failed running terraform apply (exit 1)

```
Need to set up key specifically for s3 tomorrow

infra/Makefile 
```diff
@@ -0,0 +1,15 @@
...
+fmt:
+    @tf fmt ../**/*.tf
+    @tf fmt ../**/*.tfvars
+    @tf fmt ../**/*.tftest.hcl
+
+.PHONY: check
+check:
+    @tf validate
+
+.PHONY: yolo
```
ewdurbin: I think it would be best to drop this, since we intend to use terraform cloud.

infra/ngwaf/variables.tf
```diff
@@ -0,0 +1,42 @@
...
+variable "NGWAF_SITE" {
+  type        = string
+  description = "Site SHORT name for NGWAF"
+  default     = "test"
+}
+
+variable "NGWAF_EMAIL" {
+  type        = string
+  description = "Email address associated with the token for the NGWAF API."
+  default     = "jacob.coffee@pyfound.org"
```
JacobCoffee: bad

infra/ngwaf/variables.tf
```diff
@@ -0,0 +1,42 @@
...
+variable "NGWAF_CORP" {
+  type        = string
+  description = "Corp name for NGWAF"
+  default     = "python"
+}
+
+variable "NGWAF_SITE" {
+  type        = string
+  description = "Site SHORT name for NGWAF"
+  default     = "test"
```
JacobCoffee, line 29: bad

infra/ngwaf/README.md
JacobCoffee, line 1: will yank this module into its own PR

infra/dns/README.md
JacobCoffee, line 1: really need to do a `tf import` against route53 config for py.org because there is a lot there and dont want to break things

ewdurbin: I think we should remove DNS handling as a concern for this right now. Terraforming DNS for python.org will be a much bigger project. I think the terraform states should only concern themselves with Fastly/NGWaf

ewdurbin (commented):
I wasn't able to do any local testing since vars are missing in terraform cloud, but I did have a few notes!

PR #2519, opened 2024-09-04
feat: terraform python.org fastly configuration

JacobCoffee:
### What

- Ingests currently Fastly configuration for `www.Python.org` to provide Terraform configuration for test.python.org and python.org.


JacobCoffee:
I think I would like to include both environments DNS entries in this as well (not the entirety of Python.org DNS) I know that we manage a bunch outside of `www.python.org` specifically, but the core things needed for `{test,www}.python.org` to work I think should be here, and the rest could be elsewhere. 

If that is the case, I think pulling in the DNS work from the other branch would get us mostly there

ewdurbin:
I'm largely -1 on including management of DNS at this time (and likely ever) for this service as part of this repository.

The project to import python.org DNS into configuration management with terraform will be pretty large, and likely live in its own independent repository. Thus we'd end up with multiple Terraform states managing it, if we also did it here.

For now let's hold off on that until we get a better idea what the larger project looks like.

ewdurbin (approved):

infra/cdn/README.md
```diff
@@ -0,0 +1,32 @@
...
+This module creates Fastly services for the Python.org staging and production instances.
+
+## Usage
+
+```hcl
+module "fastly_production" {
+  source = "./cdn"
+
+  name            = "CoolPythonApp.org"
+  domain          = "CoolPythonApp.org"
```
JacobCoffee, line 12:
```suggestion
  domain          = "CoolPythonApp.org"
  subdomain       = "www.CoolPythonApp.org"
```

PR #2529, opened 2024-09-11
give admin user a preview of sponsorship package revenue splits

indepndnt:
This adds information to the admin screen for Sponsorship Packages about the revenue split for unmodified sponsorships.

Screenshot from 2024-09-11 08-17-52

### DIFF
--- a/sponsors/admin.py
+++ b/sponsors/admin.py
@@ -198,19 +198,13 @@ class SponsorshipPackageAdmin(OrderedModelAdmin):
         return {}
 
     def get_benefit_split(self, obj: SponsorshipPackage) -> str:
-        # colors selected from defined css variables in order to look like they fit in the theme
         colors = [
-            "secondary",  # blue
-            "admin-interface-header-background-color",  # dark green
-            "admin-interface-title-color",  # yellow
-            "admin-interface-header-text-color",  # light green
-            "admin-interface-delete-button-background-hover-color",  # red
-            "admin-interface-save-button-background-hover-color",  # darker green
-            "admin-interface-module-background-selected-color",  # light yellow
-            "admin-interface-generic-link-hover-color",  # medium green
+            "#ffde57",  # Python Gold
+            "#4584b6",  # Python Blue
+            "#646464",  # Python Grey
         ]
         split = obj.get_default_revenue_split()
-        # I hope there's never more than 8 things in the split, but just in case...
+        # rotate colors through our available palette
         if len(split) > len(colors):
             colors = colors * (1 + (len(split) // len(colors)))
         # build some span elements to show the percentages and have the program name in the title (to show on hover)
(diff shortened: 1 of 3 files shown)


sponsors/models/sponsorship.py
```diff
@@ -117,6 +117,18 @@ def clone(self, year: int):
             slug=self.slug, year=year, defaults=defaults
         )
 
+    def get_default_revenue_split(self) -> list[tuple[str, float]]:
+        """
+        Give the admin an indication of how revenue for sponsorships in this package will be divvied up
+        """
+        values, key = {}, "program__description"
```
ewdurbin:
```suggestion
        values, key = {}, "program__name"
```

This is a required field, so I recommend using that.. Description is not provided for most of the packages!

sponsors/admin.py
```diff
@@ -196,6 +197,36 @@ def get_prepopulated_fields(self, request, obj=None):
             return {'slug': ['name']}
         return {}
 
+    def get_benefit_split(self, obj: SponsorshipPackage) -> str:
```
ewdurbin, line 200: I'd reccommend hard-coding here since the admin pallete is not going to consistently have all values populated.

With the prod  theme for instance this looks like:

Screenshot 2024-09-11 at 4 22 30 PM


Something like

```python
    def get_benefit_split(self, obj: SponsorshipPackage) -> str:
        colors = [
            "#ffde57",  # Python Gold
            "#4584b6",  # Python Blue
            "#646464",  # Python Grey
        ]
        split = obj.get_default_revenue_split()
        # rotate colors through our available palette
        if len(split) > len(colors):
            colors = colors * (1 + (len(split) // len(colors)))
        # build some span elements to show the percentages and have the program name in the title (to show on hover)
        widths, spans = [], []
        for i, (name, pct) in enumerate(split):
            pct_str = f"{pct:.0f}%"
            widths.append(pct_str)
            spans.append(f"<span title='{name}' style='background-color:{colors[i]}'>{pct_str}</span>")
        # define a style that will show our span elements like a single horizontal stacked bar chart
        style = f'color:#fff;text-align:center;cursor:pointer;display:grid;grid-template-columns:{" ".join(widths)}'
        # wrap it all up and put a bow on it
        html = f"<div style='{style}'>{''.join(spans)}</div>"
        return mark_safe(html)
```

is probably fine and gives a lil pythonic flair!

Screenshot 2024-09-11 at 4 21 14 PM

ewdurbin (commented):
A couple of suggestions after running this locally.

I'm +1 to this if we add a bit stronger of a test, perhaps having multiple packages with multiple different values.

ewdurbin (approved):

indepndnt:
Thanks Ee! I had initially thought about using the Python colors but then I went a different way. I beefed up the test and applied your suggestions.

ewdurbin:
Thanks @indepndnt, this looks great!

PR #2540, opened 2024-09-12
infra: format with ruff

### LINKED ISSUE #2539: infra(dev): add code formatting/linting
It would be nice to not worry about reviewing things and editing line lengths.

I would like to add ruff for at least code formatting, and as a bonus linting

JacobCoffee:
#### What

- format with ruff

#### Close

- Closes #2539


JacobCoffee:
there's about 7000 issues found with ruff lint. taking away docstring and missing type annotations its not so bad at around ~500 so working through some of those to see if its something we want to do (i do 🥹)

ewdurbin:
Zero opposition, but please can handle all concerns then rebase this down to one commit so we can do something like https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

blogs/tests/test_parser.py
```diff
@@ -15,17 +14,13 @@ def setUpClass(cls):
...
-        self.assertEqual(
-            self.entries[0]['title'],
-            'Introducing Electronic Contributor Agreements'
-        )
+        self.assertEqual(self.entries[0]["title"], "Introducing Electronic Contributor Agreements")
         self.assertIn(
-            "We're happy to announce the new way to file a contributor "
-            "agreement: on the web at",
-            self.entries[0]['summary']
+            "We're happy to announce the new way to file a contributor " "agreement: on the web at",
```
hugovk, line 19: Turn on ISC to catch these:

```suggestion
            "We're happy to announce the new way to file a contributor agreement: on the web at",
```

https://docs.astral.sh/ruff/rules/#flake8-implicit-str-concat-isc

JacobCoffee:
this was done in #2882

PR #2554, opened 2024-09-13
Check collectstatic

ewdurbin:
Ensure that our collectstatic step can complete, since it isn't otherwise exercised before image build.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -1,0 +1,29 @@
+name: Check collectstatic
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.12.6
+      - name: Cache Python dependencies
+        uses: actions/cache@v4
+        env:
+          cache-name: pythondotorg-cache-pip
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-${{ github.job }}-${{ env.cache-name }}-${{ hashFiles('requirements.txt', '*-requirements.txt') }}
+          restore-keys: |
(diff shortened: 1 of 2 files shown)


.github/workflows/static.yml
```diff
@@ -0,0 +1,29 @@
...
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.12.6
```
hugovk: If you put this in a `.python-version`, you can remove it from here:

```suggestion
```

And here: https://github.com/python/pythondotorg/blob/019c0629fd98c7ed4005387eb4ece5182995b14f/.github/workflows/ci.yml#L33-L34

https://github.com/actions/setup-python#basic-usage

.github/workflows/static.yml
```diff
@@ -0,0 +1,29 @@
+name: Check collectstatic
+on: [push, pull_request]
```
hugovk, line 2: Allows triggering a build from the UI:
```suggestion
on: [push, pull_request, workflow_dispatch]
```

ewdurbin: Is this necessary? Its mainly just a check to ensure that a code change doesn't break our collect static step.

hugovk: Nope, definitely not necessary.

I generally just add it to most workflows, because when I do want to trigger one, it can be easier to click a button than make a dummy commit or re-run an old one.

Feel free to skip if it's not useful here.

JacobCoffee (approved):

.github/workflows/static.yml
ewdurbin: without _something_ set there's a warning

```
Warning: Neither 'python-version' nor 'python-version-file' inputs were supplied. Attempting to find '.python-version' file.
```

I'll explicitly set it.

PR #2555, opened 2024-09-13
upgrade allauth

ewdurbin:
bypass the migration issue around socialaccount... since we do not use that feature!


ewdurbin:
This is somewhat "urgent" since a slight bug in the early version of allauth's account middleware that we're using is causing `Vary: Cookie` to go out with nearly all HTTP responses and tanking our CDN caching.

Screenshot 2024-09-13 at 9 12 37 AM

ewdurbin:
Screenshot 2024-09-13 at 9 21 34 AM

ewdurbin:
Immediately better on deploy 💜

PR #2558, opened 2024-09-13
feat(#2492): use newest compose command

### LINKED ISSUE #2492: [Docker] Switch to the built-in compose plugin
I ran into this when trying to run the development setup for the first time.
The makefile uses the standalone `docker-compose` command which is now deprecated (some more information [here](https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose) and [here](https://docs.docker.com/compose/install/))

If you're running a more recent version of Docker, it's better to use the built-in plugin, i.e., `docker compose ...` instead.

This should in theory be a drop-in replacement - as a test I tried replacing all instances of `docker-compose` in the makefile with `docker compose` and the website builds without issues.

JacobCoffee:
#### Description

- Update to new compose subcommand

#### Closes

- Closes #2492

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ default:
 
 .state/docker-build-web: Dockerfile dev-requirements.txt base-requirements.txt
 	# Build web container for this project
-	docker-compose build --force-rm web
+	docker compose build --force-rm web
 
 	# Mark the state so we don't rebuild this needlessly.
 	mkdir -p .state && touch .state/docker-build-web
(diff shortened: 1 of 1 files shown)


JacobCoffee:
Unsure of the impact on https://github.com/python/pythondotorg/blob/624dc3a3a5b0861b88a8cebfc76dc51bb33361d8/docker-compose.yml#L27 but make targets  themselves ran 
```
✗ make clean && make serve
docker compose down -v
```

ewdurbin:
> Unsure of the impact on
> 
> https://github.com/python/pythondotorg/blob/624dc3a3a5b0861b88a8cebfc76dc51bb33361d8/docker-compose.yml#L27

That's just the resulting tagged name of the image so it can be re-used in the worker container. So all is well.

PR #2565, opened 2024-09-16
Pages: Also purge trailing slash

ewdurbin:
Fastly purge requests are sensitive to trailing slash, so to ensure we get the result we want we should also purge that.

### DIFF
--- a/pages/models.py
+++ b/pages/models.py
@@ -137,6 +137,8 @@ def purge_fastly_cache(sender, instance, **kwargs):
     Requires settings.FASTLY_API_KEY being set
     """
     purge_url(f'/{instance.path}')
+    if not instance.path.endswith('/'):
+        purge_url(f'/{instance.path}/')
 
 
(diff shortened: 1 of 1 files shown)


pages/models.py
```diff
@@ -137,6 +137,8 @@ def purge_fastly_cache(sender, instance, **kwargs):
     Requires settings.FASTLY_API_KEY being set
     """
     purge_url(f'/{instance.path}')
+    if not instance.path.endswith('/'):
```
JacobCoffee, line 140: do we need to do this in the job models, too?

JacobCoffee (approved):

pages/models.py
ewdurbin: Those urls are already properly resolved with a trailing slash, see

https://github.com/python/pythondotorg/blob/main/jobs/urls.py#L27

for example, which is purged via https://github.com/python/pythondotorg/blob/main/jobs/models.py#L274.

PR #2568, opened 2024-09-16
fix: recreate dictionary each run

JacobCoffee:
#### Description

- TF cloud runs fail due to the dictionary not being editable
image

### DIFF
--- a/infra/cdn/main.tf
+++ b/infra/cdn/main.tf
@@ -392,8 +392,6 @@ resource "fastly_service_vcl" "python_org" {
       product_enablement,
     ]
   }
-
-  force_destroy = true
 }
 
 output "service_id" {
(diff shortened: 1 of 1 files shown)


infra/cdn/main.tf
```diff
@@ -391,6 +392,8 @@ resource "fastly_service_vcl" "python_org" {
       product_enablement,
     ]
   }
+
+  force_destroy = true
```
JacobCoffee, line 396:
```suggestion
```

infra/cdn/main.tf
```diff
@@ -391,6 +392,8 @@ resource "fastly_service_vcl" "python_org" {
...
 }
```
JacobCoffee:
```suggestion
  }
}
```

ewdurbin (approved):

PR #2569, opened 2024-09-16
feat(#1612): add rss feed for latest downloads

### LINKED ISSUE #1612: An up-to-date and reliable feed (atom/rss) for python releases
There seems to be no really easy and obvious way to automate the monitoring of python releases.

It would be great  if there was an up-to-date and reliable feed for python releases.

JacobCoffee:
#### Description

- Adds RSS feed URL for latest releases
image

#### Closes

- Closes #1612

### DIFF
--- a/downloads/tests/test_views.py
+++ b/downloads/tests/test_views.py
@@ -554,3 +554,46 @@ class DownloadApiV2ViewsTest(BaseDownloadApiViewsTest, BaseDownloadTests, APITes
             headers={"authorization": self.Authorization}
         )
         self.assertEqual(response.status_code, 405)
+
+class ReleaseFeedTests(BaseDownloadTests):
+    """Tests for the downloads/feed.rss endpoint.
+
+    Content is ensured via setUp in BaseDownloadTests.
+    """
+
+    def setUp(self) -> None:
+        super().setUp()
+        self.url = reverse("downloads:feed")
+
+    def test_endpoint_reachable(self) -> None:
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+
+    def test_feed_content(self) -> None:
+        """Ensure feed content is as expected.
+
+        Some things we want to check:
+         - Feed title, description, pubdate
+         - Feed items (releases) are in the correct order
+         - We get the expected number of releases (10)
+        """
+        response = self.client.get(self.url)
+        content = response.content.decode()
+
+        self.assertIn("Python 2.7.5", content)
+        self.assertIn("Python 3.10", content)
+        # Published but hidden show up in the API and thus the feed
+        self.assertIn("Python 0.0.0", content)
+
+        # No unpublished releases
+        self.assertNotIn("Python 9.7.2", content)
+
+        # Pre-releases are shown
+        self.assertIn("Python 3.9.90", content)
+
+    def test_feed_item_count(self) -> None:
+        response = self.client.get(self.url)
+        content = response.content.decode()
+
+        # In BaseDownloadTests, we create 5 releases, 4 of which are published, 1 of those published are hidden..
+        self.assertEqual(content.count("<item>"), 4)
--- a/downloads/tests/test_views.py
+++ b/downloads/tests/test_views.py
@@ -561,9 +561,8 @@ class ReleaseFeedTests(BaseDownloadTests):
     Content is ensured via setUp in BaseDownloadTests.
     """
 
-    def setUp(self) -> None:
-        super().setUp()
-        self.url = reverse("downloads:feed")
+    url = reverse("downloads:feed")
+
 
     def test_endpoint_reachable(self) -> None:
         response = self.client.get(self.url)
--- a/downloads/urls.py
+++ b/downloads/urls.py
@@ -9,4 +9,5 @@ urlpatterns = [
     path('release/<slug:release_slug>/', views.DownloadReleaseDetail.as_view(), name='download_release_detail'),
     path('<slug:slug>/', views.DownloadOSList.as_view(), name='download_os_list'),
     path('', views.DownloadHome.as_view(), name='download'),
+    path("feed.rss", views.ReleaseFeed(), name="feed"),
 ]
--- a/downloads/views.py
+++ b/downloads/views.py
@@ -1,4 +1,3 @@
-import logging
 from typing import Any
 
 import requests
@@ -16,7 +15,6 @@ from django.utils.feedgenerator import Rss201rev2Feed
 
 from .models import OS, Release, ReleaseFile
 
-logger = logging.getLogger(__name__)
 
 class DownloadLatestPython2(RedirectView):
     """ Redirect to latest Python 2 release """
--- a/downloads/views.py
+++ b/downloads/views.py
@@ -210,12 +210,15 @@ class ReleaseFeed(Feed):
         """Return the URL to the release page on python.org."""
         return reverse("downloads:download_release_detail", args=[item.slug])
 
-    def item_pubdate(self, item: Release) -> datetime:
+    def item_pubdate(self, item: Release) -> datetime | None:
         """Return the release date as the item publication date."""
-        return timezone.make_aware(item.release_date) if item.release_date else None
+        if item.release_date:
+            if timezone.is_naive(item.release_date):
+                return timezone.make_aware(item.release_date)
+            return item.release_date
+        return None
 
-    @staticmethod
-    def item_guid(item: Release) -> str:
+    def item_guid(self, item: Release) -> str:
         """Return a unique ID for the item based on DB record."""
         return str(item.pk)
 
--- a/downloads/views.py
+++ b/downloads/views.py
@@ -185,16 +185,6 @@ class ReleaseFeed(Feed):
         """Return the latest Python releases."""
         return Release.objects.filter(is_published=True).order_by("-release_date")[:10]
 
-    @staticmethod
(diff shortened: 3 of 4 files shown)


downloads/views.py
```diff
@@ -147,3 +159,92 @@ def get_context_data(self, **kwargs):
...
+            logger.error(f"Unexpected error in items method: {str(e)}")
+        return []
+
+    @staticmethod
+    def _fetch_releases(url: str) -> list[dict[str, Any]]:
+        """Grabs the latest Python releases from API.
+
+
+        """
+        response = requests.get(url, timeout=10)
```
JacobCoffee:
```suggestion
        """Grabs the latest Python releases from API."""
        response = requests.get(url, timeout=10)
```

downloads/views.py
```diff
@@ -147,3 +159,92 @@ def get_context_data(self, **kwargs):
...
+    def link() -> str:
+        """Return the URL to the main downloads page."""
+        return reverse('downloads:download')
+
+    def get_feed(self, obj: Any, request: WSGIRequest) -> Feed:
+        """Store the request object for later use."""
+        self.request = request
+        return super().get_feed(obj, request)
+
+    def items(self) -> list[dict[str, Any]]:
```
ewdurbin, line 176: Seems really odd to pull these by making a request from ourselves rather than using a database query. Can you help me understand why you approached it this way?

downloads/views.py
```diff
@@ -147,3 +159,92 @@ def get_context_data(self, **kwargs):
...
+            return pytz.utc.localize(datetime.now())
+
+    @staticmethod
+    def item_guid(item: dict[str, Any]) -> str:
+        """Return the release URI as the item GUID."""
+        return item.get("resource_uri", "")
+
+    def create_url(self, path: str) -> str:
+        """Create a full URL using the current site domain."""
+        current_site = get_current_site(self.request)
```
JacobCoffee: it would be cool to have localdev site set to not example.com, which i think would require a data migration or maybe we can edit fixture data?

downloads/views.py
JacobCoffee: bad habits :) i will learn the better way

ewdurbin: I'd recommend replacing this with a Django ORM query, which would allow for cleaning up the static methods throughout this class for constructing urls and making requests.

JacobCoffee: [44fde54](https://github.com/python/pythondotorg/pull/2569/commits/44fde54bcd34086ea351c9c113d430577e59c1b6)

downloads/urls.py
```diff
@@ -9,4 +9,5 @@
     path('release/<slug:release_slug>/', views.DownloadReleaseDetail.as_view(), name='download_release_detail'),
     path('<slug:slug>/', views.DownloadOSList.as_view(), name='download_os_list'),
     path('', views.DownloadHome.as_view(), name='download'),
+    path("feed.rss", views.ReleaseFeed(), name="feed"),
```
hugovk, line 12: Would this URL be https://www.python.org/downloads/feed.rss?

Thinking about consistency with our existing feeds, we have these listed in each page's `<head>`:

```html
    <link rel="alternate" type="application/rss+xml" title="Python Enhancement Proposals"
          href="https://peps.python.org/peps.rss">
    <link rel="alternate" type="application/rss+xml" title="Python Job Opportunities"
          href="https://www.python.org/jobs/feed/rss/">
    <link rel="alternate" type="application/rss+xml" title="Python Software Foundation News"
          href="https://feeds.feedburner.com/PythonSoftwareFoundationNews">
    <link rel="alternate" type="application/rss+xml" title="Python Insider"
          href="https://feeds.feedburner.com/PythonInsider">
```

They're all quite different. https://peps.python.org/peps.rss is generated by the https://github.com/python/peps repo. Those feedburner are external. And https://www.python.org/jobs/feed/rss/ is a pretty unusual looking URL for a feed!

So I think it's fine to pick a new scheme for future feeds. There's a lot of variation in feed, see https://blog.jim-nielsen.com/2021/feed-urls/, and something like https://www.python.org/downloads/feed.rss looks good for a feed of data from https://www.python.org/downloads/, and also fits that analysis's conclusions of `mysite.com/{links,posts,notes,podcasts}/feed.{rss,atom,json}`.

---

Let's also add the new feed to the `<head>` for discoverability.

JacobCoffee: > Would this URL be [python.org/downloads/feed.rss](https://www.python.org/downloads/feed.rss)?

Yep!

downloads/tests/test_views.py
```diff
@@ -554,3 +554,46 @@ def test_filter_release_file_delete_by_release(self):
...
+
+class ReleaseFeedTests(BaseDownloadTests):
+    """Tests for the downloads/feed.rss endpoint.
+
+    Content is ensured via setUp in BaseDownloadTests.
+    """
+
+    def setUp(self) -> None:
+        super().setUp()
+        self.url = reverse("downloads:feed")
```
ewdurbin: This can be defined as a class attribute, which is a bit cleaner.

```
class ReleaseFeedTests(BaseDownloadTests):
    url = reverse("downloads:feed")
    ...
```

and then there's no need to define a setUp.

ewdurbin (commented):
Just a review on the initial tests.

downloads/views.py
```diff
@@ -147,3 +157,70 @@ def get_context_data(self, **kwargs):
...
+        return reverse("downloads:download")
+
+    def get_feed(self, obj: Any, request: WSGIRequest) -> Feed:
+        """Store the request object for later use."""
+        self.request = request
+        return super().get_feed(obj, request)
+
+    def items(self) -> list[dict[str, Any]]:
+        """Return the latest Python releases."""
+        return Release.objects.filter(is_published=True).order_by("-release_date")[:10]
```
ewdurbin, line 178: Using the DB query should mean that `_fetch_releases` and `create_url` can be deleted, right?

JacobCoffee: should be done in [075ef04](https://github.com/python/pythondotorg/pull/2569/commits/075ef047b6463570bce1c85d4f87946a4d28e6ec)

downloads/tests/test_views.py
JacobCoffee: should be done in [075ef04](https://github.com/python/pythondotorg/pull/2569/commits/075ef047b6463570bce1c85d4f87946a4d28e6ec)

downloads/views.py
```diff
@@ -147,3 +155,54 @@ def get_context_data(self, **kwargs):
...
+    description = "Latest Python releases from Python.org"
+
+    @staticmethod
+    def link() -> str:
+        """Return the URL to the main downloads page."""
+        return reverse("downloads:download")
+
+    def get_feed(self, obj: Any, request: WSGIRequest) -> Feed:
+        """Store the request object for later use."""
+        self.request = request
```
ewdurbin: We don't appear to use this at any point elsewhere in the code. Can we drop this method?

downloads/views.py
```diff
@@ -147,3 +155,54 @@ def get_context_data(self, **kwargs):
...
+        using Django's Syndication framework.
+        Docs: https://docs.djangoproject.com/en/4.2/ref/contrib/syndication/
+    """
+
+    feed_type = Rss201rev2Feed
+    title = "Python Releases"
+    description = "Latest Python releases from Python.org"
+
+    @staticmethod
+    def link() -> str:
```
ewdurbin, line 172: Similarly I don't see this being used.

downloads/views.py
```diff
@@ -147,3 +155,54 @@ def get_context_data(self, **kwargs):
...
+
+    def item_title(self, item: Release) -> str:
+        """Return the release name as the item title."""
+        return item.name
+
+    def item_description(self, item: Release) -> str:
+        """Return the release version and release date as the item description."""
+        return f"Version: {item.version}, Release Date: {item.release_date}"
+
+    def item_link(self, item: Release) -> str:
```
ewdurbin: This should not be necessary since Release defines a `get_absolute_url` method.

ewdurbin: See https://docs.djangoproject.com/en/4.2/ref/contrib/syndication/

downloads/views.py
JacobCoffee: i think this is used by the super class?

downloads/views.py
JacobCoffee: [9746cb8](https://github.com/python/pythondotorg/pull/2569/commits/9746cb893e773aaa1f35be4930614fff01d8897f)

ewdurbin:
Probably want to add the RSS feed to the purge method [here](https://github.com/python/pythondotorg/blob/main/downloads/models.py#L263) so it is purged when new releases are uploaded.

downloads/views.py
ewdurbin: Oh, I see, sorry!

> Returns the URL of the HTML version of the feed as a normal Python string.

Probably do wanna keep that.

downloads/views.py
JacobCoffee: [41e28d5](https://github.com/python/pythondotorg/pull/2569/commits/41e28d554ae9e9699651729ec8fd31ee94c9c013)

JacobCoffee:
@ewdurbin [4cdd90a](https://github.com/python/pythondotorg/pull/2569/commits/4cdd90ac76ffefb23024ce7fa0f18c52d12275c4)

ewdurbin (approved):

PR #2579, opened 2024-09-17
add a name for the ngwaf edge dictionary

ewdurbin:
#### Description

Fastly requires a name for edge dictionaries.


JacobCoffee (approved):

ewdurbin:
Bypassing CI tests since this is terraform only

PR #2580, opened 2024-09-17
infra: activate by default

JacobCoffee:
#### Description

- activate


infra/cdn/main.tf
```diff
@@ -392,6 +392,8 @@ resource "fastly_service_vcl" "python_org" {
       product_enablement,
     ]
   }
+
+  force_destroy = true
```
JacobCoffee:
```suggestion
  }
```

ewdurbin (approved):

PR #2588, opened 2024-09-18
fix: guid must be a full URL, use item link url

### LINKED ISSUE #2585: Bug: Downloads RSS feed doesnt fully validate
Although it doesn't validate:

```
line 2, column 609: guid must be a full URL, unless isPermaLink attribute is false: 925 (10 occurrences) [help]

... p 2024 01:32:54 +0000</pubDate><guid>925</guid></item><item><title>Pytho ...
```

https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fwww.python.org%2Fdownloads%2Ffeed.rss

https://validator.w3.org/feed/docs/error/InvalidHttpGUID.html says:

> `guid` must be a full URL, unless isPermaLink attribute is false

I think it would be most useful to re-use the URL from `link`: 

```html
<link>https://www.python.org/downloads/release/python-31015/</link>
```
_Originally posted by @hugovk in https://github.com/python/pythondotorg/issues/1612#issuecomment-2359232071_

JacobCoffee:
#### Description

- Fixes guid by just using the link of the item

#### Closes

- Resolves #2585

### DIFF
--- a/downloads/views.py
+++ b/downloads/views.py
@@ -192,11 +192,3 @@ class ReleaseFeed(Feed):
                 return timezone.make_aware(item.release_date)
             return item.release_date
         return None
-
-    def item_guid(self, item: Release) -> str:
-        """Return a link as the GUID for the item.
-
-        ``guid`` must be a full URL, unless `isPermaLink` attribute is false
-        https://validator.w3.org/feed/docs/error/InvalidHttpGUID.html
-        """
-        return str(item.get_absolute_url())
--- a/downloads/views.py
+++ b/downloads/views.py
@@ -199,4 +199,4 @@ class ReleaseFeed(Feed):
         ``guid`` must be a full URL, unless `isPermaLink` attribute is false
         https://validator.w3.org/feed/docs/error/InvalidHttpGUID.html
         """
-        return str(self.link())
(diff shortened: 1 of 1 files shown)


downloads/views.py
```diff
@@ -194,5 +194,9 @@ def item_pubdate(self, item: Release) -> datetime | None:
...
 
     def item_guid(self, item: Release) -> str:
-        """Return a unique ID for the item based on DB record."""
-        return str(item.pk)
+        """Return a link as the GUID for the item.
+
+        ``guid`` must be a full URL, unless `isPermaLink` attribute is false
+        https://validator.w3.org/feed/docs/error/InvalidHttpGUID.html
+        """
+        return str(self.link())
```
ewdurbin:
```suggestion
        return str(item.get_absolute_url())
```

or similar. I think we want the item_guid to map to a specific release.

JacobCoffee: ooo

ewdurbin:
Docs say `item_guid` is optional, and if not provided the items link is used by default. Should we just remove the method override?

JacobCoffee:
@ewdurbin that would be okay, too. ill do that

ewdurbin (approved):

PR #2591, opened 2024-09-19
ci: add pr combiner

JacobCoffee:
#### Description

- Uses 3rd party GH action to combine dependabot PRs weekly 
Other Option #2592


ezio-melotti:
Is there any actual reason to prefer this approach to #2592?

PR #2592, opened 2024-09-19
ci: add native pr combiner

JacobCoffee:
#### Description

- Updates config to combine dependabot PRs weekly 

Other Option #2591


shenxianpeng:
Copied the following from [combine-prs](https://github.com/github/combine-prs?tab=readme-ov-file#notice-) README

> As mentioned above is this README, a core reason why this Action exists is to "combine multiple Dependabot PRs into one". Work for this Action was completed before the [GitHub Blog Post](https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/) was published and the Dependabot Grouped Version Updates feature was released. While it may seem like this Action is no longer needed due to this feature, there are actually still quite a few use cases for this Action. The first one that is front of mind is that the PRs which Dependabot opens are grouped by package manager. This means that if you have a project that uses multiple package managers, you'll still end up with multiple PRs. This Action can be used to combine those PRs into a single one. You may also want to combine pull requests that are not related to Dependabot in anyway, which this Action can also do.

Since the rest only uses the pip package manager, it might be a good idea to use `groups`, as demonstrated in this repository:

https://github.com/python/pythondotorg/blob/main/.github/dependabot.yml#L28-L31

ezio-melotti (approved):

hugovk (approved):

JacobCoffee:
I haven't merged this yet because we have a bunch of packages from the Django 2 migration that were bare-minimum bumped up to where they support 4.2, and as you may see in the PRs there are many dependabots saying "bump projectX==3.123 to 8.1!" and i dont want this to bring in all that mess and stomp on the important ones i do have time to look at and merge (CVE-related mostly)

I will continue my work in https://github.com/python/pythondotorg/pull/2741 when pycon craziness settles or when i get a wild hair at night as part of cpython triage (I don’t know if that counts? its fun though either way!) and THEN work on bumping packages up and then finally merge this... unless someone has a better idea or thinks i am thinking this through poorly :)

PR #2593, opened 2024-09-19
fix: links are direct to asset

### LINKED ISSUE #2577: Bug: View provided assets button takes you to the full assets list
#### What

When a sponsor is viewing their profile, `users/sponsorships/$ID`,they can click into a specific asset:

image

However, clicking this link browses to `users/sponsorships/$ID/provided-assets/?provided_asset=##` where the query passed, `?provided_asset=##` does nothing. 

#### Expected Behavior

When you click “view asset” , go to only the single specific asset and not the list of all assets

#### Extra

- Get rid of “Or you can also[ click here](https://www.python.org/users/sponsorships/266/provided-assets/) to view all the assets under the same page?

JacobCoffee:
#### Description

- Fixes asset links on sponsor user page

#### Closes

- Closes: #2577

#### funny

- code change: 30 minutes
- tests: 1.5 hours


templates/users/sponsorship_detail.html
```diff
@@ -114,9 +114,17 @@ <h3>Provided Assets</h3>
           <p><small>Assets from the PSF related to your sponsorship.</small></p>
           <br/>
           <ul>
-            {% for asset in provided_assets %}
-              <li><b>{{ asset.label }}</b>: <a href="{{ asset.user_view_url }}">View asset</a>.</li>
-            {% endfor %}
+              {% for asset in provided_assets %}
```
ewdurbin: We're effectively displaying all the same information here now that was previously available in the `users:view_provided_assets` view.

Should we just replicate the template logic from https://github.com/python/pythondotorg/blob/ccbc9e42488f2bf6c12831c5e12d31f5eec4c2a7/templates/users/sponsorship_assets_view.html#L21-L32 and remove the "Or you can also click here to view all the assets under the same page" part?

ewdurbin: Note only reason I bring up replicating the template logic is to restore the help text and context for each.

Compare

Screenshot 2024-09-23 at 8 28 40 AM

on the existing "view all" page with 

Screenshot 2024-09-23 at 8 29 00 AM

On the new dashboard section.

JacobCoffee: It did add a chunk of text if that is a concern. Keeping the core dashboard leaner feels better imo but @loren-c may like the all-in-one approach :D

JacobCoffee: Current:
image

@ewdurbin Suggested:
image

As an aside I think this could use some rearranging or love because its very squished and would look better some other way

JacobCoffee: Something like
image


Seems one of my main issues comes from https://github.com/python/pythondotorg/blob/301bfe2eee1050a848e48cef9f2c1eea690a3b04/templates/users/base.html#L13 because we just leave a ton of empty space when there is no sidebar

JacobCoffee: or 
image

loren-c: I do like the 'all in one' approach provided we can make it more manageable and not just a giant list, and I do like Jacob's second version. 

Looking at the dashboard with all the information on one page, an ability to sort the required assets by due date or by incomplete/complete would be nice, but that seems like a whole other functionality and just a nice to have.

JacobCoffee:
Changes in [e53f422](https://github.com/python/pythondotorg/pull/2593/commits/e53f422f2b1b2558f54f04b81570f7384df4c806)

Although I am not 100%, my checks into the various user views for the removal of`with-right-sidebar` have shown that everything just fills the empty space instead of wasting it. Have yet to find a page that uses it but I will continue to check.

How it looks ish (latest changes I removed excessive padding as you can see in image 1 but further images have it in screenshot.. has since been fixed)

<details>
<summary>Details</summary>


image

image

image

image


</details>

static/sass/style.css
```diff
@@ -4000,25 +4000,157 @@ span.highlighted {
...
+    display: inline-block;
+    padding: 0.375rem 0.75rem;
+    font-size: 0.9rem;
+    text-align: center;
+    text-decoration: none;
+    border-radius: 4px;
+    transition: background-color 0.2s ease;
+}
+
+.btn-link {
```
JacobCoffee: could maybe make this more fancy

static/sass/style.css
```diff
@@ -4000,25 +4000,157 @@ span.highlighted {
 .hidden {
```
ewdurbin, line 3979: Just an overall concern of modifying the CSS without updating the relevant SCSS file (styles.scss). This is something we should avoid!

JacobCoffee: ahh sorry, im still not familiar with this SASS stuff 😬

ewdurbin: Reason: the CSS files here are the result of compiling the SCSS files. we would either need to abandon the SCSS pipeline completely or the next time SCSS is compiled by someone it will blow away these changes.

JacobCoffee: I think ive done this. Reverted all CSS changes in  [097fafc](https://github.com/python/pythondotorg/pull/2593/commits/097fafc5081a76b2216f3910e36eb3c1c4f2b587) then ran the sass compile after i migrated my changes over in [1ab3053](https://github.com/python/pythondotorg/pull/2593/commits/1ab3053e364fab77e2334bd0d94c9a57691fd12b)

JacobCoffee: image

JacobCoffee:
@loren-c requested a more neutral benefits view instead of the green checks, and also asked if we could include the category. 

image

image


[45d654f](https://github.com/python/pythondotorg/pull/2593/commits/45d654fd9faebb1e5cc70184d905bc40b88c73f0)

loren-c:
the new version looks great to me!

templates/users/base.html
```diff
@@ -10,7 +10,8 @@
 {% endblock %}
 
 
-{% block content_attributes %}with-right-sidebar{% endblock %}
```
ewdurbin, line 13: It's fine to remove this rather than comment it.

JacobCoffee:
ready for review then, @ewdurbin :)

ewdurbin (approved):

PR #2609, opened 2024-09-25
Improve static deploy

ewdurbin:
#### Description

During deploys, we sometimes end up with 404s for static files in CDN cache. Add a surrogate-key to static files using white noise's machinery and a post deploy task that will run after all the new backends are up to purge that key.

### DIFF
--- a/fastly/utils.py
+++ b/fastly/utils.py
@@ -20,3 +20,23 @@ def purge_url(path):
         return response
 
     return None
+
+
+def purge_surrogate_key(key):
+    """
+    Purge a Fastly.com Surrogate-Key given a key.
+    """
+    if settings.DEBUG:
+        return
+
+    api_key = getattr(settings, 'FASTLY_API_KEY', None)
+    service_id = getattr(settings, 'FASTLY_SERVICE_ID', None)
+    if api_key and service_id:
+        response = requests.request(
+            "POST",
+            f'https://api.fastly.com/service/{service_id}/purge/{key}',
+            headers={'Fastly-Key': api_key},
+        )
+        return response
+
+    return None
--- a/pydotorg/settings/base.py
+++ b/pydotorg/settings/base.py
@@ -285,8 +285,10 @@ PYTHON_BLOG_URL = "https://blog.python.org"
 MAILING_LIST_PSF_MEMBERS = "psf-members-announce-request@python.org"
 
 ### Fastly ###
-FASTLY_API_KEY = False  # Set to Fastly API key in production to allow pages to
-                        # be purged on save
+FASTLY_SERVICE_ID = False  # Set to a Fastly Service ID in production to allow
+                           # purges by Surrogate-Key
+FASTLY_API_KEY = False     # Set to Fastly API key in production to allow
+                           # pages to be purged on save
 
 # Jobs
 JOB_THRESHOLD_DAYS = 90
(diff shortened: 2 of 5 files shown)


ewdurbin:
Note, this needs a little bit more review before it's ready. Just getting the idea out there quickly.

fastly/utils.py
```diff
@@ -20,3 +20,23 @@ def purge_url(path):
...
+def purge_surrogate_key(key):
+    """
+    Purge a Fastly.com Surrogate-Key given a key.
+    """
+    if settings.DEBUG:
+        return
+
+    api_key = getattr(settings, 'FASTLY_API_KEY', None)
+    service_id = getattr(settings, 'FASTLY_SERVICE_ID', None)
+    if api_key and service_id:
```
JacobCoffee, line 34: should we do response code checking here to catch issues?

fastly/utils.py
```diff
@@ -20,3 +20,23 @@ def purge_url(path):
...
+
+def purge_surrogate_key(key):
+    """
+    Purge a Fastly.com Surrogate-Key given a key.
+    """
+    if settings.DEBUG:
+        return
+
+    api_key = getattr(settings, 'FASTLY_API_KEY', None)
+    service_id = getattr(settings, 'FASTLY_SERVICE_ID', None)
```
JacobCoffee, line 33: why getattr instead of grabbing directly as we do `DEBUG`?
`settings.FASTLY_API_KEY, settings.FASTLY_SERVICE_ID`

pydotorg/settings/base.py
```diff
@@ -349,6 +351,10 @@
 
 GLOBAL_SURROGATE_KEY = 'pydotorg-app'
 
+### pydotorg.settings.cabotage.add_surrogate_keys_to_static
+
```
JacobCoffee, line 355: do we clear caches on deploy for pydotorg-app? Just wondering if we could just do it all in one instead of special casing static files

JacobCoffee (approved):
i think this looks good from what i know

fastly/utils.py
ewdurbin: copied from above function, but I think the point is that settings _may not_ have these values set at all, which would lead to an AttributeError. this approach is similar to a `{}.get(key, None)`

```
>>> from django.conf import settings
>>> settings.DOES_NOT_EXIST
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python3.12/site-packages/django/conf/__init__.py", line 104, in __getattr__
    val = getattr(_wrapped, name)
          ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Settings' object has no attribute 'DOES_NOT_EXIST'
>>> getattr(settings, 'DOES_NOT_EXIST', None)
>>> 
```

fastly/utils.py
ewdurbin: Maybe worth doing a retry, but in general I'm unsure. It's going to run automated/unsupervised so if we do consider the response code the error handling needs to be automatic.

In general, fastly's API has pretty great availability (we purge hundreds of times a minute for PyPI without seeing incidents often).

pydotorg/settings/base.py
ewdurbin: No, and that's intentional. Purging `pydotorg-app` is more of a "break glass" situation since that key covers hundreds of pages that will need to be re-populated.

PR #2622, opened 2024-09-30
feat(infra): do not cache static 404s

JacobCoffee:
#### Description

- Dont cache /static 404s based on https://docs.fastly.com/en/guides/about-conditions


infra/cdn/main.tf
```diff
@@ -129,6 +129,13 @@ resource "fastly_service_vcl" "python_org" {
     type      = "REQUEST"
   }
 
+  condition {
```
ewdurbin, line 140: How is this condition applied?

I think this needs an associated [`cache_setting`](https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_vcl#nestedblock--cache_setting)

JacobCoffee: [926b463](https://github.com/python/pythondotorg/pull/2622/commits/926b463fbb4addfc4ceab29d4547a8876918187a) 😬

ewdurbin (approved):

Issue #2631, opened 2024-10-04
Bug: Main Page and Download Python Source Still Reference 3.12.6


questionlp:
### Describe the bug

With Python 3.12.7 released on 2 Oct 2024, the latest version listed in the Download block on the main page still lists 3.12.6. Also, the Downloads popup in the site navigation also references 3.12.6.

### To Reproduce

Downloads navigation popup

1. Open a web browser
2. Browse to <https://www.python.org>
3. Hover cursor over the **Downloads** navigation item

Downloads Latest

1. Open a web browser
2. Browse to <https://www.python.org>
3. Scroll down to the **Get Started** and **Downloads** blocks, Latests lists the previous version

### Expected behavior

In both locations, 3.12.7 should be listed as the latest version and should link to the appropriate [release page](https://www.python.org/downloads/release/python-3127/) or to the appropriate [release tarball](https://www.python.org/ftp/python/3.12.7/Python-3.12.7.tar.xz)

### URL to the issue

https://www.python.org/

### Screenshots

![image](https://github.com/user-attachments/assets/224d68ea-5fd3-40c9-862a-14b2e72fdd25)

![image](https://github.com/user-attachments/assets/fa2842f7-c57f-4991-84db-2c6aadb01019)

### Browsers

Firefox, Chrome

### Operating System

Linux

### Browser Version

130

ned-deily:
@JacobCoffee, any ideas? Everything looks OK from the admin side and at least the generated homepage-downloads box shows 3.12.7. Cache issue? Generating the web page issue?

JacobCoffee:
Not sure! I'm checking.

I believe this comes from from 
- [downloads/models.py:178](https://github.com/python/pythondotorg/blob/0d5432a04dbfb78b582c808856f912b412b1201b/downloads/models.py#L178)
- https://github.com/python/pythondotorg/blob/802aa1e9c7499fd178bad75b7768760ef4c0b049/templates/downloads/supernav.html#L11

JacobCoffee:
I did purge the Fastly cache with no luck

ned-deily:
I'm pretty sure we've had problems in the past with the download button not updating right away but I think it has been working fine in recent releases. Perhaps something was inadvertently broken by the recent website updates? In any case, it can wait until daylight.

JacobCoffee:
If I dive in manually into the shell for this service and run that supernav update I get 3.12.7 :\

```
# Shell Plus Django Imports
[...]

PR #2638, opened 2024-10-07
fix: Ensure that Box content is rendered after update

ewdurbin:
This resolves the underlying problem from #2631.

When updated via Django's [`update_or_create`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#update-or-create), the new `content` for a `Box` is not rendered by django-markupfield's [`pre_save`](https://github.com/jamesturk/django-markupfield/blob/2.0.1/markupfield/fields.py#L163-L179) method unless `save()` is called.

This is due to the fact that [`update`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.update) bypasses calls to `save`, `pre_save`, and `post_save`, from the docs:

> Finally, realize that update() does an update at the SQL level and, thus, does not call any save() methods on your models, nor does it emit the pre_save or post_save signals (which are a consequence of calling Model.save())

### DIFF
--- a/blogs/parser.py
+++ b/blogs/parser.py
@@ -48,10 +48,12 @@ def update_blog_supernav():
         pass
     else:
         rendered_box = _render_blog_supernav(latest_entry)
-        box, _ = Box.objects.update_or_create(
+        box, created = Box.objects.update_or_create(
             label='supernav-python-blog',
             defaults={
                 'content': rendered_box,
                 'content_markup_type': 'html',
             }
         )
+        if not created:
+            box.save()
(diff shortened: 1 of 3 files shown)


ewdurbin:
Summary here is that since the [`/boxes` view](https://github.com/python/pythondotorg/blob/ae76ce1104685aff13136cd99289195434ead7ab/boxes/views.py#L7) and [templatetag](https://github.com/python/pythondotorg/blob/ae76ce1104685aff13136cd99289195434ead7ab/boxes/templatetags/boxes.py#L15) sensibly use `Box.content.rendered` and not `Box.content`, the `pre_save` method is critical to ensuring that the value passed to content is actually... well... rendered.

JacobCoffee (approved):

Issue #2636, opened 2024-10-07
Bug: Python Insider sidebar outdated (twitter, mailing list)


tosh:
### Describe the bug

The "subscriptions" sidebar on [/blogs](https://www.python.org/blogs/) links to a twitter account that is no longer maintained (last post October 2020) and to a Python-Dev mailing list archive (that ends in June 2019).

![Screenshot 2024-10-07 at 18 16 11](https://github.com/user-attachments/assets/ca590385-d198-449e-bb9f-5ab8ba1930e7)

The same applies to the sidebar of the Python Insider blog on blogger:

![Screenshot 2024-10-07 at 18 22 56](https://github.com/user-attachments/assets/728bd599-20f1-4c46-843a-6f8fab68cac0)

I'm not sure what these links should be updated to or what they should be replaced with but it seems a bit misleading to endorse them so prominently when they are no longer maintained.

### To Reproduce

go to https://www.python.org/blogs/ and https://blog.python.org and look at the sidebars and follow the links.

### Expected behavior

Useful links for people to learn more or stay up to date (?)

### URL to the issue

https://www.python.org/blogs/

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Browsers

Safari

### Operating System

macOS

### Browser Version

18.1

PR #2652, opened 2024-11-06
Changed links on navbar from an outdated twitter page to a more recen…

marvinl803:
This solves issue #2636.

I have updated the links in the sidebars to the official Python Software Foundation (PSF) Twitter account (twitter.com/thespf), which is currently active and better represents the official social media presence for Python. This ensures that the sidebar links now lead to more relevant, up-to-date resources for users.


JacobCoffee:
Thanks for the PR! I know it's confusing but box data is actually controlled via Django Admin and the fixture data is just exports of that for easy bootstrapping of development environments; this means that your PR would have no effect :\

We could make the relevant changes at `boxes/box/40/change/` though depending on what @ewdurbin thinks.

hugovk:
The box is specifically about Python Insider so I suggest removing the old Twitter link. 

Another option is to replace it with something else. It's easy to hook up an RSS feed to Mastodon using https://feedsin.space

I used it for [@peps](https://feedsin.space/feed/peps) and [@python_releases](https://feedsin.space/feed/python_releases) and it would be good to do one for Insider too.

ewdurbin:
> We could make the relevant changes at `boxes/box/40/change/` though depending on what @ewdurbin thinks.

Updating the boxes fixture in addition to making the change in Django admin for prod is a good way of documenting what's changed. So if we settle on a correct solution here, update the box in admin then merge the PR.

PR #2653, opened 2024-11-07
disable logging to datadog by default

### LINKED ISSUE #2651: Investigate logs forwarding to Datadog after IaC conversion
something about our fastly configuration changed Oct 21 (when we moved to terraform handling) that caused _all_ logs to go to datadog, not just the rate limit events.

is that something you could investigate?

[Slack Message](https://thepsf.slack.com/archives/G7MG5TXNX/p1730840056006569?thread_ts=1730839703.152459&cid=G7MG5TXNX)

ewdurbin:
Resolves #2651. Determined by downloading the VCL for current service version and the version active before #2519.

I think this is a result of a bit of an inconsistency in how rate limiting is configured in the UI vs terraform. The log statement is [added by default to `vcl_log` routine](https://gist.github.com/ewdurbin/bacc8b9be7cf9a736885549d94c73c21#file-141-154-diff-L53-L71).

This should disable logging by default, and only be called in the rate limit section.

### DIFF
--- a/infra/cdn/main.tf
+++ b/infra/cdn/main.tf
@@ -136,6 +136,12 @@ resource "fastly_service_vcl" "python_org" {
     statement = "req.http.host == \"python.org\""
     type      = "REQUEST"
   }
+  condition {
+    name      = "False"
+    priority  = 10
+    statement = "false"
+    type      = "RESPONSE"
+  }
 
   condition {
     name      = "Don't cache 404s for /static"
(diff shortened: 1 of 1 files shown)


infra/cdn/main.tf
```diff
@@ -136,6 +136,12 @@ resource "fastly_service_vcl" "python_org" {
     statement = "req.http.host == \"python.org\""
     type      = "REQUEST"
   }
+  condition {
+    name      = "False"
```
JacobCoffee: Could you ELI5 this name or maybe can we add more detail to it?

ewdurbin: The name of this condition is "False", and it is resolved from the statement `false` which will always be false. This gives us a way to assign this condition to something we never want to do (such as log to Datadog by default).

Perhaps renaming to "Always False" would be more clear?

ewdurbin: https://github.com/python/pythondotorg/pull/2653/commits/c38e543cd307028c2af1098c051999429fcadf51

JacobCoffee (approved):

PR #2656, opened 2024-11-15
Hide GPG column if no GPG signatures in release

hugovk:
#### Description

[PEP 761](https://peps.python.org/pep-0761/) ("Deprecating PGP signatures for CPython artifacts") has been accepted and means we will no longer provide PGP signatures for Python 3.14 and later. Verifiers are recommended to use Sigstore instead.

Let's hide the GPG column from the download page's files table when there are no such files in the release.

I considered hiding the column when the version `>= (3, 14)`, but went for this approach because it will also hide the column for very old releases which don't have GPG files, such as https://www.python.org/downloads/release/python-201/, which I think is also useful?

cc @sethmlarson

### DIFF
--- a/downloads/templatetags/download_tags.py
+++ b/downloads/templatetags/download_tags.py
@@ -9,6 +9,11 @@ def strip_minor_version(version):
 
 
 @register.filter
+def has_gpg(files: list) -> bool:
+    return any(f.gpg_signature_file for f in files)
+
+
+@register.filter
 def has_sigstore_materials(files):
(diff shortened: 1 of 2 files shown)


JacobCoffee (approved):

sethmlarson (approved):
LGTM, thanks for submitting this!

JacobCoffee:
https://www.python.org/downloads/release/python-201/ looks good to me now! :)

PR #2680, opened 2024-12-27
fix: return an empty profile page when not found

miketheman:
#### Description

Return an identical empty page to prevent user enumeration.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from django.contrib import messages
 from django.contrib.auth import get_user_model
 from django.contrib.auth.mixins import UserPassesTestMixin
+from django.contrib.auth.models import AnonymousUser
 from django.conf import settings
 from django.core.mail import send_mail
 from django.db.models import Subquery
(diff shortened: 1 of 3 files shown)


users/views.py
```diff
@@ -126,13 +127,20 @@ def get_object(self, queryset=None):
...
         queryset = User.objects.select_related()
         if self.request.user.username == self.kwargs['slug']:
             return queryset
         return queryset.searchable()
 
+    def get_object(self, queryset=None):
+        try:
+            return super().get_object(queryset)
+        except Http404:
+            return AnonymousUser()
```
JacobCoffee, line 142: Would returning an anonymous user make it known that a user doesn't exist and the malicious actor could continue enumerating over usernames?

miketheman: Initially, yes - since the HTML contents used to include the user's name in the `<title>` tag - that's been removed in the template.

I examined the HTML output of an existing vs anonymous user to confirm that the only difference is in the actual URL requested - nothing else.

ewdurbin:
I’m not positive we even _need_ or use the public profile page. Removing unauthenticated access to this view entirely is likely the correct move.

users/views.py
pradyunsg: > I examined the HTML output of an existing vs anonymous user to confirm that the only difference is in the actual URL requested - nothing else.

non-blocking suggestion: If it's not too expensive in terms of effort, it would be good to add a test that does this.

ewdurbin:
On a little closer review, I think we should probably do away with the slugged URLs:

https://github.com/python/pythondotorg/blob/023121f3efd852679d31814c55c1d56a8a426033/users/urls.py#L41-L42

and replace them with `/users/profile/detail` and `/users/profile/delete`, then add a test_func matching the UserDeleteView  to the UserDetail view that allows people to view their own details only.

PR #2717, opened 2025-04-24
Update website for Python install manager

zooba:
This is not ready to merge yet! But it's the area I'm most uncomfortable with updating, so I'm getting the PR out early.

@ambv suggested that we should treat PyManager releases as another Python release, so I think that's what I've enabled. This will put them in all the same release lists as Python (as long as downloads only exist for Windows, I _believe_ it won't appear on other OS lists), and it makes updating the "Download latest" buttons a bit easier.

I'm leaving the old installer available in those popups for as long as it exists, so that we don't seem quite as pushy. The installer will have its own message warning about deprecation. Eventually I'd like to replace these messages with a "Run `py install {version}`" message, but haven't quite figured that part out - it's not urgent.

Very open to any kind of help/feedback/assistance here, this is somewhat out of my depth, and while I _could_ spend time learning all the ins and outs, I'd rather just ask for help :)


zooba:
The CI errors aren't mine:

```
psycopg2.errors.ForeignKeyViolation: insert or update on table "sponsors_benefitfeatureconfiguration" violates foreign key constraint "sponsors_benefitfeat_polymorphic_ctype_id_0ce3d544_fk_django_co"
DETAIL:  Key (polymorphic_ctype_id)=(100) is not present in table "django_content_type".
```

I suspect in any case this PR probably needs to be split up to change the DB first so that we can have at least one release in it before updating the templates?

zooba:
Okay, with those template tweaks the visuals should be entirely unaffected if there are no PyManager releases registered, which _should_ mean it can all be deployed at once and we get no changes until we put up a release.

downloads/views.py
```diff
@@ -45,13 +45,30 @@ def get_redirect_url(self, **kwargs):
...
+    def get_redirect_url(self, **kwargs):
+        try:
+            latest_pymanager = Release.objects.latest_pymanager()
+        except Release.DoesNotExist:
+            latest_pymanager = None
+
+        if latest_pymanager:
+            return latest_pymanager.get_absolute_url()
+        else:
+            return reverse('download')
```
JacobCoffee:
```suggestion
            return reverse('downloads')
```

JacobCoffee (approved):

JacobCoffee:
Hi @zooba, all looks good after review—are you ready to merge this?

Thanks for the PR :)

zooba:
> are you ready to merge this?

I think so, but I am exceedingly nervous about it! If it looks good to you, and it's easy enough to roll back if needed, then go ahead.

PR #2725, opened 2025-05-06
feat: sanitize release names upon submission

### LINKED ISSUE #2724: Enhancement: Sanitize release names to require string
### Is your feature request related to a problem? Please describe.

When creating a new release to test #2717, I thought the PR was bad because it was possible to create a release named just `3.20`, causing the `downloads/windows/` page to crash because it lacked a `None` check in the PR for release version.

### Describe the solution you'd like

It'd be nice to just give a little error message in the admin form submission via:

```py
def clean(self):
  super().clean()
  if not re.match(r'^Python\s[\d.]+$', self.name):
      raise ValidationError({
          'name': 'Name must be in the format "Python X.Y.Z" (e.g., "Python 3.9.0")'
      })
```
to get:
Image

### Basic Example

_No response_

### Describe alternatives you've considered

_No response_

### Drawbacks and Impact

_No response_

### Additional context

_No response_

JacobCoffee:
#### Description

- Clean release name field to ensure no errors when loading release pages

Closes #2724

### DIFF
--- a/downloads/models.py
+++ b/downloads/models.py
@@ -17,7 +17,7 @@ from fastly.utils import purge_url
 from pages.models import Page
 
 from .managers import ReleaseManager
-
+from .validators import is_valid_python_release
 
 DEFAULT_MARKUP_TYPE = getattr(settings, 'DEFAULT_MARKUP_TYPE', 'restructuredtext')
 
@@ -50,6 +50,7 @@ class Release(ContentManageable, NameSlugModel):
         (PYTHON2, 'Python 2.x.x'),
         (PYTHON1, 'Python 1.x.x'),
     )
+    name = models.CharField(max_length=200, validators=[is_valid_python_release])
     version = models.IntegerField(default=PYTHON3, choices=PYTHON_VERSION_CHOICES)
     is_latest = models.BooleanField(
         verbose_name='Is this the latest release?',
@@ -145,13 +146,6 @@ class Release(ContentManageable, NameSlugModel):
     @property
     def is_version_at_least_3_9(self):
         return self.is_version_at_least((3, 9))
-
-    def clean(self):
-        super().clean()
-        if not re.match(r'^Python\s[\d.]+$', self.name):
-            raise ValidationError({
-                "name": "Release name must be in the format 'Python X.Y.Z' (e.g., 'Python 3.14.0')"
-            })
 
 
 def update_supernav():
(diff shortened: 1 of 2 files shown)


downloads/models.py
```diff
@@ -146,6 +146,20 @@ def is_version_at_least_3_5(self):
...
         return self.is_version_at_least((3, 9))
 
+    def clean(self):
+        super().clean()
+        if not re.match(r'^Python\s[\d.]+$', self.name):
+            raise ValidationError({
+                "name": "Release name must be in the format 'Python X.Y.Z' (e.g., 'Python 3.14.0')"
+            })
+
+    # def clean(self):
```
ewdurbin: is this necessary?

JacobCoffee: ohh i copied from the issue i created 🤦🏼‍♂️

ewdurbin:
@JacobCoffee is there an issue associated with this PR? I mean it makes sense but hard to say if it is "correct" without some documenting reason.

JacobCoffee:
yeah sorry, i thought if i put it in the commit message it would link it

ewdurbin (changes requested):
I would recommend using a [Validator](https://docs.djangoproject.com/en/5.2/ref/validators/) rather than overriding clean.

downloads/validators.py
```diff
@@ -0,0 +1,9 @@
+"""Model valdators for the DOwnloads app."""
```
JacobCoffee:
```suggestion
"""Model validators for the Downloads app."""
```

JacobCoffee:
doh...

before i do all the migrations and things, do you think overriding the name on the NameSlugModel and adding the validator is still the best path @ewdurbin? I'm okay with either just wanting to make sure

hugovk:
This should also allows pre-release names such as:

* Python 3.15.0a1, Python 3.15.0b2, Python 3.15.0rc3

And we now also have Python install manager releases with names like:

* Python install manager 25.0
* Python install manager 25.1 beta 1

https://www.python.org/admin/downloads/release/

---

Also this PR has flipped the order of Slug and Name:

image

Let's keep Name first like in production:

image

JacobCoffee:
https://github.com/python/pythondotorg/pull/2819 did this well enough i think?

hugovk:
Yep, that one's working.

PR #2730, opened 2025-05-12
fix(event): change events upcoming order

shenxianpeng:
#### Description

-

#### Closes

-


shenxianpeng:
I want to see the preview page of https://www.python.org/events/, but it doesn't seem to have one. Sorry for the noise.

EDIT: use `make serve` command from local can not (see) list of events.

PR #2736, opened 2025-05-26
Hide releases from Downloads page when they have no files to download

zooba:
The problem is most obvious on [the macOS page](https://www.python.org/downloads/macos/), which has a lot of releases listed with no files. For a _downloads_ page, it looks a bit silly.

Most critically, this should hide the Python install manager releases from macOS and Source pages, as it will have no files. An alternative would be to hard-code its ID into the templates, but this way seemed better.

There might be a better way to check if there are no files, I'm not sure. And yes, this makes the `{% empty %}` blocks redundant, but they're also harmless, and if this check is ever changed at least the page will still work.

### DIFF
--- a/templates/downloads/os_list.html
+++ b/templates/downloads/os_list.html
@@ -40,6 +40,7 @@
                 <h2>Stable Releases</h2>
                 <ul>
                     {% for r in releases %}
+                    {% if r.files.all %}
                     <li>
                         <a href="{{ r.get_absolute_url }}">{{ r.name }} - {{ r.release_date|date }}</a>
                         {% if os.slug == 'windows' %}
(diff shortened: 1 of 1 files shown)


JacobCoffee (approved):
<details>
<summary>Details</summary>

before
image

after
image

</details>


lgtm

zooba:
I hope those before/after images are swapped, because it looks like it's _added_ empty entries.

Also, should this have deployed yet? I don't see any change on the live site, but maybe I shouldn't expect to - I really don't know.

ewdurbin:
> Also, should this have deployed yet? I don't see any change on the live site, but maybe I shouldn't expect to - I really don't know.

It has deployed, but everything under /downloads has a massive cache TTL: https://github.com/python/pythondotorg/blob/142583bd66d5efe4c86b01ffb299b1979c85b91d/config/nginx.conf#L349

You can purge a given URL with `curl -X PURGE https://www.python.org/downloads/macos/` to force the page to render sooner.

PR #2759, opened 2025-07-29
fix: order events properly

### LINKED ISSUE #2723: Bug: Events page shows events in wrong order
### Describe the bug

https://www.python.org/events/ shows entries in the wrong order.

For a better UX, the page should show upcoming events starting from the today onwards first, not from the end of the year going backwards, as is done at the moment.

Note: The same problem may also be present on the two more specific calendar pages: https://www.python.org/events/python-events/ and https://www.python.org/events/python-user-group/ but those don't show any events at the moment, so it's not possible to check (see https://github.com/python/pythondotorg/issues/2632)

### To Reproduce

https://www.python.org/events/ shows entries in the wrong order.

### Expected behavior

For a better UX, the page should show upcoming events starting from the today onwards first, not from the end of the year going backwards, as is done at the moment.

### URL to the issue

https://www.python.org/events/

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Browsers

Chrome

### Operating System

Windows

### Browser Version

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

JacobCoffee:
#### Description

- Fix sort order for events

<details>
<summary>Details</summary>

before
image

after
image

</details>

#### Closes

- Fixes #2723

### DIFF
--- a/events/views.py
+++ b/events/views.py
@@ -44,16 +44,26 @@ class EventHomepage(ListView):
 
     def get_queryset(self) -> Event:
         """Queryset to return all events, ordered by START date."""
-        return Event.objects.all().order_by("-occurring_rule__dt_start")
+        return Event.objects.all().order_by("occurring_rule__dt_start")
 
     def get_context_data(self, **kwargs: dict) -> dict:
         """Add more ctx, specifically events that are happening now, just missed, and upcoming."""
         context = super().get_context_data(**kwargs)
-        context["events_just_missed"] = Event.objects.until_datetime(timezone.now())[:2]
-        context["upcoming_events"] = Event.objects.for_datetime(timezone.now())
+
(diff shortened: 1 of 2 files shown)


events/views.py
```diff
@@ -44,16 +44,26 @@ class EventHomepage(ListView):
 
     def get_queryset(self) -> Event:
         """Queryset to return all events, ordered by START date."""
-        return Event.objects.all().order_by("-occurring_rule__dt_start")
+        return Event.objects.all().order_by("occurring_rule__dt_start")
```
JacobCoffee, line 47: descending//ascending flip for template

ewdurbin (approved):

PR #2776, opened 2025-09-18
Change "Non-English Docs" link to the translation dashboard

### LINKED ISSUE #2775: Enhancement: "Non-English Docs" should link to an up-to-date place
docs.python.org currently only provides language selection via the switcher, though there has been talk of creating a selection page (see https://github.com/python/docs-community/issues/14).

Instead of linking to the wiki, which does not list many of the available official translations now available, I suggest linking to the selection page, and in the mean-time, the [translation dashboard](https://python-docs-translations.github.io/dashboard/) which has links to docs.python.org/LANG/ and also contribution links.

StanFromIreland:
Closes #2775

### DIFF
--- a/fixtures/sitetree_menus.json
+++ b/fixtures/sitetree_menus.json
@@ -661,7 +661,7 @@
     "fields": {
         "title": "Non-English Docs",
         "hint": "",
-        "url": "http://wiki.python.org/moin/Languages",
+        "url": "https://python-docs-translations.github.io/dashboard/",
         "urlaspattern": false,
         "tree": 1,
         "hidden": false,


JacobCoffee (approved):

StanFromIreland:
@JacobCoffee I tested this today and it still points to the Wiki, is there something else we need to do?

ewdurbin:
Fixtures are a red herring and don't change the actual site tree or CMS content.

@JacobCoffee these are just fixtures for local dev, they don't impact the deployed site. You can make the equivalent change [here](https://www.python.org/admin/sitetree/tree/1/change/).

ewdurbin:
I went ahead and made the update. It will take a bit for caches to expire.

StanFromIreland:
Thanks!

PR #2790, opened 2025-10-08
fix: handle release objects with 'invalid' names

JacobCoffee:
#### Description

- Fixes issue with `/downloads/windows` when trying to parse a non-standard relase object version

### DIFF
--- a/downloads/models.py
+++ b/downloads/models.py
@@ -128,8 +128,11 @@ class Release(ContentManageable, NameSlugModel):
         return None
 
     def is_version_at_least(self, min_version_tuple):
+        version = self.get_version()
+        if version is None:
+            return False
         v1 = []
-        for b in self.get_version().split('.'):
+        for b in version.split('.'):
             try:
                 v1.append(int(b))
             except ValueError:
--- a/downloads/models.py
+++ b/downloads/models.py
@@ -125,14 +125,11 @@ class Release(ContentManageable, NameSlugModel):
         version = re.match(r'Python\s([\d.]+)', self.name)
         if version is not None:
             return version.group(1)
-        return None
+        return ""
 
     def is_version_at_least(self, min_version_tuple):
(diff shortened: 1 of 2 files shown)


zooba:
The root cause that got us here was that I ticked "show on download page" for a PyManager release, which I shouldn't have done. https://github.com/python/pythondotorg/blob/main/downloads/managers.py#L16 would have filtered it out and we never would've hit this path.

zooba:
Not disagreeing with making the function not crash, to be clear. Maybe it's just as good to make `get_version` return an empty string in the "not a real version" case?

JacobCoffee:
whichever you prefer, or if it's just better to document/know that `show on download page` shouldn't be enabled for pymanager releases? you all are the release experts :D

zooba:
> if it's just better to document/know ...

It never is! 😆

Let's at least make sure that checking the box doesn't bring down the site, and that unchecking it (which I have already done) fixes up any weird display on the page. Then we can rely on documentation from there, I think.

JacobCoffee:
@zooba does [f809fd1](https://github.com/python/pythondotorg/pull/2790/commits/f809fd19933ff7c9da6cde8799fe1d77fd5c8951) work for you? or did i misunderstand

```
➜ docker compose run --rm web ./manage.py test downloads
[+] Creating 2/2
 ✔ Container pythondotorg-redis-1     Running              0.0s 
 ✔ Container pythondotorg-postgres-1  Running              0.0s 

System check identified 2 issues (0 silenced).
.................................................................
----------------------------------------------------------------------
Ran 65 tests in 10.299s

OK
Destroying test database for alias 'default'...
```

zooba:
Yeah, that's the other approach I was thinking of. Both are fine, but I think the latter is a bit cleaner/safer over time.

ewdurbin (approved):

PR #2793, opened 2025-10-08
Update downloads supernav page for Python install manager

zooba:
Includes tests (written blind, so let's see if they work...)

### DIFF
--- a/downloads/tests/test_models.py
+++ b/downloads/tests/test_models.py
@@ -1,4 +1,4 @@
-from ..models import Release
+from ..models import Release, ReleaseFile
 from .base import BaseDownloadTests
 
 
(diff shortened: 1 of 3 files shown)


zooba:
Uh yeah, the test passed first try (pay no attention to the force push).

This should fix up the download popup on the main page. Please only merge and deploy when you're watching the site, in case things break again :)

templates/downloads/supernav.html
```diff
@@ -10,8 +10,10 @@ <h4>Download for {{ data.os.name }}</h4>
 
         {% if data.pymanager %}
         <p><a class="button" href="{{ data.pymanager.url }}">Python install manager</a></p>
+        {% if data.python3 %}
         <p>Or get the standalone installer for <a class="button" href="{{ data.python3.url }}">{{ data.python3.release.name }}</a></p>
-        {% else %}
+        {% endif %}
+        {% elif data.python3 %}
```
zooba, line 16: FWIW, I don't expect we'll ever get into a situation where we have a latest PyManager but not a latest Python 3 (which is what the `elif` instead of `else` implies is possible). But just in case it happens, at least we won't break everything - there'll just be some missing text.

downloads/tests/test_models.py
```diff
@@ -95,3 +95,54 @@ def test_is_version_at_least_with_invalid_name(self):
         self.assertFalse(invalid_release.is_version_at_least_3_5)
         self.assertFalse(invalid_release.is_version_at_least_3_9)
         self.assertFalse(invalid_release.is_version_at_least_3_14)
+
+    def test_update_supernav(self):
+        from ..models import update_supernav
+        from boxes.models import Box
+
+        release = Release.objects.create(
```
zooba, line 103: I assume it's okay to create entries in a test and not clean them up? I haven't worked with Django tests much, but it seems to not be upsetting anything else...

JacobCoffee (approved):
before
image
after
image

PR #2805, opened 2025-11-07
Add featured downloads

### LINKED ISSUE #2419: Enhancement: Add a `featured downloads` section
Here's some more ideas. Some programming languages have a banner with featured downloads, and extra guidance text recommending 64-bit over 32-bit.

<details>
<summary>Node.js</summary>

# Node.js

For example, [Node.js](https://nodejs.org/en/download/) has the 64-bit Windows installer, a macOS installer and the source tarball above the full table: 

image

</details>

<details>
<summary>Go</summary>

# Go

Similarly, [Go](https://go.dev/dl/) shows the 64-bit Windows installer next to macOS installers and Linux and source tarballs. These files are also in bold in the full table:

image

</details>

<details>
<summary>Ruby</summary>

# Ruby

The [Ruby installer for Windows](https://rubyinstaller.org/downloads/) puts the 64-bit installer in bold next to an arrow, along with "Not sure what version to download? Please read the right-hand column for recommendations" which recommends 64-bit, and doesn't recommend the 32-bit:

image

</details>

<details>
<summary>PHP</summary>

# PHP

The [PHP for Windows](https://windows.php.net/download) page puts the 64-bit downloads first, and in bold:

image

Further down in the left sidebar a recommendation for x64 builds:

image

</details>

_Originally posted by @hugovk in https://github.com/python/pythondotorg/issues/2194#issuecomment-1874289095_

hugovk:
#### Description

- On download pages such as https://www.python.org/downloads/release/python-3140/, it's not immediately clear what to click in the big table of a dozen files.

- Let's add the most common ones as feature downloads at the top. 

## Before

image

## After

image

## Notes

Each "Release" object in the CMS has a number of release files. Some of these have a "Download button" boolean set, max one per OS, which is used for displaying it at https://www.python.org/downloads/:

image

Let's also use this flag to decide what to feature here.

(Aside: the flag is set during the release process by `download=True` in https://github.com/python/release-tools/blob/d8077eae1dcea2653775e4b3ba9f51700b354f5e/add_to_pydotorg.py#L105-L190.)

## Windows

However, for Windows, let's use the Python install manager and link to https://www.python.org/downloads/latest/pymanager/ using similar logic for the button at https://www.python.org/downloads/

That page says:

> The install manager can install versions of Python as far back as 3.5, but only supports Windows 10 operating systems (or Windows Server 2022) and later.

Re: "The install manager can install versions of Python as far back as 3.5"

So let's only show the install manager for downloads of 3.5 and later. Older ones can use the file set by the CMS. For example, for 3.4.0:

image

Re: "but only supports Windows 10 operating systems (or Windows Server 2022) and later"

For the purpose of this featured release for 3.5+, it's fine to cater for the common case of new Windows (especially as Win10 is EOL). The full list is right there if you need a file.

@zooba How does this sound?

### DIFF
--- a/downloads/views.py
+++ b/downloads/views.py
@@ -2,7 +2,7 @@ from typing import Any
 
 from datetime import datetime
 
-from django.db.models import Prefetch
+from django.db.models import Case, IntegerField, Prefetch, When
 from django.urls import reverse
 from django.utils import timezone
 from django.views.generic import DetailView, TemplateView, ListView, RedirectView
@@ -156,6 +156,20 @@ class DownloadReleaseDetail(DownloadBase, DetailView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+
+        # Add featured files (files with download_button=True)
+        # Order: macOS first, Windows second, Source last
+        context['featured_files'] = self.object.files.filter(
+            download_button=True
+        ).annotate(
+            os_order=Case(
+                When(os__slug='macos', then=1),
+                When(os__slug='windows', then=2),
+                When(os__slug='source', then=3),
+                default=4,
+                output_field=IntegerField(),
+            )
+        ).order_by('os_order')
 
         # Manually add release files for better ordering
         context['release_files'] = []
--- a/static/sass/style.css
+++ b/static/sass/style.css
@@ -2113,6 +2113,50 @@ table tfoot {
 .download-widget p:last-child a {
   white-space: nowrap; }
 
+.featured-downloads-list {
+  display: flex;
+  flex-wrap: wrap;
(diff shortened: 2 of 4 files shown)


zooba:
I like it, though worth bearing in mind that 3.15 is the last release with the long list of downloads for Windows. From 3.16, it'll go down to just the JSON file ("release manifest").

hugovk:
Thanks, it'll also be useful for 3.16, the https://www.python.org/downloads/latest/pymanager/ will have more useful info than a [JSON file](https://www.python.org/ftp/python/3.14.0/windows-3.14.0.json) :)

---

Ready for review!

JacobCoffee (approved):
love it

Issue #2809, opened 2025-11-15
Bug: Update "Non-English Docs" link


StanFromIreland:
### Describe the bug

It has moved to https://translations.python.org

### To Reproduce

Click "Non-English Docs"

### Expected behavior

You are sent to https://translations.python.org directly avoiding the redirect

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Browsers

Firefox, Safari

### Operating System

Linux

### Relevant log output

```shell

```

hugovk:
Updated in the CMS, and also the FAQ link like in #2810.

These are the CMS links:

https://www.python.org/admin/sitetree/tree/1/change/item_29/
https://www.python.org/admin/sitetree/tree/1/change/item_30/

StanFromIreland:
Thanks Hugo!

PR #2810, opened 2025-11-15
Update URLs in sitetree_menus.json

StanFromIreland:
Part of #2809 

This also has to be implemented in the CMS.

### DIFF
--- a/fixtures/sitetree_menus.json
+++ b/fixtures/sitetree_menus.json
@@ -637,7 +637,7 @@
     "fields": {
         "title": "FAQ",
         "hint": "",
-        "url": "https://docs.python.org/faq/",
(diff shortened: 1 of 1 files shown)


fixtures/sitetree_menus.json
```diff
@@ -637,7 +637,7 @@
     "fields": {
         "title": "FAQ",
         "hint": "",
-        "url": "https://docs.python.org/faq/",
+        "url": "https://docs.python.org/3/faq/",
```
StanFromIreland, line 640: I also noticed this which should be updated, it is a redirect currently.

JacobCoffee (approved):

StanFromIreland:
Thanks!

PR #2816, opened 2025-11-21
Redirect `downloads/latest/python3.x` for given 3.x

hugovk:
#### Description

Add new redirects for the latest of each 3.x feature release.

We already have (with current redirects):
* https://python.org/downloads/latest/python2 -> https://www.python.org/downloads/release/python-2718/
* https://python.org/downloads/latest/python3 -> https://www.python.org/downloads/release/python-3140/
* https://www.python.org/downloads/latest -> https://www.python.org/downloads/release/python-3140/
* https://www.python.org/downloads/latest/pymanager/ -> https://www.python.org/downloads/release/pymanager-250/

This adds some for 3.x. For example (based on local test CMS data):

* http://0.0.0.0:8000/downloads/latest/python3.10 -> http://0.0.0.0:8000/downloads/release/python-3107/
* http://0.0.0.0:8000/downloads/latest/python3.9 -> http://0.0.0.0:8000/downloads/release/python-3914/

Unknown versions go back to the downloads page:

* http://0.0.0.0:8000/downloads/latest/python3.26 -> http://0.0.0.0:8000/downloads/

(I didn't bother with `downloads/latest/python2.x`, people can use https://python.org/downloads/latest/python2 or  https://www.python.org/downloads/release/python-2718/ or upgrade.)

---

After merge, we can add an extra downloads column to the `downloads-active-releases` box in the CMS, something like:

<details>
<summary>Details</summary>

```html
<div class="list-row-headings">
    <span class="release-version">Python version</span>
    <span class="release-status">Maintenance status</span>
    <span class="release-dl">&nbsp;</span>
    <span class="release-start">First released</span>
    <span class="release-end">End of support</span>
    <span class="release-pep">Release schedule</span>
</div>
<ol class="list-row-container menu">
    <li>
        <span class="release-version">3.15</span>
        <span class="release-status"><a href="https://www.python.org/download/pre-releases/">pre-release</a></span>
        <span class="release-dl"><a href="/downloads/latest/python3.15/"><span aria-hidden="true" class="icon-download"></span>Download</a></span>
        <span class="release-start">2026-10-07 (planned)</span>
        <span class="release-end">2031-10</span>
        <span class="release-pep"><a href="https://peps.python.org/pep-0790/">PEP 790</a></span>
    </li>
    <li>
        <span class="release-version">3.14</span>
        <span class="release-status">bugfix</span>
[...]

PR #2821, opened 2025-11-24
Add tests for download redirects and refactor DownloadLatestPython3

hugovk:
#### Description

- Followup from https://github.com/python/pythondotorg/pull/2816
- Add tests for the new redirects, plus the old `/downloads/latest/python3`
- In the [original](https://github.com/python/pythondotorg/pull/2816#discussion_r2550555421) @JacobCoffee suggested combining `latest_python3x` into `latest_python3`, and we can do the same by combining `DownloadLatestPython3x` into `DownloadLatestPython3`


JacobCoffee (approved):
smooth

PR #2822, opened 2025-11-24
Sort downloads table by version

### LINKED ISSUE #2817: Sort downloads table by version number, not release date
- Sort the ["Looking for a specific release?"] table in reverse order by (major, manor, bugfix) release -- having the last 3.7 appear before the last 3.10 makes little sense. (Maybe split into multiple tables by (major, minor) to avoid having to scroll through 3.13.1-9 before you see 3.12?)

_Originally posted by @gvanrossum in https://github.com/python/pythondotorg/issues/2816#issuecomment-3563970863_
            
Image

hugovk:
#### Description

- Sort the downloads table, with new x.y.z versions at the top, instead of sorting by date.

From the issue:

> (Maybe split into multiple tables by (major, minor) to avoid having to scroll through 3.13.1-9 before you see 3.12?)

I considered this, but decided against it -- the first table now has download links for the latest version of each 3.x feature release (https://github.com/python/pythondotorg/pull/2816).

Before:

(Note this is with local test data)

image

After:

image

#### Closes

- Fixes https://github.com/python/pythondotorg/issues/2817

### DIFF
--- a/downloads/tests/test_views.py
+++ b/downloads/tests/test_views.py
@@ -56,6 +56,21 @@ class DownloadViewsTests(BaseDownloadTests):
         response = self.client.get(url)
         self.assertEqual(response.status_code, 200)
 
+    def test_download_releases_ordered_by_version(self):
+        url = reverse("download:download")
+        response = self.client.get(url)
+        releases = response.context["releases"]
+        self.assertEqual(
+            releases,
+            [
+                self.python_3,
+                self.python_3_10_18,
+                self.python_3_8_20,
+                self.python_3_8_19,
+                self.release_275,
+            ],
+        )
+
     def test_latest_redirects(self):
         latest_python2 = Release.objects.released().python2().latest()
         url = reverse('download:download_latest_python2')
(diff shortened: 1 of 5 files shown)


gvanrossum:
Curious why the new "Looking for a specific release?" table lists only 3.10 releases. Are there *no* bugfix releases of 3.11, 3.12 or 3.13?

hugovk:
It's a screenshot from a local server on my machine, and it comes with an old test database that only goes up to 3.10.

There are bugfix releases for 3.11-3.13 in the production database, and they'll show up after merge.

downloads/tests/test_views.py
```diff
@@ -594,5 +609,5 @@ def test_feed_item_count(self) -> None:
         response = self.client.get(self.url)
         content = response.content.decode()
 
-        # In BaseDownloadTests, we create 5 releases, 4 of which are published, 1 of those published are hidden..
-        self.assertEqual(content.count("<item>"), 4)
+        # In BaseDownloadTests, we create 8 releases, 7 of which are published, 1 of those published are hidden..
+        self.assertEqual(content.count("<item>"), 7)
```
merwok, line 613: So the hidden one is visible?

gvanrossum (approved):
I'm not sure it matters, and I don't know this code base, but LGTM anyway!

downloads/tests/test_views.py
hugovk: It's published, so you can see its own release page, but specifically hidden from the `/downloads` page. 

https://github.com/python/pythondotorg/blob/bf66e9d85fb8ff2df70a32380031b1714948838a/downloads/tests/base.py#L86-L92

JacobCoffee (approved):

PR #2823, opened 2025-11-25
Add `/downloads/latest/prerelease` redirect

hugovk:
#### Description

One task for the release manager (👋) for every pre-release (and the .0 final) is to manually edit https://www.python.org/download/pre-releases/ to point to the new one.

Instead, let's create `/downloads/latest/prerelease` to redirect to the latest pre-release available.

For example:

* `downloads/latest/prerelease` -> `/downloads/release/python-3150a2/`

In the ~one week period when 3.x.0 has just been released and 3.x+1.a1 is not out yet:

* `/downloads/latest/prerelease` -> `/downloads/`

This PR also updates the `download-banner` and `downloads-active-releases` boxes in the test data to match production, for easier testing, and then to use the new redirects in these places:

image

After merge, I'll update:

* [x] the boxes in prod to use the new redirect
* [x] the old static page https://www.python.org/download/pre-releases/ to use the new redirect
* [x] [PEP 101](https://peps.python.org/pep-0101/) to remove "Update the “Current pre-release testing versions” page": [peps#4717](https://github.com/python/peps/pull/4717)
* [x] [python/release-tools](https://github.com/python/release-tools/blob/f408a389581a14499fb461bb30ef1d8b202857aa/run_release.py#L1113) to remove `modify_the_prereleases_page()`: [release-tools#308](https://github.com/python/release-tools/pull/308)


downloads/managers.py
```diff
@@ -52,5 +55,8 @@ def latest_python2(self):
     def latest_python3(self, minor_version: int | None = None):
         return self.get_queryset().latest_python3(minor_version).first()
 
+    def latest_prerelease(self):
```
JacobCoffee, line 58: does this not inherit the base class's latest_prerelease?

JacobCoffee: ohh you add .first method, nvm

JacobCoffee (approved):

JacobCoffee:
needs rebase @hugovk :)

hugovk:
Updated!

PR #2825, opened 2025-11-26
Remove redundant print() in homepage code sample

ArminHaberl:
### Description
I noticed that the Fibonacci code example on the Python homepage includes a redundant empty `print()` statement. After testing the sample in Python 3.14, I confirmed that the output is identical with or without this line.

### Proposed Change
Remove the unnecessary `print()` statement from the code example.

### Rationale
- The empty `print()` does not affect program output.
- Simplifying the example improves clarity for beginners.

Thank you for reviewing this update!

### DIFF
--- a/codesamples/factories.py
+++ b/codesamples/factories.py
@@ -126,7 +126,6 @@ def initial_data():
             ...     while a &lt; n:
             ...         print(a, end=' ')
             ...         a, b = b, a+b
-            ...     print()
             ...
             >>> fib(1000)
             <span class=\"output\">0 1 1 2 3 5 8 13 21 34 55 89 144 233 377 610</span>


hugovk:
Thanks for the suggestion, but it's there to clean up after the `print`s with `end=' '`, which don't add a newline.

Without, we have the `>>>` prompt on the same line:

```pycon
>>> def fib(n):
...     a, b = 0, 1
...     while a < n:
...         print(a, end=' ')
...         a, b = b, a+b
...
>>> fib(1000)
0 1 1 2 3 5 8 13 21 34 55 89 144 233 377 610 987 >>>
```

But with, the `>>>` is on a new line, as usual:

```pycon
>>> def fib(n):
...     a, b = 0, 1
...     while a < n:
...         print(a, end=' ')
...         a, b = b, a+b
...     print()
...
>>> fib(1000)
0 1 1 2 3 5 8 13 21 34 55 89 144 233 377 610 987
>>>
```

ArminHaberl:
Thanks for pointing that out!
I see why the print statement was included to keep the REPL output tidy.
I'll go ahead and close the PR.

PR #2827, opened 2025-11-26
Add 'superseded by' notice to older Python 3 releases

hugovk:
#### Description

Automating some more manual RM stuff from [PEP 101](https://peps.python.org/pep-0101/):

> * If this is a final release:
>   * For 3.X.Y, edit all the previous X.Y releases’ page(s) to point to the new release. This includes the content field of the Downloads -> Releases entry for the release:
>    ```
>    Note: Python 3.x.(y-1) has been superseded by
>    `Python 3.x.y </downloads/release/python-3xy/>`_.
>    ```

This tedious manual work isn't often followed, and is perfect for automating over here.

So this adds a note to release pages to link to the latest one in the feature series. The latest one doesn't have the link.

image
image

For example:

* http://0.0.0.0:8000/downloads/release/python-3150a2/ no note
* http://0.0.0.0:8000/downloads/release/python-3150a1/ notes superseded by 3.15.0a2
* http://0.0.0.0:8000/downloads/release/python-3140/ no note
* http://0.0.0.0:8000/downloads/release/python-3140rc3/ notes superseded by 3.14.0
* http://0.0.0.0:8000/downloads/release/python-3139/ no note
* http://0.0.0.0:8000/downloads/release/python-3138/ notes superseded by 3.13.9
* http://0.0.0.0:8000/downloads/release/python-3137/ notes superseded by 3.13.9
* http://0.0.0.0:8000/downloads/release/python-3130/ notes superseded by 3.13.9
* and so on

I didn't bother adding it for Python 2 releases.

PR #2832, opened 2025-11-28
Add EOL banner to EOL downloads

hugovk:
#### Description

- Add an EOL banner to the relevant releases, similar to https://docs.python.org/3.9/.
- Thanks to @ned-deily for the idea.

We look up the release status and EOL date from https://peps.python.org/api/python-releases.json and cache it.

That API only has data back to 1.6, and we only have releases back to 2.0, so we're covered.

I used the existing red `.user-feedback.level-error` CSS, and adjust the "Warning:" and link colours to meet the WCAG AA guidelines for contrast ratio.

I also switched the superseded-by note added in https://github.com/python/pythondotorg/pull/2827 to use the yellow `.user-feedback.level-notice` and fixed the contrast for "Note:".

Both banners:

image

Only the EOL banner:

image

Only the superseded by banner:

image

And this is how the contrast looked before (3.30, 3.45 and 2.02 respectively, now 5.53, 5.10, and 6.06):

image

### DIFF
--- a/downloads/templatetags/download_tags.py
+++ b/downloads/templatetags/download_tags.py
@@ -1,6 +1,76 @@
+import logging
+import re
+
+import requests
 from django import template
+from django.core.cache import cache
 
 register = template.Library()
+logger = logging.getLogger(__name__)
+
+PYTHON_RELEASES_URL = "https://peps.python.org/api/python-releases.json"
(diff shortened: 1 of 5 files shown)


JacobCoffee (approved):

JacobCoffee:
image
making its way through the cache, https://www.python.org/downloads/release/python-390/

PR #2837, opened 2025-11-29
Show release cycle charts on downloads page

hugovk:
#### Description

- Currently, we only show this in the CPython developer's guide: https://devguide.python.org/versions/
- But best practice is to show it where end users visit, see https://endoflife.date/recommendations#bad---python
- This PR depends first on merging [devguide#1708](https://github.com/python/devguide/pull/1708)

image

### DIFF
--- a/templates/downloads/index.html
+++ b/templates/downloads/index.html
@@ -53,6 +53,8 @@
                     <h2 class="widget-title">Active Python releases</h2>
                     <p class="success-quote"><a href="https://devguide.python.org/versions/#versions">For more information visit the Python Developer's Guide</a>.</p>
 
+                    <p class="release-cycle-chart"><img src="https://hugovk-devguide.readthedocs.io/en/standalone-svg/release-cycle.svg" alt="Python release cycle"></p>
+
                     {% box 'downloads-active-releases' %}
                 </div>
 
@@ -89,9 +91,9 @@
                     </div>
                 </div>
 
-        		<div class="row">
+                <div class="row">
 
-        		    <div class="small-widget download-widget1">
+                    <div class="small-widget download-widget1">
                         {% box 'download-widget1' %}
 
                     </div>
@@ -101,7 +103,7 @@
 
                     </div>
 
-        		    <div class="small-widget download-widget3">
+                    <div class="small-widget download-widget3">
                         {% box 'download-widget3' %}
 
                     </div>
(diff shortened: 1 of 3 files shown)


templates/downloads/index.html
```diff
@@ -53,6 +53,8 @@ <h1 class="call-to-action">Download the latest version of Python</h1>
                     <h2 class="widget-title">Active Python releases</h2>
                     <p class="success-quote"><a href="https://devguide.python.org/versions/#versions">For more information visit the Python Developer's Guide</a>.</p>
 
+                    <p class="release-cycle-chart"><img src="https://hugovk-devguide.readthedocs.io/en/standalone-svg/release-cycle.svg" alt="Python release cycle"></p>
```
hugovk: For demo purposes.

When [devguide#1708](https://github.com/python/devguide/pull/1708) is merged, replace with:

```suggestion
                    <p class="release-cycle-chart"><img src="https://devguide.python.org/_static/release-cycle.svg" alt="Python release cycle"></p>
```

templates/downloads/index.html
```diff
@@ -53,6 +53,8 @@ <h1 class="call-to-action">Download the latest version of Python</h1>
                     <h2 class="widget-title">Active Python releases</h2>
                     <p class="success-quote"><a href="https://devguide.python.org/versions/#versions">For more information visit the Python Developer's Guide</a>.</p>
 
+                    <p class="release-cycle-chart"><img src="https://hugovk-devguide.readthedocs.io/en/standalone-svg/_static/release-cycle.svg" alt="Python release cycle"></p>
```
hugovk:
```suggestion
                    <p class="release-cycle-chart"><img src="https://devguide.python.org/_static/release-cycle.svg" alt="Python release cycle"></p>
```

hugovk:
Failure is unrelated, looks like `test_events_scheduled_current_year_does_not_display_current_year` assumes it's never run in the last three days of the year!

Maybe I'll fix it tomorrow, maybe I'll just wait until the 1st :)

hugovk:
> maybe I'll just wait until the 1st :)

A successful strategy, now passes!

JacobCoffee (approved):

templates/downloads/index.html
```diff
@@ -53,6 +53,8 @@ <h1 class="call-to-action">Download the latest version of Python</h1>
                     <h2 class="widget-title">Active Python releases</h2>
                     <p class="success-quote"><a href="https://devguide.python.org/versions/#versions">For more information visit the Python Developer's Guide</a>.</p>
 
+                    <p class="release-cycle-chart"><img src="https://devguide.python.org/_static/release-cycle.svg" alt="Python release cycle"></p>
```
JacobCoffee, line 56: would this url or static change for any reason?

hugovk: No, only if we intentionally decided to change it, but we specifically chose this URL in preparation for this PR:

https://github.com/python/devguide/pull/1708#pullrequestreview-3520747445

PR #2841, opened 2025-12-06
Improve contrast ratio of H2

hugovk:
#### Description

- Improve accessibility by increasing the contrast ratio to meet WCAG AA guidelines of at least 4.5. Let's use the same colour as the H3.

**Before**

ratio: 2.78

image

**After**

ratio 15.37

image


JacobCoffee (approved):
so much better

PR #2842, opened 2025-12-06
Fix updating the supernav downloads

### LINKED ISSUE #2839: Bug: templates/downloads/supernav.html not updating
### Describe the bug

For some reason, the supernav box isn't updating. I'm guessing it's related to PyManager changes, but I don't know how to see what the issues are. Hopefully there's a log file somewhere?

### To Reproduce

Image

According to the comment in the main site:
```
<!-- Download supernav from templates/downloads/supernav.html - Last updated at Oct. 9, 2025, 8:41 p.m. -->
```

I've purged the URL with no effect, so I'm pretty sure it's on the backend.

### Expected behavior

It should have updated to show the PyManager 25.0 or 25.1 download in the Windows div, but hasn't.

### URL to the issue

_No response_

### Screenshots

```bash

```

### Browsers

Other

### Operating System

Windows

### Browser Version

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

hugovk:
#### Description

The Downloads supernav menu shows Python 3.14.0 instead of 3.14.2.

The supernav box was last updated on 9th October 2025 (when 3.14.0 was released) and hasn't updated since.

This was caused by https://github.com/python/pythondotorg/pull/2793 (merged 10th October) which changed `update_supernav()` to support the Python install manager, but introduced a regression.

The code iterates over each operating system, looking for download files with `download_button=True` (let's call them "featured files") to put in the supernav.

Before, operating systems without featured files were skipped:

```python
release_file = latest_python3.download_file_for_os(o.slug)
if not release_file:
    continue  # Skip this OS
data['python3'] = release_file
```

After, all operating systems are added to the list, even without featured files:

```python
data['python3'] = latest_python3.download_file_for_os(o.slug)
if latest_pymanager:
    data['pymanager'] = latest_pymanager.download_file_for_os(o.slug)

python_files.append(data)  # Always appends, even if no featured files
```

But then a later new check requires every OS to have a featured file:

```python
if not all(f['python3'] or f['pymanager'] for f in python_files):
    return  # Silently return without updating
```

However, Android has no featured files, so for Python 3.14.1 and 3.14.2, the function silently returns without updating the supernav box.

The fix is to restore the skip: only add an OS to `python_files` if it has a featured file. 

#### Closes

- Fixes https://github.com/python/pythondotorg/issues/2839

### DIFF
--- a/downloads/models.py
+++ b/downloads/models.py
@@ -175,14 +175,11 @@ def update_supernav():
         if latest_pymanager:
             data['pymanager'] = latest_pymanager.download_file_for_os(o.slug)
 
-        python_files.append(data)
+        # Only include OSes that have at least one download file
+        if data['python3'] or data['pymanager']:
+            python_files.append(data)
 
     if not python_files:
-        return
-
-    if not all(f['python3'] or f['pymanager'] for f in python_files):
-        # We have a latest Python release, different OSes, but don't have release
-        # files for the release, so return early.
         return
 
     content = render_to_string('downloads/supernav.html', {
--- a/downloads/tests/test_models.py
+++ b/downloads/tests/test_models.py
@@ -157,7 +157,7 @@ class DownloadModelTests(BaseDownloadTests):
                 release=self.python_3,
                 slug=slug,
                 name='Python 3.10',
-                url='/ftp/python/{}.zip'.format(slug),
+                url=f'/ftp/python/{slug}.zip',
                 download_button=True,
             )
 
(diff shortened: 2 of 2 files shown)


hugovk:
This is the priority of my recent PRs.

Now we have had https://github.com/python/pythondotorg/pull/2845 for two days, we can see there are about 100x more clicks for 3.14.0 (presumably via the Downloads menu/supernav on every page) than 3.14.2 (presumably via the button at https://www.python.org/downloads/):

image

JacobCoffee (approved):

PR #2849, opened 2025-12-12
Use endpoint for release edit button that's only rendered for staff

hugovk:
#### Description

- Follow on from [python/pythondotorg#2843](https://github.com/python/pythondotorg/pull/2843)
- I can see the edit button locally on http://0.0.0.0:8000/downloads/release/python-3142/ but not at https://www.python.org/downloads/release/python-3142/
- I can see the edit button on pages such as https://www.python.org/psf/developersinresidence/ have an edit button
- I think this is because Fastly is caching `/downloads` for all users in the same way, because the Django `Vary` header is being hidden for `/downloads` but not other pages
- This caching was added in [python/pythondotorg#1390](https://github.com/python/pythondotorg/pull/1390) along with other measures, and the [majority](https://analytics.python.org/python.org/pages) of visits are to `/downloads`
- Perhaps it's safe to expose `Vary` and keep the other measures?

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -347,6 +347,7 @@ http {
                         proxy_redirect off;
                         add_header Cache-Control "max-age=604800, public";
                         add_header Surrogate-Control "max-age=604800, stale-if-error=3600, stale-while-revalidate=300";
+                        proxy_hide_header Vary;
                         proxy_pass http://app_server/downloads/;
                 }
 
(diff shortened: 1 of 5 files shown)


JacobCoffee (changes requested):
i think this might be better done in fastly?

JacobCoffee:
if i understand correctly, what about https://github.com/python/pythondotorg/pull/2850

hugovk:
Yeah, Fastly sounds better, but I'm not too familiar with its config to review it.

JacobCoffee:
Ee pointed out a better way, they said /authenticated is hit each time.
https://github.com/python/pythondotorg/blob/6afea27471d2de3fcd82ef83952577bd43212c2d/pydotorg/views.py#L44-L45
so maybe make somehting similar to this or just inherit it even

hugovk:
Yep, that sounds better, how's this?

JacobCoffee (approved):

PR #2856, opened 2025-12-23
Remove support for Internet Explorer

hugovk:
#### Description

I don't think we need this any more.

IE11 is mostly unsupported: https://endoflife.date/internet-explorer

Looking at stats for the last 28 days: https://analytics.python.org/python.org?period=28d&keybindHint=F

* 58 out of 11,749,647 visitors used IE:

image

image

image

image

I've done this in stages, the first commit removing IE6, then the next IE7, then IE8, IE9, then all IE, just in case we want to keep any of it.


JacobCoffee (approved):

JacobCoffee:
conflicts @hugovk

hugovk:
Resolved!

PR #2859, opened 2026-01-03
Remove fundraiser banner as year 2025 is over

corneliusroemer:
This reverts commit 27aef89cab44c2126d1a934b0da86bfd1f42211d.

The year 2025 is over, the banner has been up for more than 3 weeks (in production).

For discussion see:
- https://discuss.python.org/t/accessibility-issues-in-the-pypi-donation-banner/105383
- https://discuss.python.org/t/feedback-on-year-end-donation-banner-when-will-it-be-removed/105487

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -34,9 +34,6 @@
             crossorigin="anonymous"
     ></script>
     <script src="{{ STATIC_URL }}js/libs/modernizr.js"></script>
-    <script async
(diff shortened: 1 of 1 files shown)


ewdurbin (approved):
Thanks! The banner has been soft disabled so this isn't urgent. Ready to merge when fundraising team confirms we're done with banner.

PR #2866, opened 2026-01-10
Use native Django admin interface to support dark mode

hugovk:
#### Description

- I'd like the dark admin theme here, for those late night releases :)

I noticed there's a light/dark/auto theme button in the top right corner:

image

But no matter how many times I click it, it stays on light mode:

image

image

Django does natively support dark mode, but we actually have a custom theme that doesn't support it: https://github.com/fabiocaccamo/django-admin-interface/issues/129

This PR reverts back to the native theme.

This was originally changed in https://github.com/python/pythondotorg/pull/2263 because:

> psf staff spend a lot of time between us.pycon.org and python.org django admin this enables configuration of a different theme for python.org so its easier to differentiate. thanks to `@swiencks` for pointing out how easy to confuse they are.

I don't know what us.pycon.org looks like behind the scenes, but perhaps we could apply this other custom theme over there? Or stick some big logos at the top?

Other options involve custom CSS for the custom theme here, but that's a fair chunk of work and extra code to maintain, when it's already been done in Django itself.

## Before

image

## After

image

image

### DIFF
--- a/base-requirements.txt
+++ b/base-requirements.txt
@@ -1,9 +1,6 @@
 dj-database-url==0.5.0
 django-pipeline==3.1.0  # 3.0.0 is first version that supports Django 4.2
 django-sitetree==1.18.0  # >=1.17.1 is (?) first version that supports Django 4.2
-django-apptemplates==1.5
-django-admin-interface==0.28.9
-django-translation-aliases==0.1.0
 Django==4.2.27
 docutils==0.21.2
 Markdown==3.7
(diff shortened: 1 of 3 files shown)


JacobCoffee:
>I don't know what us.pycon.org looks like behind the scenes, but perhaps we could apply this other custom theme over there? Or stick some big logos at the top?

us.pycon.org admin is like this

image

image

JacobCoffee (approved):

Issue #1227, opened 2018-03-01
Python download checksum in MD5


NicBright:
The python download checksum is given in MD5:

![image](https://user-images.githubusercontent.com/1225659/36835002-05f9dbfc-1d35-11e8-9b08-a4cc898b020d.png)

In my understanding, MD5 security is completely broken and thus MD5 should no longer be used. See https://en.wikipedia.org/wiki/MD5#Security for example.

So I kindly ask you to update your checksum algorithm. For example you could provide a SHA-256 checksum instead.

ned-deily:
Thank you for the suggestion and it is something that is worth doing.  It will require changes to both the web site and our release management process and both are volunteer-led efforts.  More importantly, for each release file we already provide a GPG signature file (via the SIG link after each file) which provides more robust verification than a simple checksum.

bordencastle:
This is a great suggestion, or at least do what nodejs does where they place the SHA256 checksum in a text file at the same location as the download, so that checksum validation is more scriptable. Example.... https://nodejs.org/download/release/v16.20.0

hugovk:
More recently, release artifacts have been signed with Sigstore, [for example](https://www.python.org/downloads/release/python-3126/):

![image](https://github.com/user-attachments/assets/b7ea8f6b-a980-4478-bc12-1fb87cecba2f)

More info: https://www.python.org/download/sigstore/

@sethmlarson Do you think it's worth switching MD5 for something like SHA-256, or are people better off using Sigstore instead? Should we drop MD5?

ezio-melotti:
Checking the MD5 is easy, and as long as it's provided, there will be people using that.
If the MD5 can't be relied upon, it will just give people a false sense of security.
Removing it will encourage people to use other more reliable methods (which should be documented clearly).

JacobCoffee:
@ezio-melotti, @hugovk What do you all think about the future of this issue? Should we close it, or point to sigstore, or something else?

sethmlarson:
I would like to provide checksums that aren't MD5 (in addition to making the Sigstore verification steps more visible). Back-filling the correct values is in progress (cc @woodruffw) since we need to make sure the artifacts are the correct ones first (ie by verifying GPG sigs)

ezio-melotti:
ISTM that two things could be done:
* replace MD5 checksums with SHA-256 ones
* add a paragraph that explains briefly how to verify the downloads and/or links to a different page with more exhaustive instructions

woodruffw:
[...]

PR #2867, opened 2026-01-13
Only show MD5 or SHA-256 columns if there's data to show

hugovk:
#### Description

Part of https://github.com/python/pythondotorg/issues/1227.

The table is now very wide with SHA-256 checksums:

image

This PR:

* Move checksums to the last columns
* Only show a checksum column if there's checksums available
* If there's both SHA-256 and MD5, only show SHA-256
* Use sentence case for headers and add hyphen to SHA-256

### DIFF
--- a/downloads/templatetags/download_tags.py
+++ b/downloads/templatetags/download_tags.py
@@ -97,6 +97,16 @@ def has_sbom(files):
 
 
 @register.filter
+def has_md5(files):
+    return any(f.md5_sum for f in files)
+
+
+@register.filter
+def has_sha256(files):
+    return any(f.sha256_sum for f in files)
+
+
+@register.filter
 def sort_windows(files):
     if not files:
         return files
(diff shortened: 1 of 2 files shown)


JacobCoffee:
i was wondering if its better to have it full on hover/clickable to copy but by default it is clipped to the first.. 8? 16? characters.

JacobCoffee (approved):

hugovk:
Improved with this PR:

image

But I think we can do better. Here's a 10px Courier New:

image

---

> i was wondering if its better to have it full on hover/clickable to copy but by default it is clipped to the first.. 8? 16? characters.

Are SHA-256 often shown in the first X chars? Git's SHA-1s are often shortened.

If we showed less, we'd need to make sure the full thing is still shown with JavaScript disabled, and is also accessible.

One idea is to not show them in the table at all, but to put them in a single page/file: https://github.com/python/pythondotorg/issues/1227#issuecomment-1560112159. This could be generated by the Django downloads app.

PR #2852, opened 2025-12-16
Deduplicate CI spam

JacobCoffee:
#### Description

- removes multiple runs in PRs. we still run on merge, and in PRs but we dont run twice in PRs
- pulls migrations checkk into its own step
- spaces are freee

before:
image

after:
image

PR #2868, opened 2026-01-13
update GHA conditional to allow running on merge to main

JacobCoffee:
#### Description

- In https://github.com/python/pythondotorg/pull/2852 it seems it caused some checks to not run on merge on `main` which may or may not affect cabotage deploys

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   migrations:
-    if: github.event_name != 'push' || github.event.repository.fork == true
+    if: github.event_name != 'push' || github.event.repository.fork == true || github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     services:
       postgres:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -92,4 +92,4 @@ jobs:
           DATABASE_URL: postgres://postgres:postgres@localhost:5432/pythonorg
 
       - name: Check coverage
-        run: coverage report -m --fail-under=75
\ No newline at end of file
+        run: coverage report -m --fail-under=75
```
JacobCoffee, line 95: wat

JacobCoffee: oh, i used vim and it has hooks to add newline

hugovk (approved):

ewdurbin:
tl;dr: Cabotage was correctly handling all hooks, but creation of the deployment failed because by default all required checks on the repo must pass. We have `test` as a required check, so after getting the web hook for a merge to main, it was unable to create a GitHub deployment object since `test` was skipped.

ewdurbin (approved):

Issue #2567, opened 2024-09-16
tests: fix flaky sponsor test 


JacobCoffee:
This is a flaky test that fails periodically after latest updates
https://github.com/python/pythondotorg/actions/runs/10887332034/job/30209329967?pr=2527

```

======================================================================
ERROR: test_save_info_for_image_asset (sponsors.tests.test_forms.SponsorRequiredAssetsFormTest.test_save_info_for_image_asset)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/django/db/backends/utils.py", line 87, in _execute
    return self.cursor.execute(sql)
           ^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.ForeignKeyViolation: insert or update on table "sponsors_benefitfeatureconfiguration" violates foreign key constraint "sponsors_benefitfeat_polymorphic_ctype_id_0ce3d544_fk_django_co"
DETAIL:  Key (polymorphic_ctype_id)=(100) is not present in table "django_content_type".


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/django/test/testcases.py", line 419, in _setup_and_call
    self._post_teardown()
  File "/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/django/test/testcases.py", line 1279, in _post_teardown
    self._fixture_teardown()
  File "/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/django/test/testcases.py", line 1509, in _fixture_teardown
    connections[db_name].check_constraints()
  File "/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/django/db/backends/postgresql/base.py", line 383, in check_constraints
    cursor.execute("SET CONSTRAINTS ALL IMMEDIATE")
  File "/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/django/db/backends/utils.py", line 80, in _execute_with_wrappers
[...]

PR #2876, opened 2026-01-19
Fix tests: Intermittent FK constraint violation on polymorphic_ctype_id

bckohan:
#### Description

#2567 documents intermittent failures on tests involving the polymorphic models in the sponsors directory. The exception is always along the lines of:

```
psycopg2.errors.ForeignKeyViolation: insert or update on table "..." violates foreign key constraint "..."
DETAIL:  Key (polymorphic_ctype_id)=(100) is not present in table "django_content_type".
```

[django-polymorphic](https://pypi.org/project/django-polymorphic) uses the ``polymorphic_ctype`` field to track the concrete instance type on the base table. It sets this field automatically. Whats happening here is that *model bakery is overriding this field and setting it to a random content type*. Most of the time this passed the foreign key constraint - but ``ContentType`` objects are cached and ``mailing/tests/models.py`` created a temporary model (``MockEmailTemplate``) in a transaction that is rolled back after the mailing tests are complete. The cache that model bakery was drawing content types from was not cleared however. You get the fk constraint error when it happened to randomly set the content type of a polymorphic model to ``MockEmailTemplate``.

There are two parts to the fix here:

1. I explicitly clear the ContentType cache after the mailing tests run. This is not strictly necessary because 2 fixes the real problem.
2. I configure the test to [use a custom model baker class](https://model-bakery.readthedocs.io/en/latest/how_bakery_behaves.html#customizing-baker) that detects polymorphic models and ignores the polymorphic_ctype field on them, allowing the base class to set it properly. I think this is the best approach because:
    -    I looked into using recipes but you'd have to specify a recipe for each polymorphic model type and the diff in the   tests would be huge.
    -    This method will also work for any future accidental uses of baker.make(PolymorphicModel).
    -    It is my hope that a future major release of [django-polymorphic](https://pypi.org/project/django-polymorphic) will no longer rely on the contenttypes framework, eliminating the need for special model bakery logic for polymorphic models.

#### Closes

- #2567

### DIFF
--- a/mailing/tests/test_forms.py
+++ b/mailing/tests/test_forms.py
@@ -1,5 +1,6 @@
 """Tests for mailing app forms."""
 from django.test import TestCase
+from django.contrib.contenttypes.models import ContentType
 
 from mailing.tests.forms import TestBaseEmailTemplateForm
 
@@ -12,6 +13,10 @@ class BaseEmailTemplateFormTests(TestCase):
             "subject": "Hello",
             "internal_name": "notification 01",
         }
+
+    def tearDown(self):
+        super().tearDown()
+        ContentType.objects.clear_cache()
 
     def test_validate_required_fields(self):
         required = set(self.data)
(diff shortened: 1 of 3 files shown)


JacobCoffee (approved):

JacobCoffee:
thanks @bckohan!

PR #2885, opened 2026-02-06
Run `django-upgrade` with `--target-version=5.2`

hugovk:
#### Description

- Follow on from https://github.com/python/pythondotorg/pull/2741.
- Also `prek autoupdate` the others.


JacobCoffee:
we did this in another pr :D

PR #2886, opened 2026-02-08
added password_reset and password_reset_done templates

### LINKED ISSUE #2617: Bug: RESET password page don't follow same styling as signin page
### Describe the bug

RESET password page don't follow same styling as sign in page.
image
image

### To Reproduce

1. Goto forgot password while sign in

### Expected behavior

image

### URL to the issue

https://www.python.org/accounts/password/reset/

### Screenshots

_No response_

### Browsers

Chrome

### Operating System

macOS

### Browser Version

Version 128.0.6613.120

### Relevant log output

_No response_

### Additional context

_No response_

jamie-ferguson:
#### Description

Fixes missing password reset templates in the accounts pages.

Previously, attempting to reset a password would show unstyled pages at the following urls
/accounts/password/reset
/accounts/password/reset/done/

This PR adds two new template files which fixes this.  

Tested locally using the Docker setup and triggering the password reset workflow.

#### Closes

Closes #2617

### DIFF
--- a/templates/account/password_reset.html
+++ b/templates/account/password_reset.html
@@ -1,0 +1,20 @@
+{% extends "account/base.html" %}
+{% load i18n account %}
+
+{% block page_title %}{% trans "Password Reset" %}{% endblock page_title %}
+
+{% block content %}
+    <h1>{% trans "Password Reset" %}</h1>
+
+    <p>{% trans "Forgotten your password? Enter your email address below and we'll send you an email allowing you to reset it." %}</p>
+
+    <form method="post" action="{% url 'account_reset_password' %}" class="password_reset">
+        {% csrf_token %}
+        {{ form.as_p }}
+        {% if redirect_field_value %}
+            <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}" />
+        {% endif %}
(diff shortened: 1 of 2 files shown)


jefftriplett:
Thanks Jamie, and sorry this sat so long. I checked it out locally and both pages render correctly inside the site chrome, the styling matches `login.html`, and CI is green.

One thing before we merge: this covers the first half of the reset flow but not the second. The page a user lands on from the emailed link still falls through to allauth's own templates, with no site nav at all:

```
/accounts/password/reset/key/<key>/
  account/password_reset_from_key.html
  account/base_entrance.html
  allauth/layouts/entrance.html
  allauth/layouts/base.html
```

So requesting a reset looks right, then clicking the link in the email drops you onto a bare allauth page to type the new password. The two still needed are:

* `templates/account/password_reset_from_key.html`
* `templates/account/password_reset_from_key_done.html`

They should be short, in the same shape as the two here. If you would like to finish it off, great. If you would rather not, say the word and we will add them and get this landed.

jamie-ferguson:
Hi, I've added the other two files requested and successfully tested them locally. This should be good to go now.

Thanks.

PR #2892, opened 2026-02-17
Fix DRF throttle behinnd proxy

### LINKED ISSUE #2891: https://www.python.org/api/v2/downloads/release/ endpoint returns error
### Describe the bug

Hello everybody,
I hope, this is the right place for opening this issue.
Since 2026-02-12, we receive an error in renovate, when the python dependency should be updated.
This seems to be, because the url that is used to fetch the versions (https://www.python.org/api/v2/downloads/release/) returns an error: `{"detail":"Request was throttled. Expected available in 42578 seconds."}` The number of seconds varies. I also get the error when I open the link in the browser, so I think, it is not a renovate issue.
Do you know by chance what is going on here?
Thanks a lot in advance for your help!

### To Reproduce

Go to https://www.python.org/api/v2/downloads/release/
Error is displayed: ` {"detail":"Request was throttled. Expected available in 42578 seconds."}`

### Expected behavior

No error is returned

### URL to the issue

https://www.python.org/api/v2/downloads/release/

### Screenshots

```bash

```

### Browsers

Chrome

### Operating System

Linux

### Browser Version

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

JacobCoffee:
Fixes https://github.com/python/pythondotorg/issues/2891

### DIFF
--- a/pydotorg/settings/base.py
+++ b/pydotorg/settings/base.py
@@ -343,6 +343,7 @@ REST_FRAMEWORK = {
         'anon': '100/day',
         'user': '3000/day',
     },
+    'NUM_PROXIES': 3,
 }
 
 ### pydotorg.middleware.GlobalSurrogateKey


pydotorg/settings/base.py
```diff
@@ -343,6 +343,7 @@
         'anon': '100/day',
         'user': '3000/day',
     },
+    'NUM_PROXIES': 3,
```
ewdurbin, line 346: Before:

AWS ALB (Network) -> nginx-ingress -> nginx (container) -> gunicorn

After:

AWS ELB (Classic) ->. Traefik -> nginx (container) -> gunicorn

So... yeah I think 3 is correct since ALB doesn't do anything with X-Forwarded-For.

But why did it work before without being configured as 2?

JacobCoffee: maybe just traefik behaving differently from nginx ingress?

PR #2902, opened 2026-02-18
fix: ci and group pkg installs

JacobCoffee:
## Summary
- Use `pip install --group dev` (PEP 735) instead of `pip install '.[dev]'` in CI and Dockerfile
  - Dev deps are under `[dependency-groups]`, not `[project.optional-dependencies]`
  - This was causing all dev deps (coverage, debug_toolbar, factory-boy, etc.) to silently not install
- Guard `import debug_toolbar` in `urls.py` and `local.py` with scoped `ModuleNotFoundError` check
  - Only suppresses when `debug_toolbar` itself is missing; re-raises transitive import failures

## Test plan
- [ ] CI migrations check passes
- [ ] CI test suite runs (coverage, factory-boy, etc. installed)
- [ ] Local `make serve` works with Docker

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,8 +39,8 @@ COPY pyproject.toml /code/
 RUN --mount=type=cache,target=/root/.cache/pip \
     set -x \
     && pip --disable-pip-version-check \
-        install \
-        '.[dev]'
+        install --group dev \
+        .
 
 COPY . /code/
 
--- a/pydotorg/settings/local.py
+++ b/pydotorg/settings/local.py
@@ -25,13 +25,19 @@ HAYSTACK_CONNECTIONS = {
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 
 
-INSTALLED_APPS += [
-    "debug_toolbar",
-]
+try:
+    import debug_toolbar  # noqa: F401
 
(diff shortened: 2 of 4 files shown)


pydotorg/settings/local.py
```diff
@@ -25,13 +25,19 @@
...
+    import debug_toolbar  # noqa: F401
+
+    INSTALLED_APPS += [
+        "debug_toolbar",
+    ]
+
+    MIDDLEWARE += [
+        "debug_toolbar.middleware.DebugToolbarMiddleware",
+    ]
+except ModuleNotFoundError as exc:
```
JacobCoffee, line 38: feel like this would be better with contextlib suppress

JacobCoffee: maybe itd be too broad.

Dockerfile
```diff
@@ -39,8 +39,8 @@ COPY pyproject.toml /code/
 RUN --mount=type=cache,target=/root/.cache/pip \
     set -x \
     && pip --disable-pip-version-check \
-        install \
-        '.[dev]'
+        install --group dev \
+        .
```
JacobCoffee, line 43: gross

PR #2932, opened 2026-02-21
Purge CDN selectively

JacobCoffee:
#### Description

- Since we reorganized we can more easily tell which templates were changed and where we need to purge.


JacobCoffee:
All three review comments addressed in 60d913d1:
- Uses `github.event.before..github.sha` for full push range diff
- Updated to `actions/checkout@v6`
- Condition now uses `github.event_name == 'push'`

PR #2934, opened 2026-02-22
Fix #2626: Duplicate scheduled year for single day events

### LINKED ISSUE #2626: Duplicate scheduled year for single day events whose scheduled year is rendered
## What

For **single day events** (events that start and end on the same day) that require year string to be displayed [_which are_]:
- Single day events scheduled for a future year
- Single day events that have passed rendered at the "You just missed..." section of the main events page **{pythondotorg url}/events/**

The year string is rendered multiple times.

### 1. Past event scenario
image

### 2. Future event scenario
image

### To Reproduce

- Set up a single day event scheduled to start in the future year or scheduled in the past (e.g. event has just been concluded)
- Navigate to the main events page (http://localhost:8000/events) and observe the year string rendered multiple times.

### Expected behavior

- Single day events (regardless of schedule year) should only render a single year string (if necessary) since the event is guaranteed to start and end within the same year.

### Addressing "If necessary"
- Future single day event scheduled within the current year does not need to render the current year as there is no date/year ambiguity.
- Past single day events (Events just missed) would need to render the year string dispelling any ambiguity on year event took place.

### URL to the issue

_No response_

### Screenshots

_No response_

### Browsers

Firefox, Chrome, Safari, Microsoft Edge, Other

### Operating System

Linux

### Browser Version

_No response_

### Relevant log output

_No response_

### Additional context

I've traced down the offending template logic:
https://github.com/python/pythondotorg/blob/1bfeace44973dff8310f695ca15b40c277047cc7/templates/events/includes/time_tag.html#L1-L9

iampujan:
Fixes #2626. Removed duplicate render for `dt_start|date:"Y"` in `time_tag.html` when `single_day` is true. Includes a unit test to enforce single-year rendering for same day events.


jefftriplett:
Thanks for this, and sorry it sat so long. The only thing failing was `ruff format`: the new test file was wrapped narrower than the project's line length. Since maintainer edits were enabled I pushed the formatting commit to your branch rather than make you round-trip it. `ruff check` and the full events suite were already passing.

jefftriplett (approved):
LGTM once that lint check passes

PR #2935, opened 2026-02-22
Handle malformed release versions in docs versions view

adityagiri3600:
#### Description

- Skip malformed release versions in `DocsByVersionView` when a release name does not include at least `major.minor` parts.
- Add a regression test to ensure `/doc/versions/` still returns HTTP 200 when such a release exists.

#### Closes

-

### DIFF
--- a/pydotorg/tests/test_views.py
+++ b/pydotorg/tests/test_views.py
@@ -34,3 +34,13 @@ class ViewsTests(TestCase):
         self.assertContains(response, "Browse Python 3.6.0 Documentation")
         self.assertContains(response, "https://docs.python.org/3/whatsnew/3.6.html")
         self.assertContains(response, "What's new in Python 3.6")
+
+    def test_docs_versions_ignores_malformed_release_version(self):
+        Release.objects.create(
+            name="Python 4",
+            is_published=True,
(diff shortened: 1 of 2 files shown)


hugovk:
Is this about https://www.python.org/doc/versions/ ?

What exactly is malformed?

adityagiri3600:
Yes.

By malformed I mean an incomplete release version that still matches the current regex like `Python 4` matches `Python ([\d.]+)`, but it doesn’t include at least `major.minor`. So `full_version.split(".")` has only one element, and the previous code accessed `version_parts[1]`, which can  raise `IndexError` and break the page. 

If you prefer not to accept major only strings at all, I can update this to tighten parsing instead like require X.Y or X.Y.Z via a stricter regex instead of `IndexError`.

hugovk:
I don't think this is needed, we'll never have a simple Python X release.

Python releases (with docs) are always X.Y.Z, and in the past we've had X.Y (the last was in 2011).

(Pre-releases can have a1, b1, rc1 suffixes, but they don't have docs.)

adityagiri3600:
That makes sense, I appreciate this review. I'll close this PR.

PR #2945, opened 2026-02-24
Fix deprecation warnings in dependencies

hugovk:
#### Description

There's lots of deprecation warnings at https://github.com/python/pythondotorg/actions/runs/22328504402/job/64605126832#step:9:13

Most are from Django, but these are from dependencies:

```
/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/dateutil/tz/tz.py:37: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
  EPOCH = datetime.datetime.utcfromtimestamp(0)
/opt/hostedtoolcache/Python/3.12.6/x64/lib/python3.12/site-packages/pytz/tzinfo.py:27: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
  _epoch = datetime.utcfromtimestamp(0)
```

The first is from dateutil. We can fix by upgrading to the latest version.

The second is from pytz. We only use `pytz.UTC` from here, and can instead replace it with the stdlib's `datetime.UTC`, and ditch the dependency altogether.

And run `uv lock` to update `uv.lock`.

### DIFF
--- a/apps/events/utils.py
+++ b/apps/events/utils.py
@@ -3,7 +3,6 @@
 import datetime
 import re
 
-import pytz
 from django.utils.timezone import is_aware, make_aware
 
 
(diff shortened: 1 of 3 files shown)


JacobCoffee:
thanks!

PR #2947, opened 2026-02-25
Force 'https://www.python.org/' URLs for ReleaseFile properties

sethmlarson:
#### Description

Adds a new `CheckConstraint` that forces all `URLField` fields on `ReleaseFile` model to not accept URLs that don't start with `http(s)://www.python.org/`.


hugovk:
Locally, if I change the URL for a file of an existing release to `www.example.com/[etc]`, clicking save correctly gives an error:

> Constraint “only_python_dot_org_urls” is violated.

image

Also when creating a new release with a file.

But can we make that error more descriptive?

I don't need to know what a "constraint" is; tell me what's not allowed and what is.

sethmlarson:
Done in https://github.com/python/pythondotorg/pull/2947/commits/93e2795498a604ddf5f67d3831470a174223632a, the new message is "All file URLs must begin with 'https://www.python.org/'".

apps/downloads/models.py
```diff
@@ -357,6 +357,17 @@ def update_boxes_on_release_file_delete(sender, instance, **kwargs):
...
 
 
+def condition_url_is_blank_or_python_dot_org(column: str):
+    """Conditions for a URLField column to force 'http[s]://python.org'."""
+    return (
+        models.Q(**{f"{column}__exact": ""})
+        | models.Q(**{f"{column}__startswith": "https://www.python.org/"})
+        # Older releases allowed 'http://'. 'https://' is required at
+        # the API level, so shouldn't show up in newer releases.
+        | models.Q(**{f"{column}__startswith": "http://www.python.org/"})
```
hugovk, line 367: Not as important, but we could update those old `http://` to `https://`

Could also be a followup.

hugovk (approved):
Thanks!

sethmlarson:
FYI: The `Lint` action failed to install, I ran `make lint` locally and received no errors.

hugovk:
Re-ran lint and passed: https://github.com/python/pythondotorg/actions/runs/22413902653/job/64899885730

JacobCoffee (approved):

PR #2965, opened 2026-03-18
fix: remap dead hg.python.org release-notes URLs to GitHub equivalents

### LINKED ISSUE #2865: Bug: Dead "Release notes" links on the download page for old releases
### Describe the bug

"Release notes" links on the Downloads page for releases 3.3.6 and before are broken. Example link:

https://hg.python.org/cpython/file/v3.3.6/Misc/NEWS

### To Reproduce

1. Go to Downloads pages https://www.python.org/downloads/
2. Click "Release notes" for a release before 3.3.6

### Expected behavior

Should navigate to the release notes for that release.

### URL to the issue

### Screenshots

```bash

```

### Browsers

Firefox

### Operating System

Linux

### Browser Version

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

nagasrisai:
The "Release notes" links on the downloads page for Python 3.3.6 and earlier are broken — they point to hg.python.org which was retired and now returns 404.

Added a `corrected_release_notes_url` property to the `Release` model that checks whether a stored URL points to hg.python.org and, if so, remaps it to the equivalent path on GitHub. For example, `http://hg.python.org/cpython/file/v3.3.6/Misc/NEWS` becomes `https://github.com/python/cpython/blob/v3.3.6/Misc/NEWS`. The database field itself is left untouched so no migration is needed.

Also updated the downloads index template to use the corrected URL, and added a guard so releases with no URL at all render as plain text rather than an empty anchor.

Four tests added to `ReleaseNotesURLTests` covering the http and https hg variants, a modern URL that should pass through unchanged, and an empty URL.

Closes #2865

### DIFF
--- a/apps/downloads/models.py
+++ b/apps/downloads/models.py
@@ -123,18 +123,29 @@ class Release(ContentManageable, NameSlugModel):
         the "Release notes" links on the downloads page still work for legacy
         releases (3.3.6 and earlier).
 
-        Example::
+        Examples::
 
             http://hg.python.org/cpython/file/v3.3.6/Misc/NEWS
             → https://github.com/python/cpython/blob/v3.3.6/Misc/NEWS
+
+            http://hg.python.org/cpython/raw-file/v2.7.3/Misc/NEWS
+            → https://raw.githubusercontent.com/python/cpython/v2.7.3/Misc/NEWS
         """
         url = self.release_notes_url
         if not url:
             return url
-        match = re.match(r"https?://hg\.python\.org/cpython/file/([^/]+)/(.+)", url)
-        if match:
-            tag, path = match.group(1), match.group(2)
-            return f"https://github.com/python/cpython/blob/{tag}/{path}"
+        for prefix in (
+            "http://hg.python.org/cpython/file/",
+            "https://hg.python.org/cpython/file/",
+        ):
+            if url.startswith(prefix):
+                return "https://github.com/python/cpython/blob/" + url[len(prefix):]
+        for prefix in (
+            "http://hg.python.org/cpython/raw-file/",
+            "https://hg.python.org/cpython/raw-file/",
+        ):
+            if url.startswith(prefix):
+                return "https://raw.githubusercontent.com/python/cpython/" + url[len(prefix):]
         return url
 
     def download_file_for_os(self, os_slug):
--- a/apps/downloads/models.py
+++ b/apps/downloads/models.py
@@ -114,6 +114,29 @@ class Release(ContentManageable, NameSlugModel):
             return self.release_page.get_absolute_url()
         return reverse("download:download_release_detail", kwargs={"release_slug": self.slug})
 
+    @property
+    def corrected_release_notes_url(self):
+        """Return the release notes URL, converting dead hg.python.org links to GitHub.
+
+        Old Mercurial-hosted URLs (hg.python.org) are no longer reachable.
+        This property remaps them to their equivalent paths on GitHub so that
+        the "Release notes" links on the downloads page still work for legacy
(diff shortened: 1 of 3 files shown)


nagasrisai:
@JacobCoffee @ewdurbin @sethmlarson — could one of you take a look when you get a chance? This fixes the broken release notes links on the downloads page for older Python releases. No migration needed. Thanks!

nagasrisai:
Looks like the Actions workflows need maintainer approval to run since this is coming from a fork — it's the first contribution from this account. Could @JacobCoffee or @ewdurbin approve the three workflow runs (CI, Lint, Check collectstatic) when you get a chance? The links are in the Checks tab on this PR. Thanks!

apps/downloads/models.py
```diff
@@ -114,6 +114,29 @@ def get_absolute_url(self):
...
+            http://hg.python.org/cpython/file/v3.3.6/Misc/NEWS
+            → https://github.com/python/cpython/blob/v3.3.6/Misc/NEWS
+        """
+        url = self.release_notes_url
+        if not url:
+            return url
+        match = re.match(r"https?://hg\.python\.org/cpython/file/([^/]+)/(.+)", url)
+        if match:
+            tag, path = match.group(1), match.group(2)
+            return f"https://github.com/python/cpython/blob/{tag}/{path}"
```
hugovk: `re.match` can be confusing because it anchors the start of the string, but not the end.

Can we use `re.fullmatch` instead? Or `re.search` with explicit `^`?

https://docs.python.org/3/library/re.html#search-vs-match

---

Or better yet, ditch the regex:

```suggestion
        for prefix in (
            "http://hg.python.org/cpython/file/",
            "https://hg.python.org/cpython/file/",
        ):
            if url.startswith(prefix):
                return "https://github.com/python/cpython/blob/" + url[len(prefix):]
        return url
```

hugovk: Actually:

* https://www.python.org/downloads/release/python-336/ uses a URL with `/file/`: `https://hg.python.org/cpython/file/v3.3.6/Misc/NEWS` 
* https://www.python.org/downloads/release/python-273/ uses a URL with `/raw-file/`: `http://hg.python.org/cpython/raw-file/v2.7.3/Misc/NEWS`

We should handle both.

hugovk (commented):
Rather than rewriting the URLs on serve, would this be better as a one-off database migration?

apps/downloads/models.py
nagasrisai: Good point, the `str.startswith` approach is much cleaner. Will switch to that — also makes it easier to extend for the `raw-file` case you spotted below.

nagasrisai: Good catch, missed that one. Will handle both `/file/` and `/raw-file/` in the updated version.

nagasrisai:
Thanks for the thorough review! On the migration question — happy to go either way. The on-serve approach felt lighter since there are only a handful of affected legacy releases, but a one-off migration is cleaner long-term and avoids the runtime check on every page load. I'll update the PR to use `str.startswith` covering both `/file/` and `/raw-file/`, and let me know if you'd prefer I also swap this out for a data migration instead.

nagasrisai:
Updated — thanks for the thorough review!

Two changes made per your feedback:

1. Replaced `re.match` with `str.startswith` as suggested. The loop approach also makes it straightforward to extend.
2. Added handling for the `/raw-file/` variant you spotted on the Python 2.7.3 release page — those map to `raw.githubusercontent.com` since that's the direct equivalent of Mercurial's raw file view.

Also added two new test cases covering the HTTP and HTTPS raw-file variants.

PR #2966, opened 2026-03-18
feat: add linkable heading anchors to PSF pages

### LINKED ISSUE #2349: Enhancement: Add page anchor tags for easy sharing of PSF Board resolutions
It is difficult to share specific resolutions from the official, definitive source to my Python peers from https://www.python.org/psf/records/board/resolutions/

If we had IDs of the resolution that were formed perhaps by concatenating the date the resolution was formed and the resolution # of the meeting to give anchor tags of the resolutions web page: https://www.python.org/psf/records/board/resolutions/ and allow for easier search and sharing of the PSF Resolutions. 

This would also be helpful with announcements.

I usually just screenshot or tell them the date and the name of the resolution, which is less than ideal because it removes the "source of truth" traceability. 

**Additional context**

nagasrisai:
Adds linkable anchor IDs to headings on PSF pages so that individual sections can be shared as direct URLs, which is what #2349 is asking for.

Added a `add_heading_anchors` template filter in `apps/pages/templatetags/page_tags.py`. It processes the rendered page HTML and injects an `id` attribute into each h2, h3, and h4 element based on the slugified heading text, along with a small pilcrow (¶) anchor link. Applied the filter in `templates/psf/default.html` so it covers the board resolutions page and all other PSF pages automatically.

Duplicate heading texts get a `-2`, `-3` suffix to keep IDs unique, and headings that already carry an `id` are left untouched. Ten tests included in `apps/pages/tests/test_templatetags.py`.

Closes #2349

### DIFF
--- a/apps/pages/templatetags/page_tags.py
+++ b/apps/pages/templatetags/page_tags.py
@@ -1,0 +1,69 @@
+"""Custom template tags and filters for the pages app."""
+
+import re
+
+from django import template
+from django.utils.safestring import mark_safe
+from django.utils.text import slugify
+
+register = template.Library()
+
+# Match h2–h4 elements; capture tag name, existing attributes, and inner HTML.
+# Using DOTALL so inner content can span multiple lines.
+_HEADING_RE = re.compile(r"<(h[2-4])([^>]*)>(.*?)</\1>", re.IGNORECASE | re.DOTALL)
+
+
+@register.filter(is_safe=True)
+def add_heading_anchors(html):
+    """Add ``id`` attributes and self-link anchors to h2–h4 headings.
+
+    Given the rendered HTML of a CMS page, this filter finds every ``<h2>``,
+    ``<h3>``, and ``<h4>`` element and:
+
+    1. Derives a URL-safe ``id`` from the heading's plain text (via
+       :func:`django.utils.text.slugify`).
+    2. Appends a ``-N`` suffix when the same slug appears more than once on
+       the page, preventing duplicate ``id`` values.
+    3. Injects a pilcrow (¶) anchor *inside* the heading so visitors can
+       copy a direct link to any section.
+
+    The filter is safe to apply to any page — headings that already carry an
+    ``id`` attribute are left untouched, and headings whose text produces an
+    empty slug are also skipped.
+
+    Usage in a template::
+
+        {% load page_tags %}
+        {{ page.content|add_heading_anchors }}
+    """
+    seen_slugs: dict[str, int] = {}
+
+    def _replace(match: re.Match) -> str:
+        tag = match.group(1).lower()
+        attrs = match.group(2)
+        inner = match.group(3)
+
+        # Skip headings that already have an id attribute.
+        if re.search(r'\bid\s*=', attrs, re.IGNORECASE):
+            return match.group(0)
+
+        # Derive a slug from the plain text (strip any nested HTML tags).
+        plain_text = re.sub(r"<[^>]+>", "", inner).strip()
+        base_slug = slugify(plain_text)
+
+        if not base_slug:
+            return match.group(0)
+
+        # Deduplicate: first occurrence keeps the bare slug; subsequent
+        # occurrences become slug-2, slug-3, …
+        count = seen_slugs.get(base_slug, 0) + 1
+        seen_slugs[base_slug] = count
+        anchor_id = base_slug if count == 1 else f"{base_slug}-{count}"
+
+        link = (
+            f'<a class="headerlink" href="#{anchor_id}" '
+            f'title="Link to this section" aria-label="Link to this section">¶</a>'
+        )
+        return f'<{tag} id="{anchor_id}"{attrs}>{inner} {link}</{tag}>'
+
+    return mark_safe(_HEADING_RE.sub(_replace, str(html)))
--- a/apps/pages/templatetags/page_tags.py
+++ b/apps/pages/templatetags/page_tags.py
@@ -8,28 +8,35 @@ from django.utils.text import slugify
 
 register = template.Library()
 
-# Match h2–h4 elements; capture tag name, existing attributes, and inner HTML.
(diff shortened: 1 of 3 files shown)


nagasrisai:
@JacobCoffee @ewdurbin @sethmlarson — could one of you review when you get a chance? This adds anchor IDs to headings on PSF pages so sections like the board resolutions can be linked to directly. 10 tests included. Thanks!

nagasrisai:
The Actions workflows need maintainer approval to run since this is coming from a fork. Could @JacobCoffee or @ewdurbin approve the CI, Lint, and Check collectstatic runs from the Checks tab when you get a chance? Thanks!

JacobCoffee:
@nagasrisai hi please dont @ specific people, CODEOWNERS will handle assignment for review

apps/pages/templatetags/page_tags.py
```diff
@@ -0,0 +1,69 @@
...
+
+import re
+
+from django import template
+from django.utils.safestring import mark_safe
+from django.utils.text import slugify
+
+register = template.Library()
+
+# Match h2–h4 elements; capture tag name, existing attributes, and inner HTML.
```
hugovk: Why skip H1?

nagasrisai: @hugovk  h1 is the page title ,there's only ever one per page and the page URL already points to it, so an anchor on it wouldn't be useful. The sections people actually want to link into are h2 and below (e.g. individual board resolutions or meeting minutes). This is also the convention most docs sites follow, including GitHub's own markdown renderer.

hugovk: There are some pages with multiple H1s, such as https://www.python.org/downloads/release/python-31312/.

Does this only run on pages of board meeting resolutions/minutes? Do they all at most have a single H1?

nagasrisai: @hugovk  Good catch ,the filter is only wired up to `templates/psf/default.html`, not the general pages template. So the Python release pages (like the downloads page you linked) go through a different template and won't be touched by this at all.

For PSF pages specifically, the H1 is always the page title coming from the CMS `page.title` field, rendered directly in the template as `<h1 class="page-title">{{ page.title }}</h1>` — separate from `page.content` where the filter runs. So the content fed into `add_heading_anchors` shouldn't contain any H1s.

That said, happy to extend the regex to h1–h4 if you'd prefer a more defensive approach. just let me know.

hugovk: Have you tested this PR?

nagasrisai:
Yep, tested it locally! Here's what it produces on a few real cases:

A plain h2 heading gets an id and the pilcrow link injected right inside it:
```
<h2 id="2023">2023 <a class="headerlink" href="#2023" title="Link to this section" aria-label="Link to this section">¶</a></h2>
```

Duplicate headings on the same page get deduped cleanly:
```
<h2 id="board-resolution">Board Resolution ¶</h2>
<h2 id="board-resolution-2">Board Resolution ¶</h2>
<h2 id="board-resolution-3">Board Resolution ¶</h2>
```

A heading that already has an id is left completely alone — no double-processing:
```
<h2 id="custom">Already has id</h2>
```

And on something that looks like an actual PSF board minutes page:
```
<h2 id="2024-q1-resolutions">2024 Q1 Resolutions ¶</h2>
<h3 id="resolution-1-budget-approval">Resolution 1: Budget Approval ¶</h3>
<h3 id="resolution-2-new-members">Resolution 2: New Members ¶</h3>
<h2 id="2023-q4-resolutions">2023 Q4 Resolutions ¶</h2>
```

h1 is untouched throughout, and all 10 tests in the test file pass.

hugovk:
I'm confused, because https://www.python.org/psf/records/board/resolutions/ doesn't have any h2-h4 (except one h3 in the sidebar).

nagasrisai:
@hugovk  You're right, and good catch. I looked at the actual page source —the resolutions page content is RST-generated, so the section entries use h1 tags rather than h2-h4, which means the filter wouldn't add pilcrow links there. Those sections do already have IDs on the wrapping divs though, so they're technically linkable, just without a visual anchor.

That said, other PSF pages do have h2/h3 content headings ,the bylaws page for instance has 30+ headings across h2 and h3. That's more the kind of page this would help.

So fair question about scope , if the main goal was specifically the resolutions/minutes pages, the filter would need to cover h1 as well. Happy to extend it to h1-h4 if that makes more sense, or narrow the scope to the pages where h2-h4 headings actually appear. Let me know what you think.

hugovk:
> I looked at the actual page source —the resolutions page content is RST-generated, 

How can you tell?

> so the section entries use h1 tags rather than h2-h4, which means the filter wouldn't add pilcrow links there. 

Whether the source is RST or MD or HTML is irrelevant, all can do headings down to h4 and beyond.

> Those sections do already have IDs on the wrapping divs though, so they're technically linkable, just without a visual anchor.

Are you sure? I don't see it. Please give an example.

> That said, other PSF pages do have h2/h3 content headings ,the bylaws page for instance has 30+ headings across h2 and h3. That's more the kind of page this would help.

OK, but this issue is specifically about resolutions. And when I test the bylaws page with this PR locally, I don't see any linkable headings.

nagasrisai:
@hugovk  You're right on both points, and I've now identified and fixed the two actual bugs.

The bylaws page not showing anchors locally is the main one. RST content in pythondotorg is rendered by docutils, which automatically adds `id` attributes to every section heading. My filter was bailing out completely when it saw an existing `id`, so it effectively did nothing on any RST-sourced page. The fix is to reuse the existing id and inject the pilcrow link using it rather than skipping the heading entirely.

The resolutions page was also missed because the regex only covered h2–h4, and docutils generates those sections as h1. Extending to h1–h4 fixes that.

Two commits just pushed: one for the filter, one to update the tests (added tests for RST-generated headings with existing ids, idempotency, and h1 processing).

Issue #2988, opened 2026-04-23
Locking down dependencies for www.python.org


sethmlarson:
### Describe the bug

- [x] Install from a lock file with hashes.
- [x] Enable Dependabot with cooldowns to upgrade dependencies over time.
- [x] Using Python 3.12.6 (3.12.10 is available). Should figure out a reliable upgrade strategy.
- [ ] Remove external dependencies that aren't needed:
  - [ ] Installing old pandoc from GitHub releases. Can we use `pypandoc-binary`?
  - [ ] Using `ewdurbin/ngix-static` Dockerfile, likely outdated. Can we manage Nginx separately?

### To Reproduce

N/A

### Expected behavior

N/A

### Screenshots

```bash
N/A
```

### Browsers

Other

### Operating System

Other

### Relevant log output

```shell

```

sethmlarson:
Completed in https://github.com/python/pythondotorg/pull/2994, https://github.com/python/pythondotorg/pull/2993, and https://github.com/python/pythondotorg/pull/2989 🥳

PR #2989, opened 2026-04-23
Use uv.lock for installing dependencies in Dockerfiles

sethmlarson:
Related to #2988. We also have many outstanding Dependabot PRs, so maybe even "weekly" is too often. Security updates will always get prioritized, so we don't have to worry too much about this time span being long.

I deleted `uv.lock` because it wasn't being used at all for deployment and appears to have been added accidentally? It was added in a [seemingly unrelated `README` commit](https://github.com/python/pythondotorg/commit/43c8e2a8c63604bc5729171487499e7cbfe4006d).

cc @JacobCoffee


hugovk:
> I deleted `uv.lock` because it wasn't being used at all for deployment and appears to have been added accidentally? It was added in a [seemingly unrelated `README` commit](https://github.com/python/pythondotorg/commit/43c8e2a8c63604bc5729171487499e7cbfe4006d).

Add it to `.gitignore`?

---

Looks like this weekend's pip release will add experimental support for installing from pylock.toml! https://github.com/pypa/pip/pull/13876

Not necessarily suggesting waiting for that, there may be some edges that need polish, and pip-compile is a very good way to lock right now.

JacobCoffee:
We use the `uv.lock` exclusively now (all of the old-style requirements.txt files were removed for this reason). used in makefile targets for docs, etc.

what we could do is migrate dockerfiles over to `uv sync --frozen`

JacobCoffee (changes requested):
dependabot changes +1, the rest needs  undone

sethmlarson:
@JacobCoffee Gotcha, I didn't realize it was used because it wasn't used in the Dockerfiles. I didn't look in the `Makefile`. I'll update this PR to use `uv sync` instead.

sethmlarson:
FYI, used `-requirements.txt` because then the dependency is scannable by Dependabot and they'll offer us upgrades. I don't think that will happen with the current pin of a container within a Dockerfile. This probably doesn't matter as much because it's a installer tool, but figured I would share why I went this route.

JacobCoffee:
i cant leave code review due to platform issues so can you check @sethmlarson 
image

https://github.com/python/pythondotorg/pull/2989/changes#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R35

sethmlarson:
@JacobCoffee Yeah that looks good to me, Jacob and I discussed Dependabot updates and it sounds like Dependabot can handle Dockerfiles too. Neat!

JacobCoffee (approved):

PR #2993, opened 2026-05-06
Remove old exclusions, use dependency cooldowns

sethmlarson:
Updates `dependabot.yml` config to use dependency cooldowns and removes all the exclusions as we'll be updating weekly instead of daily.


JacobCoffee (approved):
may add boto back but good for now

PR #3014, opened 2026-06-02
Validate release-file URL metadata

### LINKED ISSUE #3010: Bug: Release-file sidecar URLs can reference different artifacts
### Describe the bug

Release-file sidecar URL fields are accepted when they point to `python.org`,
even if they do not correspond to the artifact URL on the same row. For example,
a source tarball entry can expose a GPG signature URL for a different tarball,
and the downloads page/API will present that sidecar as belonging to the listed
artifact.

### To Reproduce

1. In a local python.org development environment, create or submit an
   authenticated release-file API write with:
   - `url`: `https://www.python.org/ftp/python/3.14.0/Python-3.14.0.tgz`
   - `gpg_signature_file`:
     `https://www.python.org/ftp/python/3.14.0/Python-3.13.0.tgz.asc`
2. Observe that the row is accepted as long as both URLs use the allowed
   `python.org` prefix.
3. Load the release-file API response or release detail page and observe that
   the sidecar is presented for the artifact on that row.

### Expected behavior

Sidecar fields should either be blank or match the artifact URL with the
expected sidecar suffix, such as `.asc`, `.sig`, `.crt`, `.sigstore`, or
`.spdx.json`.

### Additional context

Authenticated v1/v2 `release_file` POSTs with mismatched sidecar URLs returned
`201` before the patch.

A focused regression test can cover both model validation and authenticated v1/v2
release-file API writes.

### LINKED ISSUE #3011: Bug: Release-file API accepts new HTTP python.org URLs
### Describe the bug

Authenticated release-file writes accept new `http://www.python.org/...` values
for artifact and sidecar URL fields. Older release rows may still contain HTTP
URLs, but new or changed release-file metadata should use HTTPS.

### To Reproduce

1. In a local python.org development environment, submit an authenticated v1 or
   v2 release-file API write with:
   - `url`: `http://www.python.org/ftp/python/3.14.0/Python-3.14.0.tgz`
2. Observe that the API accepts the row.
3. Repeat with a sidecar field such as `gpg_signature_file` using an HTTP URL
   and observe that the write is accepted as well.

### Expected behavior

New or changed release-file URL values should begin with
`https://www.python.org/`. Existing legacy HTTP rows should remain editable for
unrelated metadata until a separate data cleanup can convert or remove them.

### Additional context

Authenticated v1/v2 `release_file` POSTs with `http://www.python.org` URLs
returned `201` before the patch.

The database constraint currently allows `http://www.python.org/...` to preserve
older release data. This can be handled at the validation layer for new writes
without requiring a data migration in the same change.

e-q:
#### Description

- Validate release-file sidecar URLs against the artifact URL they describe,
  including existing mismatched sidecars edited through unrelated metadata
  changes.
- Reject new or changed non-HTTPS release-file URL values.
- Apply the validation in model forms, DRF writes, and Tastypie writes.
- Keep unchanged legacy HTTP URL rows editable while rejecting sidecar mismatches
  and new invalid values.
- Add focused model and v1/v2 API regression tests for create and update paths.

#### Closes

- Closes #3010 (sidecar URL issue)
- Closes #3011 (new HTTP URL issue)

#### Testing

- Focused model and v1/v2 API regression tests for release-file URL validation,
  including create and update paths: passed.
- Whitespace check: passed.
- No migrations generated.

### DIFF
--- a/apps/downloads/models.py
+++ b/apps/downloads/models.py
@@ -19,6 +19,24 @@ from apps.pages.models import Page
 from fastly.utils import purge_surrogate_key, purge_url
 
 DEFAULT_MARKUP_TYPE = getattr(settings, "DEFAULT_MARKUP_TYPE", "markdown")
+PYTHON_DOT_ORG_HTTPS_PREFIX = "https://www.python.org/"
+PYTHON_DOT_ORG_HTTP_PREFIX = "http://www.python.org/"
+RELEASE_FILE_URL_FIELDS = (
+    "url",
+    "gpg_signature_file",
+    "sigstore_signature_file",
+    "sigstore_cert_file",
+    "sigstore_bundle_file",
+    "sbom_spdx2_file",
+)
+RELEASE_FILE_SIDECAR_SUFFIXES = {
+    "gpg_signature_file": ".asc",
+    "sigstore_signature_file": ".sig",
+    "sigstore_cert_file": ".crt",
+    "sigstore_bundle_file": ".sigstore",
+    "sbom_spdx2_file": ".spdx.json",
+}
+RELEASE_FILE_HTTPS_ERROR = "Release file URLs must begin with 'https://www.python.org/'."
 
 
 class OS(ContentManageable, NameSlugModel):
@@ -361,11 +379,60 @@ def condition_url_is_blank_or_python_dot_org(column: str):
     """Conditions for a URLField column to force 'http[s]://python.org'."""
     return (
         models.Q(**{f"{column}__exact": ""})
-        | models.Q(**{f"{column}__startswith": "https://www.python.org/"})
+        | models.Q(**{f"{column}__startswith": PYTHON_DOT_ORG_HTTPS_PREFIX})
         # Older releases allowed 'http://'. 'https://' is required at
         # the API level, so shouldn't show up in newer releases.
-        | models.Q(**{f"{column}__startswith": "http://www.python.org/"})
+        | models.Q(**{f"{column}__startswith": PYTHON_DOT_ORG_HTTP_PREFIX})
     )
+
+
+def _release_file_url_values(release_file):
+    return {field_name: getattr(release_file, field_name) or "" for field_name in RELEASE_FILE_URL_FIELDS}
+
+
+def _previous_release_file_url_values(release_file):
+    if release_file.pk is None:
+        return None
+    return type(release_file).objects.filter(pk=release_file.pk).values(*RELEASE_FILE_URL_FIELDS).first()
+
+
+def _release_file_url_changed(field_name, values, previous_values):
+    if previous_values is None:
+        return True
+    return values[field_name] != (previous_values[field_name] or "")
+
+
+def _add_validation_error(errors, field_name, message):
+    errors.setdefault(field_name, []).append(message)
+
+
+def validate_release_file_urls(release_file):
+    """Validate current ReleaseFile URL writes without rejecting unchanged legacy rows."""
+    values = _release_file_url_values(release_file)
+    previous_values = _previous_release_file_url_values(release_file)
+    errors = {}
+
+    for field_name, value in values.items():
+        if not value or value.startswith(PYTHON_DOT_ORG_HTTPS_PREFIX):
+            continue
+        if _release_file_url_changed(field_name, values, previous_values):
+            _add_validation_error(errors, field_name, RELEASE_FILE_HTTPS_ERROR)
+
+    artifact_url = values["url"]
+    if artifact_url:
+        for field_name, suffix in RELEASE_FILE_SIDECAR_SUFFIXES.items():
+            sidecar_url = values[field_name]
+            expected_url = f"{artifact_url}{suffix}"
+            if not sidecar_url or sidecar_url == expected_url:
+                continue
+            _add_validation_error(
+                errors,
+                field_name,
+                f"Sidecar URL must match the artifact URL plus '{suffix}'.",
+            )
+
+    if errors:
+        raise ValidationError(errors)
 
 
 class ReleaseFile(ContentManageable, NameSlugModel):
@@ -394,6 +461,11 @@ class ReleaseFile(ContentManageable, NameSlugModel):
     sha256_sum = models.CharField("SHA256 Sum", max_length=200, blank=True)
     filesize = models.IntegerField(default=0)
     download_button = models.BooleanField(default=False, help_text="Use for the supernav download button for this OS")
+
+    def clean(self):
+        """Validate release-file URL relationships."""
+        super().clean()
+        validate_release_file_urls(self)
 
     def validate_unique(self, exclude=None):
         """Ensure only one release file per OS has the download button enabled."""
--- a/apps/downloads/models.py
+++ b/apps/downloads/models.py
@@ -386,37 +386,24 @@ def condition_url_is_blank_or_python_dot_org(column: str):
     )
 
 
-def _release_file_url_values(release_file):
-    return {field_name: getattr(release_file, field_name) or "" for field_name in RELEASE_FILE_URL_FIELDS}
-
-
-def _previous_release_file_url_values(release_file):
-    if release_file.pk is None:
-        return None
-    return type(release_file).objects.filter(pk=release_file.pk).values(*RELEASE_FILE_URL_FIELDS).first()
-
-
-def _release_file_url_changed(field_name, values, previous_values):
-    if previous_values is None:
-        return True
-    return values[field_name] != (previous_values[field_name] or "")
-
-
-def _add_validation_error(errors, field_name, message):
-    errors.setdefault(field_name, []).append(message)
-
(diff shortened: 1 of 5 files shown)


apps/downloads/tests/test_models.py
```diff
@@ -311,3 +312,96 @@ def test_release_file_urls_not_python_dot_org(self):
...
+
+        release_file.description = "Updated metadata"
+
+        with self.assertRaises(ValidationError) as cm:
+            release_file.full_clean()
+        self.assertIn("gpg_signature_file", cm.exception.message_dict)
+
+    def test_release_file_sidecar_urls_must_extend_artifact_url(self):
+        artifact_url = "https://www.python.org/ftp/python/9.7.2/Python-9.7.2-sidecar.tgz"
+        cases = {
```
sethmlarson: Is there a way we can future-proof this test case, such as gathering the set of fields that are `URLField` that aren't `url`?

apps/downloads/tests/test_views.py
```diff
@@ -412,6 +412,88 @@ def test_post_release_file(self):
...
+            "is_source": True,
+            "url": "http://www.python.org/ftp/python/2.7.5/Python-2.7.5-http.tgz",
+            "md5_sum": "098f6bcd4621d373cade4e832627b4f6",
+            "filesize": len("098f6bcd4621d373cade4e832627b4f6"),
+            "download_button": False,
+        }
+
+        response = self.json_client("post", url, data, HTTP_AUTHORIZATION=self.Authorization)
+
+        self.assertEqual(response.status_code, 400)
```
sethmlarson, line 448: Can we assert anything about the response that tells us why the rejection?

apps/downloads/tests/test_views.py
```diff
@@ -412,6 +412,88 @@ def test_post_release_file(self):
...
+            "url": artifact_url,
+            "gpg_signature_file": "https://www.python.org/ftp/python/2.7.5/Python-2.7.4-api.tgz.asc",
+            "md5_sum": "098f6bcd4621d373cade4e832627b4f6",
+            "filesize": len("098f6bcd4621d373cade4e832627b4f6"),
+            "download_button": False,
+        }
+
+        response = self.json_client("post", url, data, HTTP_AUTHORIZATION=self.Authorization)
+
+        self.assertEqual(response.status_code, 400)
```
sethmlarson, line 471: Same as above.

apps/downloads/tests/test_views.py
```diff
@@ -412,6 +412,88 @@ def test_post_release_file(self):
...
+        data = {
+            "name": "HTTP file",
+            "slug": "http-file",
+            "os": self.create_url("os", self.linux.pk),
+            "release": self.create_url("release", self.release_275.pk),
+            "description": "This is a description.",
+            "is_source": True,
+            "url": "http://www.python.org/ftp/python/2.7.5/Python-2.7.5-http.tgz",
+            "md5_sum": "098f6bcd4621d373cade4e832627b4f6",
+            "filesize": len("098f6bcd4621d373cade4e832627b4f6"),
```
sethmlarson: It doesn't really matter, but why are we calling `len()` on the hash?

apps/downloads/models.py
```diff
@@ -19,6 +19,24 @@
 from fastly.utils import purge_surrogate_key, purge_url
 
 DEFAULT_MARKUP_TYPE = getattr(settings, "DEFAULT_MARKUP_TYPE", "markdown")
+PYTHON_DOT_ORG_HTTPS_PREFIX = "https://www.python.org/"
+PYTHON_DOT_ORG_HTTP_PREFIX = "http://www.python.org/"
+RELEASE_FILE_URL_FIELDS = (
```
sethmlarson, line 24: Do we verify anywhere that this is the complete set of `URLField` for `ReleaseFile`?

apps/downloads/models.py
```diff
@@ -19,6 +19,24 @@
...
+    "url",
+    "gpg_signature_file",
+    "sigstore_signature_file",
+    "sigstore_cert_file",
+    "sigstore_bundle_file",
+    "sbom_spdx2_file",
+)
+RELEASE_FILE_SIDECAR_SUFFIXES = {
```
sethmlarson, line 32: Can we add an assertion somewhere that the keys here are the same as `RELEASE_FILE_URL_FIELDS` minus `url`?

apps/downloads/models.py
```diff
@@ -361,13 +379,62 @@ def condition_url_is_blank_or_python_dot_org(column: str):
...
+    if previous_values is None:
+        return True
+    return values[field_name] != (previous_values[field_name] or "")
+
+
+def _add_validation_error(errors, field_name, message):
+    errors.setdefault(field_name, []).append(message)
+
+
+def validate_release_file_urls(release_file):
```
sethmlarson, line 389: nit: if the above private functions are only called in `validate_release_file_urls()` maybe we can inline `_release_file_url_values` and bring the other functions inside the `validate_release_file_urls` function definition? So we don't have to wonder if they're used elsewhere.

sethmlarson (changes requested):
Thank you! Some comments:

apps/downloads/tests/test_models.py
e-q: Done. The sidecar tests now derive their cases from
`RELEASE_FILE_SIDECAR_SUFFIXES`, and I added explicit coverage that`RELEASE_FILE_URL_FIELDS` matches the model `URLField`s.

apps/downloads/tests/test_views.py
e-q: Done. The API tests now assert the rejection body includes the `url` field and
the HTTPS-only validation message, not just the 400 status.

apps/downloads/tests/test_views.py
e-q: Done. The sidecar rejection tests now assert the response body includes
`gpg_signature_file` and the expected sidecar-suffix validation message.

apps/downloads/tests/test_views.py
e-q: Odd... Replaced the copied `len()`-of-hash `filesize` values in the new
fixtures with an arbitrary file size.

apps/downloads/models.py
e-q: Done. There is now a model test that compares `RELEASE_FILE_URL_FIELDS` against
the actual `URLField`s declared on `ReleaseFile`.

apps/downloads/models.py
e-q: Done. There is now a model test asserting the sidecar suffix keys equal
`RELEASE_FILE_URL_FIELDS` minus `url`.

apps/downloads/models.py
e-q: Done. I inlined the validation-only helper functions into
`validate_release_file_urls()` so the URL validation flow is local to the one
caller.

e-q:
Thanks for the review. I pushed a follow-up commit that addresses the comments:

- Added model tests that assert `RELEASE_FILE_URL_FIELDS` matches the actual `ReleaseFile` `URLField`s, and that the sidecar suffix map covers every URL field except `url`.
- Updated sidecar validation tests to derive their cases from `RELEASE_FILE_SIDECAR_SUFFIXES`.
- Added shared v1/v2 API assertions that check the field-specific validation response body, not just the 400 status.
- Replaced the copied `len()`-of-hash `filesize` values in the new rejection fixtures with an arbitrary file size.
- Inlined the release-file URL validation helper functions into `validate_release_file_urls()`.

apps/downloads/models.py
```diff
@@ -19,6 +19,24 @@
 from fastly.utils import purge_surrogate_key, purge_url
 
 DEFAULT_MARKUP_TYPE = getattr(settings, "DEFAULT_MARKUP_TYPE", "markdown")
+PYTHON_DOT_ORG_HTTPS_PREFIX = "https://www.python.org/"
+PYTHON_DOT_ORG_HTTP_PREFIX = "http://www.python.org/"
```
JacobCoffee, line 23: why do these if we just use it in the one spot? :\

JacobCoffee (commented):
needs a rebase for final

apps/downloads/models.py
e-q: I think there is some clarity gained by naming the http vs https variations explicitly; my eyes often scan past the protocol. Happy to remove it if that's your preference

sethmlarson (approved):
LGTM, thank you! I wasn't able to "resolve" the threads, maybe you are able to?

apps/downloads/models.py
sethmlarson: I think it's fine to keep if only for symmetry with HTTPS.

e-q:
Rebased onto latest main and resolved the conflicts. The only substantive resolution note is that v1 detail updates now keep upstream's 405 behavior.

JacobCoffee (approved):

PR #3015, opened 2026-06-02
Tighten legacy v1 API boundaries

e-q:
#### Description

- Treat v1 API requests without API-key credentials as anonymous API requests.
- Require active staff API-key authentication for v1 write operations.
- Accept v1 API-key credentials only through the `Authorization` header.
- Narrow legacy v1 collection methods while preserving release-file replacement
  only when scoped to exactly one release value.
- Add focused Tastypie and v1 downloads API regression coverage for
  authentication, credential source, and collection-method boundaries.

#### Testing

- Focused Tastypie and v1 downloads API regression tests: passed.
- Whitespace check: passed.

### DIFF
--- a/apps/downloads/api.py
+++ b/apps/downloads/api.py
@@ -6,6 +6,7 @@ from rest_framework.decorators import action
 from rest_framework.response import Response
 from tastypie import fields
 from tastypie.constants import ALL, ALL_WITH_RELATIONS
+from tastypie.exceptions import BadRequest
 
 from apps.downloads.models import OS, Release, ReleaseFile
 from apps.downloads.serializers import OSSerializer, ReleaseFileSerializer, ReleaseSerializer
@@ -88,6 +89,7 @@ class ReleaseFileResource(GenericResource):
 
         queryset = ReleaseFile.objects.all()
         resource_name = "downloads/release_file"
+        list_allowed_methods = ["get", "post", "delete"]
         fields = [
             "name",
             "slug",
@@ -116,6 +118,13 @@ class ReleaseFileResource(GenericResource):
             "description": ("contains",),
         }
         abstract = False
+
+    def delete_list(self, request, **kwargs):
+        """Delete release files only when scoped to a single release."""
+        if set(request.GET) != {"release"} or len(request.GET.getlist("release")) != 1:
+            msg = "Deleting release files requires exactly one 'release' filter."
+            raise BadRequest(msg)
+        return super().delete_list(request, **kwargs)
 
 
 # Django Rest Framework
--- a/pydotorg/resources.py
+++ b/pydotorg/resources.py
@@ -1,12 +1,16 @@
 """Tastypie API resource classes and authentication/authorization backends."""
 
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import AnonymousUser
 from tastypie.authentication import ApiKeyAuthentication
 from tastypie.authorization import Authorization
 from tastypie.exceptions import Unauthorized
 from tastypie.http import HttpUnauthorized
 from tastypie.resources import ModelResource
 from tastypie.throttle import CacheThrottle
+
+API_KEY_AUTHENTICATED_ATTR = "_pydotorg_api_key_authenticated"
+LEGACY_CREDENTIAL_PARAMS = frozenset(("username", "api_key"))
 
 
 class ApiKeyOrGuestAuthentication(ApiKeyAuthentication):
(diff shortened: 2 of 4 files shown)


pydotorg/resources.py
```diff
@@ -48,19 +66,33 @@ def is_authenticated(self, request, **kwargs):
         key_auth_check = self.get_key(user, api_key)
         if key_auth_check and not isinstance(key_auth_check, HttpUnauthorized):
             request.user = user
+            setattr(request, API_KEY_AUTHENTICATED_ATTR, True)
 
         return key_auth_check
 
+    def extract_credentials(self, request):
+        """Return API key credentials from the Authorization header only."""
+        data = self.get_authorization_data(request)
```
sethmlarson, line 75: Does this method guarantee that the authorization is in the correct format? Maybe we can validate that `data` contains only a single `:`?

pydotorg/resources.py
```diff
@@ -48,19 +66,33 @@ def is_authenticated(self, request, **kwargs):
...
+        return "{}_{}".format(
+            request.META.get("REMOTE_ADDR", "noaddr"),
+            request.META.get("REMOTE_HOST", "nohost"),
+        )
 
-    def check_active(self, user):
-        """Return True, allowing inactive users to authenticate."""
-        return True
+    def _has_legacy_credentials(self, request):
+        """Return True when credentials are supplied outside the auth header."""
```
sethmlarson: nit: 'Authorization' header

pydotorg/resources.py
```diff
@@ -48,19 +66,33 @@ def is_authenticated(self, request, **kwargs):
...
+        if any(param in request.GET for param in LEGACY_CREDENTIAL_PARAMS):
+            return True
+        if request.method == "POST":
+            return any(param in request.POST for param in LEGACY_CREDENTIAL_PARAMS)
+        return False
```
sethmlarson: Is there a way to check all query parameters in a Django request regardless of method? Trying to avoid having to filter based on HTTP method.

pydotorg/resources.py
```diff
@@ -78,40 +110,44 @@ def read_detail(self, object_list, bundle):
...
         """Allow only staff users to delete individual objects."""
-        if not bundle.request.user.is_staff:
+        if not self._is_authenticated_staff(bundle.request):
             msg = "Operation restricted to staff users."
             raise Unauthorized(msg)
         return True
 
+    def _is_authenticated_staff(self, request):
+        """Return True when the request has a valid staff API-key identity."""
+        return request.user.is_staff and getattr(request, API_KEY_AUTHENTICATED_ATTR, False)
```
sethmlarson: Since this is our own custom property we're setting, can we add a `is True` check after the `getattr`? Can we also make it clear in the method name and the docstring that this excludes authentication via other methods like cookies? Maybe a method name of `_is_authenticated_staff_via_api_key()`?

pydotorg/tests/test_resources.py
```diff
@@ -17,26 +17,67 @@ def setUp(self):
...
+        self.assertIsInstance(auth.is_authenticated(request), HttpUnauthorized)
+
+    def test_authentication_rejects_legacy_query_credentials(self):
+        auth = ApiKeyOrGuestAuthentication()
+        request = HttpRequest()
+        request.GET["username"] = self.staff_user.username
         request.GET["api_key"] = self.staff_user.api_key.key
+
+        self.assertIsInstance(auth.is_authenticated(request), HttpUnauthorized)
+        self.assertFalse(getattr(request, API_KEY_AUTHENTICATED_ATTR))
```
sethmlarson: For all the assertions about our custom propery, can we do that using `assertIsFalse` and `assertIsTrue` instead of `assertFalse`/`assertTrue`?

apps/downloads/api.py
```diff
@@ -117,6 +119,13 @@ class Meta(GenericResource.Meta):
         }
         abstract = False
 
+    def delete_list(self, request, **kwargs):
```
sethmlarson, line 122: Could you explain this change re: findings?

sethmlarson (changes requested):
Thank you! Some comments and questions:

pydotorg/resources.py
e-q: Good point. get_authorization_data() validates that the header exists, has the ApiKey scheme, and has auth data, but this override should also validate the username:key shape, requiring exactly one :

pydotorg/resources.py
e-q: request.GET is method-agnostic for URL query parameters in Django, so we can check it regardless of HTTP method. The request.POST check is intentional because Tastypie’s inherited ApiKeyAuthentication.extract_credentials() also accepts username / api_key from form body fields when the Authorization header is absent. I’ll simplify this helper so it checks request.GET and request.POST directly without branching on request.method.

pydotorg/resources.py
e-q: Agreed. The point of this helper is specifically to exclude cookie/session-authenticated staff users from satisfying v1 API write authorization.

apps/downloads/api.py
e-q: This change is tied to the broad v1 Tastypie collection-mutation finding. Tastypie exposes list-level DELETE by default for all these v1 resources, which allows broad collection deletion beyond the release workflow’s needs.

The exception here is downloads/release_file: current release tooling uses DELETE /api/v1/ downloads/release_file/?release=<id> to remove the existing files for one release before posting replacements. So this keeps that known release-tooling path working, but rejects unscoped release-file collection deletes, deletes with extra filters, or deletes with multiple release values.

sethmlarson: Thank you, I understand now!

sethmlarson (approved):
All my comments have been resolved, LGTM. Thank you!

JacobCoffee (approved):

PR #3017, opened 2026-06-02
Hide unpublished releases from public download paths

e-q:
#### Description

- Limit release detail pages and legacy release redirects to published releases
  for non-staff users.
- Hide release files attached to unpublished releases from anonymous v1/v2 API
  reads.
- Add regression coverage for direct release detail, legacy redirect, and
  release-file API reads.

#### Testing

- Focused release detail, legacy redirect, and v1/v2 release-file API
  regression tests: passed.
- Syntax check for touched Python files: passed.
- Whitespace check: passed.


sethmlarson:
Since release information is public (schedules in PEPs, git tags, CI/CD infrastructure) I don't think we gain much by hiding this information.

e-q:
Gotcha

PR #3018, opened 2026-06-02
Keep 404 legacy links on the legacy host

### LINKED ISSUE #3013: Bug: 404 legacy website link can point to an external site
### Describe the bug

The custom 404 page includes a link to try the same path on
`legacy.python.org`. For a missing URL whose decoded path starts with `//`, the
legacy-link helper currently treats the path as a network-location URL and
generates a link to that external host.

For example, a path such as `/%2Fevil.test/x` can become a 404 page whose
"same page on the legacy website" link points to `http://evil.test/x` instead
of staying under `http://legacy.python.org`.

### To Reproduce

1. Use a missing python.org path whose decoded path starts with two slashes,
   such as `/%2Fevil.test/x`.
2. Render the custom 404 page.
3. Inspect the "same page on the legacy website" link.
4. The rendered `href` can use the decoded host-like path as the URL host.

Local source-level reproduction from the current helper:

```console
>>> from apps.cms.views import legacy_path
>>> legacy_path("//evil.test/x")
'http://evil.test/x'
```

### Expected behavior

The legacy-site link should always stay under the `legacy.python.org` origin.
For the decoded path above, the link should be
`http://legacy.python.org//evil.test/x`.

### Additional context

Local source-level evidence:

```text
legacy_path('//evil.test/x') = http://evil.test/x
```

The helper currently uses URL joining to combine a fixed absolute base URL with
the request path. URL joining allows a path beginning with `//` to replace the
base URL's host.

e-q:
#### Description

- Build custom 404 legacy-site links by quoting the request path and appending
  it to the fixed `legacy.python.org` origin.
- Add focused regression tests for encoded-slash and decoded leading-slash
  paths.

#### Closes

- Closes #3013

#### Testing

- Focused 404 view/helper regression tests: passed.
- Whitespace check: passed.

### DIFF
--- a/apps/cms/tests.py
+++ b/apps/cms/tests.py
@@ -81,6 +81,9 @@ class Test404(TestCase):
     def test_legacy_path(self):
         self.assertEqual(legacy_path("/any/thing"), "http://legacy.python.org/any/thing")
 
+    def test_legacy_path_without_leading_slash(self):
+        self.assertEqual(legacy_path("any/thing"), "http://legacy.python.org/any/thing")
+
     def test_legacy_path_with_encoded_slash(self):
         self.assertEqual(legacy_path("/%2Fevil.test/x"), "http://legacy.python.org/%2Fevil.test/x")
 
--- a/apps/cms/tests.py
+++ b/apps/cms/tests.py
@@ -81,6 +81,12 @@ class Test404(TestCase):
     def test_legacy_path(self):
         self.assertEqual(legacy_path("/any/thing"), "http://legacy.python.org/any/thing")
 
+    def test_legacy_path_with_encoded_slash(self):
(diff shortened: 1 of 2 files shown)


apps/cms/tests.py
```diff
@@ -81,6 +81,12 @@ class Test404(TestCase):
     def test_legacy_path(self):
         self.assertEqual(legacy_path("/any/thing"), "http://legacy.python.org/any/thing")
 
+    def test_legacy_path_with_encoded_slash(self):
```
sethmlarson, line 87: Can we add a branch that tests if the path doesn't start with `/`?

sethmlarson (approved):
One comment, otherwise LGTM.

JacobCoffee (approved):

Issue #3041, opened 2026-07-02
Proposal: add a site-wide Content-Security-Policy (Report-Only first)


tonghuaroot:
## Summary

Right now www.python.org does not send a `Content-Security-Policy` header (a quick check of the response headers shows only `X-Frame-Options: SAMEORIGIN` and HSTS). Adding a CSP would give the site meaningful defense-in-depth against XSS: even if user-influenced HTML is ever mis-sanitized, a strict `script-src` with no `unsafe-inline` stops injected scripts and `javascript:` URIs from executing.

This follows up on a request from @sethmlarson to propose a CSP for the python.org service.

## Why Report-Only first

python.org serves some inline scripts/styles and pulls assets from a CDN plus a few third-party origins, so a strict policy dropped straight into enforcing mode would break pages. The safe path is to roll out in `Content-Security-Policy-Report-Only` mode first, collect violation reports, tune the allowlist, and only then switch to enforcing.

## Proposed rollout

1. Add [`django-csp`](https://github.com/mozilla/django-csp) and enable its middleware.
2. Wire a per-request nonce into the `<script>` / `<style>` tags in the base templates.
3. Ship a `Content-Security-Policy-Report-Only` policy with a reporting endpoint and collect violations for a couple of weeks.
4. Tighten the allowlist from the reports (CDN, analytics, fonts, donate/embed widgets, and so on).
5. Flip Report-Only to enforcing once the reports are clean.

## Starting policy (Report-Only, to be tuned from reports)

```
default-src 'self';
script-src 'self' 'nonce-{request_nonce}';
style-src 'self' 'nonce-{request_nonce}';
img-src 'self' data:;
font-src 'self';
connect-src 'self';
frame-ancestors 'self';
base-uri 'self';
object-src 'none';
form-action 'self';
```

This is only a starting point. The exact `script-src` / `style-src` / `img-src` / `connect-src` origins depend on which third-party services python.org actually loads, which the Report-Only telemetry will surface.

## Open questions for maintainers

- Which third-party origins (CDN, analytics, fonts, embedded widgets, donation flow) need allowlisting?
- Nonces vs hashes for the existing inline scripts and styles?
- Preferred reporting endpoint (self-hosted report handler vs an external collector)?

Happy to open a Report-Only PR (django-csp plus nonce wiring and the starting policy above) once the approach looks right.

sethmlarson:
[...]

PR #3042, opened 2026-07-02
Add a Content-Security-Policy in Report-Only mode (refs #3041)

tonghuaroot:
Part of #3041 (adds the CSP in Report-Only mode; enforcing and origin tuning to follow).

## What this does

www.python.org currently sends no `Content-Security-Policy` header. This adds one via [`django-csp`](https://github.com/mozilla/django-csp) as defense-in-depth against XSS.

It ships in **Report-Only** mode: responses carry `Content-Security-Policy-Report-Only` and no enforcing `Content-Security-Policy`, so nothing is blocked for users. The point of this first step is to collect violation reports and tune the allowlist before switching to an enforced policy.

## Changes

- `pyproject.toml` / `uv.lock`: add `django-csp==4.0`.
- `pydotorg/settings/base.py`:
  - add `csp.middleware.CSPMiddleware` to `MIDDLEWARE`, alongside the existing `XFrameOptionsMiddleware` (which is how the site already sets a security header from Django).
  - add a `CONTENT_SECURITY_POLICY_REPORT_ONLY` policy with a strict starting set of directives.

## Starting policy

```
default-src 'self'; script-src 'self'; style-src 'self'; img-src 'self' data:;
font-src 'self'; connect-src 'self'; frame-ancestors 'self'; base-uri 'self';
object-src 'none'; form-action 'self'
```

These are intentionally strict so the Report-Only telemetry surfaces every inline script/style and third-party origin that needs a nonce, hash, or allowlist entry. Follow-ups from #3041: add a report endpoint, tune the origins, wire nonces into the base templates, then flip Report-Only to enforcing.

## Verification

Smoke-tested the config against django-csp 4.0 in isolation: it emits the `Content-Security-Policy-Report-Only` header with the directives above and no enforcing `Content-Security-Policy` header, so pages are never blocked.

### DIFF
--- a/pydotorg/settings/base.py
+++ b/pydotorg/settings/base.py
@@ -152,6 +152,7 @@ ROOT_URLCONF = "pydotorg.urls"
 MIDDLEWARE = [
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
+    "csp.middleware.CSPMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "pydotorg.middleware.AdminNoCaching",
     "pydotorg.middleware.GlobalSurrogateKey",
@@ -296,6 +297,25 @@ MESSAGE_TAGS = {
 
 X_FRAME_OPTIONS = "SAMEORIGIN"
 SILENCED_SYSTEM_CHECKS = ["security.W019"]
+
+### django-csp (Content-Security-Policy)
+
+# Report-Only first: collect violations and tune the allowlist before
+# enforcing. Rollout tracked in #3041.
+CONTENT_SECURITY_POLICY_REPORT_ONLY = {
+    "DIRECTIVES": {
+        "default-src": ["'self'"],
+        "script-src": ["'self'"],
+        "style-src": ["'self'"],
+        "img-src": ["'self'", "data:"],
+        "font-src": ["'self'"],
+        "connect-src": ["'self'"],
+        "frame-ancestors": ["'self'"],
+        "base-uri": ["'self'"],
+        "object-src": ["'none'"],
+        "form-action": ["'self'"],
+    },
+}
 
 ### django-rest-framework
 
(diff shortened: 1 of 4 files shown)


pydotorg/settings/base.py
```diff
@@ -297,6 +298,25 @@
 X_FRAME_OPTIONS = "SAMEORIGIN"
 SILENCED_SYSTEM_CHECKS = ["security.W019"]
 
+### django-csp (Content-Security-Policy)
```
sethmlarson: Add a comment to remove this and our dependency on `django-csp` when we upgrade to Django 6.0.

pydotorg/settings/base.py
```diff
@@ -297,6 +298,25 @@
...
+
+# Report-Only first: collect violations and tune the allowlist before
+# enforcing. Rollout tracked in #3041.
+CONTENT_SECURITY_POLICY_REPORT_ONLY = {
```
sethmlarson, line 308: Can we add a test case that checks which HTTP header is emitted? Want to make sure `Content-Security-Policy-Report-Only` is emitted, not `Content-Security-Policy` so we don't break everyone on rollout.

sethmlarson (commented):
Thanks, two comments:

sethmlarson:
@tonghuaroot Are you still interested in landing this PR?

pydotorg/settings/base.py
tonghuaroot: Done, added a note to drop django-csp once we move to Django 6.0.

pydotorg/settings/base.py
tonghuaroot: Added a test that checks the Report-Only header is set and the enforcing one is not. Runs green.

tonghuaroot:
Yes, still keen to land it. Pushed both of your changes and it's ready for another look.

JacobCoffee (approved):

Issue #56, opened 2014-02-20
jobs.python.org DNS failure


carljm:
Hello,

I realize this isn't a problem with this codebase per se, but I'm not sure where else to report bugs with the new website (which looks great, btw!). If there is a "where else" feel free to close this and direct me to it.

jobs.python.org is linked from the front page of the new website, but (for me at least) that subdomain fails to resolve in DNS.

benjaminp:
Goes to old jobs site for now.

PR #3091, opened 2026-08-14
:tractor: Replace jobs.python.org references with https://www.python.org/jobs/

jefftriplett:
`jobs.python.org` redirects to `https://www.python.org/jobs/`. This updates the remaining references in the fixtures: the homepage jobs box link, and the sponsorship benefit name and description.

After seeing some tombstones like https://github.com/python/pythondotorg/issues/56 this felt pretty reasonable to assume.


hugovk (approved):
Thanks!

JacobCoffee (approved):

PR #3092, opened 2026-08-14
:tractor: Improve the news links in the header menu and on the homepage

jefftriplett:
The **Latest News** section on python.org aggregates posts from several different sources, but its “More” link currently sends people to the newer `blog.python.org` site. That is confusing because not all of the news shown on python.org appears there. In particular, PSF announcements, election information, and other posts published through the PSF blog can be easy to miss unless someone already knows to visit `pyfound.blogspot.com`.

This PR makes the News section and menu a little clearer:

* Changes the homepage “More” link to point to `/blogs/`, the existing python.org page that aggregates these sources.
* Expands `/blogs/` from 6 entries to 15. The newest entry is used as the page header, with the other 14 shown in the list.
* Restyles the `/blogs/` list to match the homepage Latest News widget: the date sits in its own column next to the title, instead of leaving a large empty gap with the title and date pushed into a second column.
* Drops the “More” link from `/blogs/` itself. That page already aggregates every source, so there was no single “more” to point at. A **News Sources** widget in the right column, below Python Insider Subscriptions, now links to each site instead: Python Insider, the PSF Blog, the PyCon Blog, and the PyPI Blog.

The homepage Latest News widget itself still shows 5 entries; only its link destination changes.

This PR no longer touches the fixtures. The News menu updates (the Python Insider item, the wording, the HTTPS links, and hiding the Python Brochure entry) and the `blogs-copyright` box are applied directly to the live rows in the admin, since fixtures are seed data and do not update existing databases.

PR #3093, opened 2026-08-14
Use https for external nav links and correct the Python News hint

jefftriplett:
Follow-up to an audit of the News dropdown in `fixtures/sitetree_menus.json`.

**Add Python Insider.** The News menu linked to the PSF, community and PyCon blogs, but not to `blog.python.org` itself. New item pk 126, "Python Insider", sits directly under "Python News". Its siblings shift by one sort_order, so the menu reads: Python News, Python Insider, Community News, PSF News, PyCon News.

**Scheme.** Every external navigation link used `http://` and redirected to `https://`. I verified each destination with curl and updated the six of them:

- pk 58 PSF Blog, pk 97 PSF News — `pyfound.blogspot.com`
- pk 98 Community News — `planetpython.org`
- pk 99 PyCon News — `pycon.blogspot.com`
- pk 109 Self-certify as voting member — used the `pyfound.blogspot.de` country domain, which redirects to `pyfound.blogspot.com`; now points straight at the final URL
- pk 114 Python Brochure — scheme only

**Wording.** The Python News item (pk 96) said "Python Insider Blog Posts" and described itself as the Core Developer blog. It points at `/blogs/`, which lists every imported feed, so PSF blog entries appear there as well. The hint and description now describe that. The old wording moves to the new Python Insider item, where it is accurate.

Note: `brochure.getpython.info` returns 503 over both schemes. That is a separate problem with the destination site, so I left the item in place.

These fixtures are seed data. Production navigation lives in the database, so it needs the same edits applied there.

### DIFF
--- a/fixtures/sitetree_menus.json
+++ b/fixtures/sitetree_menus.json
@@ -1189,7 +1189,7 @@
     "fields": {
         "title": "PSF Blog",
         "hint": "",
-        "url": "http://pyfound.blogspot.com/",
+        "url": "https://pyfound.blogspot.com/",
         "urlaspattern": false,
         "tree": 1,
         "hidden": false,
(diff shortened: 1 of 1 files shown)


jefftriplett:
Folded into #3092.

PR #3099, opened 2026-08-18
Fix nominee visibility during open elections

### LINKED ISSUE #3094: Bug: Nominee list shows nominations of you, not nominations you made, and ignores the election
### Describe the bug

Before I let the Robots take over, @jonafato reported this, and I found a second bug while exploring it. Back to the robots... 

While an election's nomination window is still open, the nominee list shows you the nominations *of you*, not the nominations *you made for other people*. It is also not scoped to the election in the URL, so it mixes in nominees from every past election cycle.

The page tells you it is doing something else. The template prints "Nominations close at ..., this is a preview of only nominations relevant to you."

The cause is in `NomineeList.get_queryset` (`apps/nominations/views.py`):

```python
if self.request.user.is_authenticated:
    return Nominee.objects.filter(user=self.request.user)
```

`Nominee.user` is the person being nominated, so this returns the rows where you are the candidate. The branch above it filters on `election=election`; this one does not.

There is a second, related problem. Each row in the list links to `NomineeDetail`, which 404s unless `Nominee.visible()` passes. `visible()` (`apps/nominations/models.py`) returns True only for staff, for the nominee themselves, or after nominations close. So even once the list shows the people you nominated, a non-staff nominator who clicks one of those links gets a 404 during the open window.

`NomineeList` has no test, which is likely why this went unnoticed.

### To Reproduce

1. Sign in as an ordinary (non-staff) user.
2. Have that user submit a nomination for somebody else in an election whose nomination window is open.
3. Go to `/nominations/elections/<election-slug>/nominees/`.
4. The person you nominated is absent. Any election in which somebody nominated *you*, including closed past elections, is present.

### Expected behavior

While nominations are open, the list should show the nominees relevant to you in the election named in the URL. That means the people you nominated, plus your own candidacy if somebody nominated you. Nominees from other election cycles should not appear. Following a link from that list to a nominee you nominated should open the nominee page rather than a 404.

### URL to the issue

https://www.python.org/nominations/elections/2026-python-packaging-council/nominees/

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Browsers

Other

### Operating System

Other

### Browser Version

_No response_

### Relevant log output

```shell

```

### Additional context

This is server-side queryset logic, so it does not depend on the browser or the operating system.

Parzival235:
## Summary

- scope the open-election nominee list to the election in the URL
- show both candidates nominated by the signed-in user and the user's own candidacy
- allow nominators to open nominee detail pages during the nomination window
- add regression coverage for election scoping and nominee-detail permissions

## Root cause

The authenticated-user branch in `NomineeList.get_queryset()` filtered only on `Nominee.user` and did not filter by election. This returned nominations *of* the user across election cycles instead of nominations the user had made in the current election. Separately, `Nominee.visible()` did not consider the nominator, so a candidate linked from the corrected list still returned 404.

## Validation

- `DATABASE_URL=sqlite:///:memory: uv run --group dev python manage.py test apps.nominations.tests` — 66 tests passed
- `ruff check apps/nominations/models.py apps/nominations/views.py apps/nominations/tests/test_views.py`
- `ruff format --check apps/nominations/models.py apps/nominations/views.py apps/nominations/tests/test_views.py`

Fixes #3094

### DIFF
--- a/apps/nominations/models.py
+++ b/apps/nominations/models.py
@@ -245,10 +245,13 @@ class Nominee(models.Model):
         if self.accepted and self.approved and not self.election.nominations_open:
             return True
 
-        if user is None:
+        if user is None or not user.is_authenticated:
             return False
 
-        return bool(user.is_staff or user == self.user)
+        if user.is_staff or user == self.user:
+            return True
+
+        return self.nominations.filter(election=self.election, nominator=user).exists()
 
 
 class Nomination(models.Model):
--- a/apps/nominations/tests/test_views.py
+++ b/apps/nominations/tests/test_views.py
@@ -112,6 +112,62 @@ class UnknownElectionSlugTests(TestCase):
         self.assertEqual(self.client.get(url).status_code, 404)
 
 
+class OpenNomineeVisibilityTests(TestCase):
+    def setUp(self):
+        self.nominator = UserFactory(first_name="Ada", last_name="Lovelace")
+        self.other_user = UserFactory()
+        self.election = open_election("2026 Board Election")
+        self.other_election = open_election("2027 Board Election")
+
+        self.nominated = Nominee.objects.create(
+            user=UserFactory(first_name="Grace", last_name="Hopper"),
+            election=self.election,
+        )
+        Nomination.objects.create(
+            election=self.election,
+            nominator=self.nominator,
+            nominee=self.nominated,
+            name="Grace Hopper",
+            email="grace@example.com",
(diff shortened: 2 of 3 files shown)


hugovk:
There's already a PR for this at https://github.com/python/pythondotorg/pull/3095.

Please (tell your agent to) always check for open PRs first, and prefer to comment on the original if something is missing.

PR #3116, opened 2026-09-17
Reduce Dependabot frequency to quarterly

hugovk:
#### Description

- We get a lot of Dependabot PRs in this repo. Currently 13 open: https://github.com/python/pythondotorg/pulls?q=sort%3Aupdated-desc+is%3Apr+state%3Aopen+author%3Aapp%2Fdependabot

- Most are from Q1 this year, showing we don't merge immediately

- Notification fatigue is a thing

- Let's reduce frequency to quarterly, this matches what we have in CPython and other repos

- Security updates are still made when needed and don't wait for the schedule 

- We group the GHA PRs, we could also group the pip ones and the Docker ones?


sethmlarson:
Agreed we need to do something here, IIRC I ran into an issue with this repository and how it gels with Dependabot in the past. I proposed migrating to a lock file format that would allow us to do grouped updates (`requirements.txt`) but I would be fine with another lock file too as long as we can get better Dependabot behavior.

JacobCoffee (approved):

